PR #2, opened 2023-03-08
Add an option to choose VM CPU core count

carzacc:
Hi,

I made this little change to kernelcraft because, on my laptop, which has a Ryzen 7 6800U CPU (8C/16T), QEMU can't run ARM VMs with an amount of cores that is greater than the amount of *physical* cores (8), but kernelcraft defaults to creating the VM with as many cores as the *logical* cores of the host system (16 on mine), so I couldn't create ARM VMs using kernelcraft on my machine.

I'm opening this pull request just in case you think it might be useful to have this option in the tool.

Best regards,
Carmine Zaccagnino

### DIFF
--- a/kernelcraft/run.py
+++ b/kernelcraft/run.py
@@ -51,6 +51,9 @@ def make_parser():
     parser.add_argument('--config', '-f', action='append',
             help='Use one (or more) specific kernel .config snippet to override default config settings')
 
+    parser.add_argument('--cpus', '-p', action='store',
+        help='Set guest CPU count (qemu -smp flag)')
+
     parser.add_argument('--memory', '-m', action='store',
             help='Set guest memory size (qemu -m flag)')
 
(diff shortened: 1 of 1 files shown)


kernelcraft/run.py
```diff
@@ -50,6 +50,9 @@ def make_parser():
 
     parser.add_argument('--config', '-f', action='append',
             help='Use one (or more) specific kernel .config snippet to override default config settings')
+    
+    parser.add_argument('--cpus', '-smp', action='store',
```
arighi, line 54: second argument should be the short form of the option, typically just one letter, considering that -c and -s are used already maybe we can go with '-p' (that would stand for processors)?

arighi (changes requested):
Apart from the 2 minor issues that I commented everything looks good, if you can change those I'll merge this, thank you so much!

kernelcraft/run.py
```diff
@@ -50,6 +50,9 @@ def make_parser():
 
     parser.add_argument('--config', '-f', action='append',
             help='Use one (or more) specific kernel .config snippet to override default config settings')
+
```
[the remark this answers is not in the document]
carzacc: My bad! Fixed!

kernelcraft/run.py
carzacc: I went with `smp` exactly because the letters that made sense were all already taken, so I went with the QEMU argument name since it felt intuitive to use to me, switched to `-p`!

arighi (approved):
Looks good to me!

PR #4, opened 2023-04-04
Fix setup.py and kernelcraft.conf

marcosps:
Hi! The first patches solve problems related to running setup.py without argcomplete package, and the last one makes sure that the kernelcraft.conf is places unded $HOME/.config/kernelcraft/kernelcraft.conf.

After your fix about checking the kernelcraft.conf being on $HOME, it was needed to also change the setup.py to install the file there, thus the last pactch of this patchset removed the check for pgk_resource, since the file now _should_ be on $HOME.

I'm not sure about the approach here, but at least I can run kc --init without errors with these patches applied

Thanks!


arighi:
Apart than my small comment about the debugging print() everything else looks really good to me! If you fix that I'll merge this, thank you so much!

marcosps:
@arighi done, removed that debug print :)

arighi:
BTW, after merging this I realized that I get the following error now:
```
$ kc 
ERROR: Missing /home/righiandr/.config/kernelcraft/kernelcraft.conf
```

It looks like we implicitly moved the config file from ~/.kernelcraft.conf to ~/.config/kernelcraft/kernelcraft.conf. Honestly it looks more standard and I may even like this more (especially I like to have a dedicated dir where we can add more configs), but I don't wanna break kc for other people as well, so I think the best is to support both locations for now. What do you think?

PR #5, opened 2023-04-04
setup: Fallback to /etc/kernelcraft.conf file .config doesn't exists

marcosps:
The message about missing file still reflect the older /etc/ path, but it shouldn't be a problem for older and new new users. Also, convert configuration paths to uppercase.


arighi:
And pushed, I think now we support all the possible configurations without breaking anything. Thanks!

PR #6, opened 2023-04-05
run: Add new --compiler argument

marcosps:
There are cases when compiling a kernel with a more recent compiler can return errors. For example, on curent openSUSE Tumbleweed, gcc-13 is present, and it errors out on a couple of places.

For such cases, --compiler would help to test using different compiler versions, like in the example, using gcc-11 can compile the mainline kernel without problems.


arighi:
Nice one! I was actually planning to add this option! Tested and merged. Thanks!

PR #14, opened 2023-05-29
run.py: Fix get_kernel_version by checking for "Version" string

marcosps:
In my local kernel build, the output of "file bzImage" shows Version, instead of the expected version string:

$ file arch/x86/boot/bzImage
arch/x86/boot/bzImage: Linux/x86 Kernel, Setup Version 0x20f, bzImage, Version 6.4.0-rc3+ ...

So this fix makes virtme-run to work on current openSUSE Tumbleweed.


arighi:
Applied, thanks!

arighi:
Actually... I think in your case we may incorrectly get `0x20f` as the version. I'd be nicer to have a more reliable way to extract the kernel version instead of relying on the output of `file`, that may change in the future.

arighi:
Alright I've pushed a fix on top that should resolve this, see 202e5bf.

PR #19, opened 2023-06-01
don't use a tmpfs for cgroupv2

tych0:
we don't need this, we can just mount cgroupv2 directly.


arighi:
Oh yes! I noticed this the other day and forgot to fix it, thanks for the PR @tych0 !

PR #24, opened 2023-06-09
Use getpass.getuser() instead of os.getlogin()

SPYFF:
Python documentation recommends the `getpass.getuser()` as a more portable and reliable alternative (at least inside Python, since thats not POSIX) of `os.getlogin()`.

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -10,6 +10,7 @@ import socket
 import shutil
 import json
 import tempfile
+import getpass
 from subprocess import check_call, check_output, Popen, DEVNULL, PIPE, CalledProcessError
 from select import select
 from pathlib import Path
@@ -304,8 +305,8 @@ def get_username():
     try:
         username = os.getlogin()
     except OSError:
-        # If os.getlogin() fails, try alternative methods
-        username = os.getenv('USER') or os.getenv('LOGNAME')
+        # If os.getlogin() fails, try alternative method
+        username = getpass.getuser()
     return username
 
 class KernelSource:
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -303,10 +303,10 @@ def create_root(destdir, arch):
 def get_username():
     """Reliably get current username."""
     try:
-        username = os.getlogin()
-    except OSError:
-        # If os.getlogin() fails, try alternative method
         username = getpass.getuser()
+    except OSError:
+        # If getpass.getuser() fails, try alternative methods
+        username = os.getenv('USER') or os.getenv('LOGNAME')
     return username
 
 class KernelSource:


SPYFF:
To be honest, I got the following error on my remote box accessed through SSH:

```
 virtme-ng -r
Traceback (most recent call last):
  File "/usr/bin/virtme-ng", line 33, in <module>
    sys.exit(load_entry_point('virtme-ng==1.6', 'console_scripts', 'virtme-ng')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/virtme_ng/run.py", line 440, in main
    return do_it()
           ^^^^^^^
  File "/usr/lib/python3/dist-packages/virtme_ng/run.py", line 433, in do_it
    ks.run(arch=args.arch, root=args.root, \
  File "/usr/lib/python3/dist-packages/virtme_ng/run.py", line 303, in run
    username = '--user ' + os.getlogin()
                           ^^^^^^^^^^^^^
OSError: [Errno 6] No such device or address
```

So it might worth to fallback to a Python API when the POSIX not works. I still not sure what went wrong with `os.getlogin()`, its a regular Ubuntu 23.04 accessed on SSH.
Nevertheless, others suggested this solution for similar error: https://bugs.python.org/issue40821

arighi:
> To be honest, I got the following error on my remote box accessed through SSH:
> 
> ```
>  virtme-ng -r
> Traceback (most recent call last):
>   File "/usr/bin/virtme-ng", line 33, in <module>
>     sys.exit(load_entry_point('virtme-ng==1.6', 'console_scripts', 'virtme-ng')())
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3/dist-packages/virtme_ng/run.py", line 440, in main
>     return do_it()
>            ^^^^^^^
>   File "/usr/lib/python3/dist-packages/virtme_ng/run.py", line 433, in do_it
>     ks.run(arch=args.arch, root=args.root, \
>   File "/usr/lib/python3/dist-packages/virtme_ng/run.py", line 303, in run
>     username = '--user ' + os.getlogin()
>                            ^^^^^^^^^^^^^
> OSError: [Errno 6] No such device or address
> ```

Hm.. that's weird, the `OSError` exception should be intercepted here and fallback to `os.getenv()`:
```
def get_username():
    """Reliably get current username."""
    try:
        username = os.getlogin()
    except OSError:
        # If os.getlogin() fails, try alternative methods
        username = os.getenv('USER') or os.getenv('LOGNAME')
    return username
```

This issue should be fixed by 92a8a39. Actually looking at the trace it seems that you're missing this commit, maybe you're using a version < 1.7.

Can you try to apply 92a8a39 (or update to a version >= 1.7) to see if it solves your issue?

> 
> So it might worth to fallback to a Python API when the POSIX not works. I still not sure what went wrong with `os.getlogin()`, its a regular Ubuntu 23.04 accessed on SSH. Nevertheless, others suggested this solution for similar error: https://bugs.python.org/issue40821

I'm ok to use `getpass.getuser()` instead of `os.getlogin()` if it's more portable, but I think we should change `get_username()` in this way:

```
def get_username():
    """Reliably get current username."""
    try:
        username = getpass.getuser()
    except OSError:
        # If getpass.getuser() fails, try alternative methods
        username = os.getenv('USER') or os.getenv('LOGNAME')
    return username
```

What do you think? If you're ok with that please update your PR and I'll merge it. Thanks!

SPYFF:
Yes it's from your repo, I just missed to mention that. Your version looks better I will update my PR.

SPYFF:
Updated to your suggestion.

arighi (commented):
I think we also need an `import getpass` somewhere at the top.

SPYFF:
I think its there (https://github.com/arighi/virtme-ng/pull/24/files) or I miss something?

arighi:
> I think its there (https://github.com/arighi/virtme-ng/pull/24/files) or I miss something?

Bah... you are absolutely right, I'm not sure why it wasn't showing it before. Anyway, looks great! Thank you for this!

PR #28, opened 2023-07-07
virtme-prep-kdir-mods: Fix depmod invocation on usrmerge'd distros

marcosps:
If a virtme user tries to execute virtme-prep-kdir-mods on a distribution that is already converted to usrmerge depmod error out:

$ virtme-prep-kdir-mods
depmod: ERROR: could not open directory /home/mpdesouza/git/linux/.virtme_mods/usr/lib/modules/0.0.0: No such file or directory depmod: FATAL: could not search modules: No such file or directory

A simple symlink can fix the problem. The new link will be called usr and will point to the existing lib/modules directory that was already being created by the script.

### DIFF
--- a/bin/virtme-prep-kdir-mods
+++ b/bin/virtme-prep-kdir-mods
@@ -2,7 +2,9 @@
 
 # This is still a bit of an experiment.
 FAKEVER=0.0.0
-MODDIR=".virtme_mods/lib/modules/$FAKEVER"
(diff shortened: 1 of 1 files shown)


marcosps:
Forgot to mention that this commit fixes a problem on openSUSE Tumbleweed :)

arighi:
Applied. Thanks @marcosps !

PR #30, opened 2023-07-25
virtme-run: Find the exact location of depmod

marcosps:
Some distributions like openSUSE don't append /sbin or /usr/sbin to the default PATH, leading to issues when trying to execute depmod.

This patch adapts the same approach done on
95630104d888 ("virtme-prep-kdir-mods: Add /usr/sbin to $PATH").


marcosps:
The last version makes use of find_binary_or_raise, instead of open coding it.

arighi:
Looks good to me, applied. Thanks!

PR #31, opened 2023-10-18
overlay-rwdir: Fix the command

bwidawsk:
Currently a space is missed in creating the user added overlay directories.

I believe this requires not using `--rw`


arighi:
Nice catch! Thanks for fixing this.

PR #33, opened 2023-11-16
virtme-ng: Don't fail on missing argcomplete

zevweiss:
The argcomplete dependency isn't critical, so let's continue on in its absence instead of blowing up (convenient for running the 'vng' command directly from the source tree without having to install any python module dependencies).


arighi:
Applied, thanks!

PR #39, opened 2023-11-20
virtme-init: mount devtmpfs before using bash redirects

tych0:
Otherwise we end up with:

/run/virtme/guesttools/virtme-init: line 51: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 52: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 55: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 56: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 57: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 58: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 62: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 59: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 65: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 80: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 80: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 80: cannot redirect standard input from /dev/null: No such file or directory
/run/virtme/guesttools/virtme-init: line 89: cannot redirect standard input from /dev/null: No such file or directory

hopefully we won't need this code eventually with the rust-based init, but I have not yet managed to migrate.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -47,6 +47,27 @@ elif [[ -d "/lib/modules/$kver" ]]; then
     mount -n -t tmpfs -o ro,mode=0000 disallow_modules "/lib/modules/$kver"
 fi
 
+# devtmpfs might be automounted; if not, mount it.
+if ! grep -q devtmpfs /proc/mounts; then
+    # Ideally we'll use devtmpfs (but don't rely on /dev/null existing).
+    if [[ -c /dev/null ]]; then
+	mount -n -t devtmpfs -o mode=0755,nosuid,noexec devtmpfs /dev \
+	    &>/dev/null
+    else
+	mount -n -t devtmpfs -o mode=0755,nosuid,noexec devtmpfs /dev
+    fi
+
+    if (( $? != 0 )); then
+	# The running kernel doesn't have devtmpfs.  Use regular tmpfs.
+	mount -t tmpfs -o mode=0755,nosuid,noexec none /dev
+
+	# Make some basic devices first, and let udev handle the rest
+	mknod -m 0666 /dev/null c 1 3
+	mknod -m 0660 /dev/kmsg c 1 11
+	mknod -m 0600 /dev/console c 5 1
+    fi
+fi
+
 # Setup rw tmpfs directories
 [ -e /var/log ] && mount -t tmpfs tmpfs /var/log/ &
 [ -e /var/tmp ] && mount -t tmpfs tmpfs /var/tmp/ &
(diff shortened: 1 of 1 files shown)


arighi:
Just out of curiosity, do you have CONFIG_DEVTMPFS_MOUNT unset in your .config? I think that would explain why I don't see these errors. But overall it makes sense, I'll merge this ASAP.

tych0:
I do have, `CONFIG_DEVTMPFS_MOUNT=y`, but is that only in the initramfs before pivot_root?

arighi:
> I do have, `CONFIG_DEVTMPFS_MOUNT=y`, but is that only in the initramfs before pivot_root?

Theoretically the mount should happen after the rootfs is mounted and it shouldn't matter if we use an initramfs or not:
```
config DEVTMPFS_MOUNT
        bool "Automount devtmpfs at /dev, after the kernel mounted the rootfs"
        depends on DEVTMPFS
        help
          This will instruct the kernel to automatically mount the
          devtmpfs filesystem at /dev, directly after the kernel has
          mounted the root filesystem. The behavior can be overridden
          with the commandline parameter: devtmpfs.mount=0|1.
          This option does not affect initramfs based booting, here
          the devtmpfs filesystem always needs to be mounted manually
          after the rootfs is mounted.
          With this option enabled, it allows to bring up a system in
          rescue mode with init=/bin/sh, even when the /dev directory
          on the rootfs is completely empty.
```

Fortuantely we shouldn't have this issue with virtme-ng-init, because we do the devtmpfs remount early (so your change definitely makes sense, to be consistent with virtme-ng-init), but I would like to figure out why I can't reproduce it. What distro/kernel are you using?

arighi:
Moreover, looking at the error lines in your output, they are all mount commands... who's your `mount` redirecting to `/dev/null`?

tych0:
> how's your mount redirecting to /dev/null?

It's not, the code has a `-c` test to decide if it should do the redirect or not.

I'm using virtme-ng head with some version of the docker://ubuntu:latest userspace.

tych0:
One thing I'm doing is not shipping with the virtme-ng-init binary in the virtme/guest dir, perhaps that code path is what's causing the weirdness?

arighi:
@tych0 it might depend on something in your specific environment, but in any case, it is definitely safer to mount devtmpfs early, that is also what we do already with virtme-ng-init, therefore: applied, thanks!

PR #40, opened 2023-11-20
virtme-init: touch /etc/sudoers if it doesn't exist

tych0:
otherwise we get:

mount: /etc/sudoers: mount point does not exist.


arighi:
Applied, thanks!

PR #42, opened 2023-11-23
setup.py: Add requests to install_requires

marcosps:
If vng is installed using pipx, which starts a venv by default for the installed package, fails to find requests library. Add requests as a requirements of vng so it can be found correctly by pipx.


arighi:
Makes sense, applied, thanks!

PR #43, opened 2023-11-23
virtme-init: Add default secure_path

marcosps:
When using vng without the --no-virtme-ng-init, running a command that resides on /sbin with sudo doesn't work:

$ vng --no-virtme-me-init
$ mpdesouza@virtme-ng:~/git/linux> modprobe
Absolute path to 'modprobe' is '/usr/sbin/modprobe', so running it may require superuser privileges (eg. root). $ mpdesouza@virtme-ng:~/git/linux> sudo modprobe
sudo: modprobe: command not found

In essence, the path when running sudo is the same of the normal user. Fix this problem by creating a secure_path when running with sudo, which includes /sbin  and usr/sbin along with the common binary paths. With the fix applied, modprobe is found as expected:

$ vng --no-virtme-me-init
$ mpdesouza@virtme-ng:~/git/linux> modprobe
Absolute path to 'modprobe' is '/usr/sbin/modprobe', so running it may require superuser privileges (eg. root). $ mpdesouza@virtme-ng:~/git/linux> sudo modprobe
modprobe: ERROR: missing parameters. See -h.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -201,7 +201,8 @@ ip link set dev lo up
 if [[ -n "virtme_user" ]]; then
     real_sudoers=/etc/sudoers
     tmpfile="`mktemp --tmpdir=/tmp`"
-    echo "root ALL = (ALL) NOPASSWD: ALL" > $tmpfile
+    echo "Defaults secure_path=\"/usr/sbin:/usr/bin:/sbin:/bin\"" > $tmpfile
+    echo "root ALL = (ALL) NOPASSWD: ALL" >> $tmpfile
     echo "${virtme_user} ALL = (ALL) NOPASSWD: ALL" >> $tmpfile
     chmod 440 "$tmpfile"
     if [ ! -f "$real_sudoers" ]; then


marcosps:
@arighi IIUC, the same issue also exists with virtme-ng-init, since it only created the user entries in the sudoers without specifying the secure_path. At least in my tests the problem also existed.

arighi:
Yep, virtme-ng-init is doing the same. I haven't seen this issue, because Ubuntu `/sbin` and `/usr/sbin` are already in the `$PATH` also for non-root users.

Could you rephrase the commit message to something like the following (or similar):

"When executing vng as a non-privileged user without /sbin in their $PATH, attempting to run a command located in that directory with sudo will not succeed."

In this way the issue is more clear and then we can fix also in virtme-ng-init. Thanks for noticing this!

marcosps:
@arighi Done! Thanks for improving the commit message :)

arighi:
Thank you, applied!

arighi:
@marcosps fixed also in virtme-ng-init https://github.com/arighi/virtme-ng-init/commit/607150cef0adf49c3cbf5b2840ed517f75cba17c

PR #47, opened 2023-11-30
virtme-init: Check if virtme_user is a valid group

marcosps:
Some distributions, e.g. openSUSE, doesn't create a group for the user, but rather assigns the user to 'users' group. Without this patch, there is an error being issues on openSUSE:

  chown: invalid group: 'mpdesouza:mpdesouza'
Fix it by checking which groups are available. If a group isn't found, chown doesn't error out when 'chown user: file' is used.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -247,7 +247,7 @@ if [[ -n "${user_cmd}" ]]; then
 
     # Set proper ownership on the virtio-ports devices
     if [[ -n "${virtme_user}" ]]; then
-        chown ${virtme_user}:${virtme_user}     \
+        chown ${virtme_user}                    \
             /dev/virtio-ports/virtme.stdin      \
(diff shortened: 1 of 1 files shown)


arighi:
Thanks @marcosps, I'm wondering if we should just ignore the gid and run `chown` with the uid only. In this way we can get rid of the logic to figure out the right group name / gid (that is pointless for the vng purpose), make the code simpler (and faster, even if it's probably negligible) and more portable across distro.

If you don't see any problem I'd rather go this way and I can take care of applying the same change to virtme-ng-init.

marcosps:
@arighi no problem at all, less code is always better :)

arighi:
@marcosps applied and updated virtme-ng-init as well: https://github.com/arighi/virtme-ng-init/commit/875a6ca288e11b53d1be62c729e9041f11e79355

Thanks!

PR #62, opened 2024-01-31
virtme-configkernel: disable lockdep by default

kuba-moo:
Sort of an "RFC" posting. We (netdev) use vng to run networking tests in two flavors - normal and debug. I was surprised to see lockdep splats in the "normal" one.

Commit cd3c35c73624 ("virtme-configkernel: enable lockdep by default") enabled lockdep, and no doubt lockdep is a useful tool. But it's a bit surprising that lockdep is enabled on a normal vng build. I would have expected "vanilla vng build" to be pretty close to kvm_guest.config. lockdep slows down the kernel, eats memory and is not exactly free of false positives.

### DIFF
--- a/virtme/commands/configkernel.py
+++ b/virtme/commands/configkernel.py
@@ -154,7 +154,6 @@ _GENERIC_CONFIG_OPTIONAL = [
     "CONFIG_UPROBES=y",
     "CONFIG_UPROBE_EVENTS=y",
     "CONFIG_DEBUG_FS=y",
-    "CONFIG_PROVE_LOCKING=y",
     "# Graphics support",
     "CONFIG_DRM=y",
     "CONFIG_DRM_VIRTIO_GPU=y",


arighi:
Glad to know that you guys are using vng in netdev!

So, the reason why lockdep is enabled by default is mostly because of my typical use case: debugging (that includes catching potential locking issues in advance). And that was added before having the `--config` option.

Considering that now we can easily do something like this to re-enable lockdep:
```
$ echo CONFIG_PROVE_LOCKING=y > /tmp/vng.config
$ vng -vb --config /tmp/vng.config
```

The other thing to consider is performance. People are starting to use vng not just to test kernels, but also to test how a certain software, application, etc. works with a specific kernel, so providing a default config that pays attention to performance can be beneficial.

All that said, I have to admit that I'm not against this change. I left a comment in the review, if you can fix that I'd be happy to merge this. Thanks.

arighi:
@kuba-moo oh well, I'll just comment here, it looks like your commit is trying to change also virtme-ng-init, I'm expecting to see a change only in configkernel.py, maybe you need to refresh your local git repo and rebase your change.

PR #64, opened 2024-02-01
virtme: let user disable KVM accel easily

kuba-moo:
Add a --disable-kvm switch. Some of our testers run without KVM support (nested virt) and developers struggle to repro similarly sluggish execution for their tests. Allow turning KVM off with a single switch even when it's usable.

### DIFF
--- a/virtme/architectures.py
+++ b/virtme/architectures.py
@@ -23,8 +23,9 @@ class Arch:
         return False
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


arighi:
Applied, thanks!

I also like the fact that we rely on `can_use_kvm()` now and we can get rid of the unnecessary `os.access("/dev/kvm")` in  architectures.py.

PR #66, opened 2024-02-04
virtme-ng: redirect kernel log to stderr in interactive mode

arighi:
Redirect kernel messages to stderr when running in interactive mode, similar to what we do in command/script mode.

In this way it is possible, for example, to save kernel logs to a file even when running in interactive mode, such as:

 $ vng -vr 2>/tmp/kernel.log

NOTE: unfortunately we can't redirect early boot kernel messages, earlyprintk doesn't seem to work with virconsole devices. So, early messages will be still sent to stdout, in this way we don't break the old behavior.

This addresses discussion #60.

### DIFF
--- a/virtme/architectures.py
+++ b/virtme/architectures.py
@@ -124,7 +124,7 @@ class Arch_x86(Arch):
 
     @staticmethod
     def serial_console_args():
-        return ["console=ttyS0"]
+        return ["virtme_console=ttyS0"]
 
     @staticmethod
     def config_base():
@@ -225,7 +225,7 @@ class Arch_arm(Arch):
 
     @staticmethod
     def serial_console_args():
-        return ["console=ttyAMA0"]
+        return ["virtme_console=ttyAMA0"]
 
     def kimg_path(self):
         return "arch/arm/boot/zImage"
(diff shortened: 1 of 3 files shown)


arighi:
That's weird, on my box (ubuntu 24.04) it seems to work both with microvm enabled and disabled. Let me investigate a bit...

arighi:
> Hm, this seems to break the ability to interact with the VM for me. Also the output of the user session goes to stderr, it seems because when I do 2>/tmp/file I see the prompt in that file.

I was able to (inadvertently) reproduce this on a different box. Basically I forgot to recompile virtme-ng-init, that also require a change to support this new stderr redirection. Maybe the same happened to you.

Basically I had to do:
```
$ git submodule update
$ BUILD_VIRTME_NG_INIT=1 pip install --break-system-packages .
```

Or run vng with `--no-virtme-ng-init` to use the bash init.

Can you double check if this is the same issue that you have? Thanks!

arighi:
@kuba-moo awesome, thanks for testing!

I'll merge this, because I think it's really useful. I even used this today to dump the kernel log to a file during a test that was messing up the console (and I couldn't scroll back to see the entire kernel oops).

Therefore, let's merge this and if nobody is screaming I'll also cut a new version.

PR #67, opened 2024-02-09
virtme-ng: generate drgn compatible memory dumps

arighi:
drgn [1] is a tool that can be used for kernel debugging and it allows to inspect memory dumps using Python scripts.

drgn was developed at Meta for debugging the Linux kernel as an alternative to the crash utility [2].

At the moment virtme-ng is only capable to generate memory dumps compatible with the crash utility.

Make sure to enable all required kernel configs by default, along with the proper qemu/QMP options, to generate memory dumps that are also compatible with drgn.

Also ensure that microvm is enabled when --debug is used (it seems that microvm prevents to generate complete memory dumps).

With these changes applied it is now possible to generate a memory dump of a running vng instance and inspect it using drgn.

Example:

 - start a virtme-ng instance with --debug:

 $ vng --debug

 - trigger the dump from a separate shell session:

 $ vng --dump /tmp/vmcore.img

 - use drgn to read the 'jiffies' from the memory dump:

 $ echo "print(prog['jiffies'])" | drgn -q -s vmlinux -c /tmp/vmcore.img
 drgn 0.0.23 (using Python 3.11.6, elfutils 0.189, with libkdumpfile)
 For help, type help(drgn).
 >>> import drgn
 >>> from drgn import NULL, Object, cast, container_of, execscript, offsetof, reinterpret, sizeof
 >>> from drgn.helpers.common import *
 >>> from drgn.helpers.linux import *
 >>> (volatile unsigned long)4294675464

[1] https://github.com/osandov/drgn
[2] https://github.com/crash-utility/crash


Decave (approved):
Nice, LGTM. Only small nit I can see is that I think it should say ensure microvm is disabled when --debug is in use in the commit summary

PR #72, opened 2024-02-16
virtme-ng: opensuse tumbleweed fixes

arighi:
More fixes for openSUSE (@marcosps FYI).


marcosps:
Tested here, and it worked as expected. I don't need to specify any additional arguments to make it work. Thanks a lot!

PR #74, opened 2024-02-17
setup.py: fix logic around the virtme_ng_init submodule

WOnder93:
First, os.listdir("virtme_ng_init") throws an exception when the directory doesn't exist, which is not desirable. Replace it with os.path.exists(), which does the right thing.

Second, it only makes sense to check for the virtme_ng_init submodule when BUILD_VIRTME_NG_INIT != 0, so add it to the conditional.


arighi:
Applied, thanks!

PR #100, opened 2024-03-30
Add additional path for virtiofsd executable

jfernandez:
The Arch Linux qemu-base package depends on the [virtiofsd package](https://archlinux.org/packages/extra/x86_64/virtiofsd/), which installs the binary @ /usr/lib/virtiofsd. This PR adds that path to the possible paths.

I noticed this because qemu appears to use a lot of CPU without virtfiosd, which was making my arch laptop run pretty hot.


arighi (approved):

jfernandez:
Thank you!

 i use arch btw

PR #103, opened 2024-04-16
Introduce "--network loop" device type and allow "--network" option to be passed multiple times

jpirko:
Introduce "--network loop" device type and allow "--network" option to be passed multiple times.


arighi (approved):
Looks pretty cool, thanks for adding this!

PR #111, opened 2024-05-03
github: new test with KVM support

matttbe:
KVM is supported on "larger Linux runners" ...

  https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/

... which is the default runner for public GitHub repos since January:

  https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

I don't think this is well known, and it needs to be manually enabled, so it might be good to show an example for those who are interested in that.

*****

The announcement of the v1.24 was the trigger for this PR. When I read it the first time, it sounded that KVM acceleration could be seen in containers and GitHub Actions, while it is the setup I use for MPTCP (GitHub Actions + Docker) :)

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -24,3 +24,16 @@ jobs:
 
       # Run `uname -r` using a vanilla v6.6 kernel
       - run: ./vng -r v6.6 -- uname -r
+
+      # Setup KVM support
+      - name: "KVM support"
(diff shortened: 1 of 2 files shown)


arighi:
I didn't know that and it is FANTASTIC! Thanks for this!

PR #114, opened 2024-05-24
virtme-ng: export real /tmp to guest

ThibF:
Use /run to store overlays instead of /tmp.

Require https://github.com/arighi/virtme-ng-init/pull/9 to be merged to keep virtme-ng-init functional.


arighi (approved):
Good job! Thanks for working on this @ThibF !

Issue #89, opened 2024-03-14
man page is missing


jpirko:
$ man vng                                                           
No manual entry for vng

Any decent pkg should have man page. Could it be added for vng please?

arighi:
Definitely, I've been thinking to create a manpage for a while and the content is pretty much already defined in the README.md.

I'll prioritize this task. :)

jpirko:
Yeah, you don't really have this README handy when you install vng as a package on distro :)

arighi:
@jpirko man page added with #118, I'm closing this for now, if you notice any issue let me know. Thanks!

PR #118, opened 2024-06-05
virtme-ng: provide a man page for vng

arighi:
Automatically generate a manpage via argparse-manpage.

We can improve the content of the man page later in the future, but for now it has all the basic information about the command line options and their description, that can be useful when it's not possible to access the main README.md on the github page.

This closes issue #89.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ dist
 .mypy_cache
 virtme-ng-prompt
 vng-prompt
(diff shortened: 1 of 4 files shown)


arighi:
@ribalda heads up, not sure if this change is disrupting the deb packaging. In that case let me know and we can adjust things accordingly, thanks!

PR #123, opened 2024-06-13
vng: introduce kernel debugging mode

arighi:
Introduce a `--gdb` option in vng to automatically attach a gdb session to a running instance previously started with `--debug`.

The `--gdb` option implement some automation to figure out the location of vmlinux and provide and easy way to access kernel debugging.

TODO:
 - improve the automation to figure out also the sources of the distro kernel searching in some known locations
 - support multiple distro (right now vng is looking only for local vmlinux or Ubuntu kernel vmlinux)

This was discussed in #101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,6 +325,25 @@ Examples
    >>> (volatile unsigned long)4294675464
 ```
 
+ - Attach a gdb session to a running instance started with `--debug`:
+```
+   # Start the vng instance in debug mode
+   $ vng --debug
+
+   # In a separate terminal run the following command to attach the gdb session:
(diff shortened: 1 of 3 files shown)


SPYFF:
Excellent! I'll try it tomorrow on Debian testing. The --gdb option is great I like it!

SPYFF:
image

I can confirm it works on Debian Sid! Thanks for the effort, I'll try my best to extend this feature in the future.

arighi:
@SPYFF awesome! Let's just merge this for now and we can continue to improve it in-tree! Thanks!

PR #124, opened 2024-06-18
init: Fall back to mounting without xino option

juergh:
Older kernels don't support the 'xino' overlayfs mount option so we end up with no overlay mounts. Fall back to mounting without that option to work around that. It's not pretty as we end up with errors like:

[    0.380206] overlayfs: unrecognized mount option "xino=off" or missing value
[    0.383246] overlayfs: unrecognized mount option "xino=off" or missing value

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -28,8 +28,9 @@ for tag in "${!virtme_rw_overlay@}"; do
     upperdir="/run/tmp/$tag/upper"
     workdir="/run/tmp/$tag/work"
     mkdir -p "$upperdir" "$workdir"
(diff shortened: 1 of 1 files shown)


arighi (approved):
LGTM, thanks @juergh!

We should apply the same change also to virtme-ng-init (https://github.com/arighi/virtme-ng-init), if you feel brave with Rust you can send another PR, otherwise I'll work on that. :)

arighi:
For the records, the same change has been applied to virtme-ng-init as well https://github.com/arighi/virtme-ng-init/commit/6e2b515c5c178ba2ae3206933024ad24f658f368

PR #131, opened 2024-06-27
version: generate verison string based on git information

arighi:
If virtme-ng is installed from the git repository add git information to the version string.

If, for any reason, we fail to determine the git information (e.g., when building virtme-ng from a distro source package), fallback to the main packaging version, typically updated manually every time we cut a new release.

Suggested-by: Jim Cromie <jim.cromie@gmail.com>

### DIFF
--- a/virtme_ng/version.py
+++ b/virtme_ng/version.py
@@ -3,7 +3,39 @@
 
 """virtme-ng version"""
 
-VERSION = "1.25"
+from subprocess import check_output, DEVNULL, CalledProcessError
 
-if __name__ == '__main__':
+PKG_VERSION = "1.25"
+
+
+def get_version_string():
+    try:
+        # Get the version from git describe
+        version = (
+            check_output(
+                "git describe --always --long --dirty",
+                shell=True,
+                stderr=DEVNULL,
+            )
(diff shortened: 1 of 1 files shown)


jimc:
we have a glitch.

```
[jimc@frodo builds]$ vng --version
virtme-ng 1.25
[jimc@frodo builds]$ vng --version
virtme-ng 6.9+54.gf89c53fdd66f
[jimc@frodo builds]$ pwd
/home/jimc/projects/lx/linux.git/builds
[jimc@frodo builds]$ 
```

after make install, vng picks up git describe from a linux tree too.

is it possible to capture it at make install, and pin it there ?

arighi:
Ah good catch @jimc ! Hm... fixing this is not trivial, because we can run vng directly from the git repo even without installing it. Maybe we should try to run git describe from the dir of the python file itself.  Let me see if I can figure out a fix for this...

arighi:
@jimc something like #132 can help... but only when we run vng from the git dir directly. For all the other cases we just fallback to the old behavior, but I guess that's the best we can do and at least in this way we don't get totally wrong version strings.

PR #134, opened 2024-07-14
version: make sure to get the git version from a virtme-ng git repo

arighi:
When we run git describe to append the git information to the version string, make sure that the version.py module is contained in a virtme-ng git repository, to avoid using information from any git repository in any parent directory.

Reported-by: Matthieu Baerts (NGI0) <matttbe@kernel.org>


matttbe (approved):
Nice, thank you for this fix!

PR #137, opened 2024-07-16
Add cross-compile argument

frelon:
When cross-compiling on SUSE distributions vng fails to find gcc, since it's named '<arch>-suse-linux-gcc', instead of the hardcoded '<arch>-linux-gnu-gcc'.

This commit provides a '--cross-compile' argument that will override the hardcoded prefix if used, otherwise it should do nothing.

### DIFF
--- a/virtme/commands/configkernel.py
+++ b/virtme/commands/configkernel.py
@@ -32,6 +32,13 @@ def make_parser():
     )
 
     parser.add_argument(
+        "--cross-compile",
+        action="store",
+        metavar="CROSS_COMPILE_PREFIX",
+        help="Cross-compile compiler prefix",
+    )
+
+    parser.add_argument(
         "--custom",
         action="append",
         metavar="CUSTOM",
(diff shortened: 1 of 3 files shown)


arighi (approved):

virtme/commands/configkernel.py
```diff
@@ -313,9 +320,13 @@ def do_it():
...
 
-    if shutil.which(f"{arch.gccname}-linux-gnu-gcc") and arch.gccname != uname.machine:
-        gccname = shlex.quote(f"{arch.gccname}-linux-gnu-")
-        archargs.append(f"CROSS_COMPILE={gccname}")
+    cross_compile_prefix = f"{arch.gccname}-linux-gnu-"
+    if args.cross_compile != "":
+        cross_compile_prefix = args.cross_compile
+
+    if shutil.which(f"{cross_compile_prefix}-gcc") and arch.gccname != uname.machine:
+        gccname = shlex.quote(cross_compile_prefix)
```
arighi, line 328: gccname is defined but never used ehre

virtme_ng/run.py
```diff
@@ -733,6 +739,9 @@ def make(self, args):
                 arg_fail(f"unsupported architecture: {arch}")
             target = ARCH_MAPPING[arch]["kernel_target"]
             cross_compile = ARCH_MAPPING[arch]["cross_compile"]
+            if args.cross_compile != "":
+                cross_compile=args.cross_compile
```
arighi, line 743: minor coding style nit: `cross_compile = args.cross_compile`

PR #138, opened 2024-07-17
version: try to get version from pip

arighi:
If we can't determine any version info from `git describe` try to get the version from pip, otherwise fallback to the hard-coded package version.

Also force to use the git version string both when we are installing (setup.py) the package or when we are directly running vng from the source repo.

### DIFF
--- a/virtme_ng/version.py
+++ b/virtme_ng/version.py
@@ -5,12 +5,23 @@
 
 import os
 from subprocess import check_output, DEVNULL, CalledProcessError
+import pkg_resources
 
 PKG_VERSION = "1.25"
 
 
+def get_package_version():
+    try:
+        return pkg_resources.get_distribution("virtme-ng").version
+    except pkg_resources.DistributionNotFound:
+        return PKG_VERSION
+
+
 def get_version_string():
     try:
+        if not os.environ.get("__VNG_LOCAL"):
+            return get_package_version()
+
         # Get the version from `git describe`.
         #
         # Make sure to get the proper git repository by using the directory
@@ -37,9 +48,9 @@ def get_version_string():
 
         return version_pep440
     except CalledProcessError:
-        # Default version if git describe fails (e.g., when building virtme-ng
-        # from a source package.
-        return PKG_VERSION
+        # If git describe fails to determine a version, try to use the version
+        # from pip, or ultimately simply return the hard-coded package version.
+        return get_package_version()
 
 
 VERSION = get_version_string()
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,10 @@ from subprocess import check_call, CalledProcessError
 from setuptools import setup, Command
 from setuptools.command.build_py import build_py
 from setuptools.command.egg_info import egg_info
-from virtme_ng.version import VERSION
+from virtme_ng.version import get_version_string
+
+os.environ["__VNG_LOCAL"] = "1"
+VERSION = get_version_string()
 
 # Source .config if it exists (where we can potentially defined config/build
 # options)
@@ -52,6 +55,7 @@ class LintCommand(Command):
             for cmd in ("flake8", "pylint"):
                 command = [cmd]
                 for pattern in (
+                    "vng",
                     "*.py",
                     "virtme/*.py",
                     "virtme/*/*.py",
--- a/vng
+++ b/vng
@@ -7,7 +7,9 @@
 
 import os
 import sys
-from virtme_ng import run
+os.environ["__VNG_LOCAL"] = "1"
+from virtme_ng import run  # noqa: E402
+
 
 # Update PATH to make sure that virtme-ng can be executed directly from the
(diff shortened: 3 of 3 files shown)


arighi:
@matttbe FYI, just merged this that should improve a bit `vng --version`. With this one vng will try to generate the git info when executed from the source dir directly or during the install phase (setup.py).

Once it's installed it'll try to get the version from pip (previously generated with the git info during install) or, if it fails, it'll fallback to the hard-coded version.

Basically with this in place `vng --version`, installed via pip, will be able to show the git info as well.

virtme_ng/version.py
```diff
@@ -37,9 +48,9 @@ def get_version_string():
 
         return version_pep440
     except CalledProcessError:
-        # Default version if git describe fails (e.g., when building virtme-ng
-        # from a source package.
-        return PKG_VERSION
+        # If git describe fails to determine a version, try to use the version
+        # from pip, or ultimately simply return the hard-coded package version.
+        return get_package_version()
```
matttbe, line 53: Is it safe to fallback to the version from pip here if you have a "local" version? I mean: if we are here, it means we are not using a version from pip from what I understood, then we might be looking at a wrong (old) version, no?

virtme_ng/version.py
```diff
@@ -37,9 +48,9 @@ def get_version_string():
...
 
 
 VERSION = get_version_string()
```
matttbe, line 56: (not related to this PR, and it's a detail, but if the version is only used when requested, maybe it would be better not to export `VERSION`, only `get_version_string`, not to query git each time vng is used :) )

(and maybe by doing that, arguments could be passed to `get_version_string()` instead of relying on `__VNG_LOCAL`, but again, detail, feel free to ignore :) )

matttbe:
@arighi thank you for the notification!

> With this one vng will try to generate the git info when executed from the source dir directly or during the install phase (setup.py).
> 
> Once it's installed it'll try to get the version from pip (previously generated with the git info during install) or, if it fails, it'll fallback to the hard-coded version.

Ah yes, that's sounds good and safer!

> Basically with this in place `vng --version`, installed via pip, will be able to show the git info as well.

Just to be sure, do you mean when installing from the source with pip?
I don't know well `pip`, but when using `pip install virtme-ng`, will it retrieve the source code with Git, or a tarball from Pypi?
If it is tarball, it will then fall back to the hard-coded version I suppose.

arighi:
> > Basically with this in place `vng --version`, installed via pip, will be able to show the git info as well.
> 
> Just to be sure, do you mean when installing from the source with pip? I don't know well `pip`, but when using `pip install virtme-ng`, will it retrieve the source code with Git, or a tarball from Pypi? If it is tarball, it will then fall back to the hard-coded version I suppose.

You can install virtme-ng via pip from the source dir, like `pip install .`, in that case the idea is to use the git version, pass it to setup.py and it'll be used to generate the package version. So once it's installed if you run vng it'll grab the git version from the package itself.

This method can be used also when I upload a new version to pypi, so the "official" package could also keep the git version (or I can add another env variable to force the hard-coded package version and keep using the same versioning for the official pypi packages - I think I like the latter more).

virtme_ng/version.py
arighi: Indeed! In this case we definitely want to use `PKG_VERSION`, good catch (feel free to send a PR if you want, otherwise I'll fix it later)

virtme_ng/version.py
arighi: I was also thinking about that and honestly I don't see any downside, but I need to do some tests. Will work on that, thanks for the review!

matttbe:
> > > Basically with this in place `vng --version`, installed via pip, will be able to show the git info as well.
> > 
> > 
> > Just to be sure, do you mean when installing from the source with pip? I don't know well `pip`, but when using `pip install virtme-ng`, will it retrieve the source code with Git, or a tarball from Pypi? If it is tarball, it will then fall back to the hard-coded version I suppose.
> 
> You can install virtme-ng via pip from the source dir, like `pip install .`, in that case the idea is to use the git version, pass it to setup.py and it'll be used to generate the package version. So once it's installed if you run vng it'll grab the git version from the package itself.

Thank you for the explanation!

> This method can be used also when I upload a new version to pypi, so the "official" package could also keep the git version (or I can add another env variable to force the hard-coded package version and keep using the same versioning for the official pypi packages - I think I like the latter more).

OK, I never uploaded packages to pypi, that's nice if you do that. On the other hand, if you only upload tags, I guess you don't need the git SHA (looks like a dev version), except if you often modify the uploaded tags, but that's maybe not a good idea :)
Or is it not to have to maintain a hard-coded version? (but then people using tarball will not have this version)

virtme_ng/version.py
matttbe: > feel free to send a PR if you want

done! #139

PR #143, opened 2024-07-27
virtme: get rid of virtme/scripts symlink

arighi:
There is no reason to maintain a symlink from virtme/scripts -> bin, we can just run the scripts directly from bin, when vng is executed directly from the source dir (otherwise all the scripts will be available in the PATH and we don't need to do anything special).

This can solve potential issues with packaging that may have problems to preserve the symlinks.


jimc:
FWIW, this appears to fix an odd make install problem
I had on my laptop, but not on my desktop.

the error looked like this:
```
  adding license file 'LICENSE'
  writing manifest file 'virtme_ng.egg-info/SOURCES.txt'
  error: can't copy 'virtme/scripts': doesn't exist or not a regular file
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  full command: /usr/bin/python3 -u -c '
  exec(compile('"'"''"'"''"'"'
  # This is <pip-setuptools-caller> -- a caller that pip uses to run setup.py
  #
  # - It imports setuptools before invoking setup.py, to enable projects that directly
  #   import from `distutils.core` to work with newer packaging standards.
  # - It provides a clear error message when setuptools is not installed.
  # - It sets `sys.argv[0]` to the underlying `setup.py`, when invoking `setup.py` so
  #   setuptools doesn'"'"'t think the script is `-c`. This avoids the following warning:
  #     manifest_maker: standard file '"'"'-c'"'"' not found".
  # - It generates a shim setup.py, for handling setup.cfg-only projects.
  import os, sys, tokenize
  
  try:
      import setuptools
  except ImportError as error:
      print(
          "ERROR: Can not execute `setup.py` since setuptools is not available in "
          "the build environment.",
          file=sys.stderr,
      )
      sys.exit(1)
  
  __file__ = %r
  sys.argv[0] = __file__
  
  if os.path.exists(__file__):
      filename = __file__
      with tokenize.open(__file__) as f:
          setup_py_code = f.read()
  else:
      filename = "<auto-generated setuptools caller>"
      setup_py_code = "from setuptools import setup; setup()"
  
  exec(compile(setup_py_code, filename, "exec"))
  '"'"''"'"''"'"' % ('"'"'/home/jimc/projects/virtme-ng.git/setup.py'"'"',), "<pip-setuptools-caller>", "exec"))' bdist_wheel -d /tmp/pip-wheel-2ba_xo99
  cwd: /home/jimc/projects/virtme-ng.git/
  Building wheel for virtme-ng (setup.py) ... error
  ERROR: Failed building wheel for virtme-ng
  Running setup.py clean for virtme-ng
  Running command python setup.py clean
  running clean
  removing 'build/lib' (and everything under it)
  'build/bdist.linux-x86_64' does not exist -- can't clean it
  'build/scripts-3.12' does not exist -- can't clean it
  removing 'build'
Failed to build virtme-ng
ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (virtme-ng)
```

PR #144, opened 2024-07-28
virtme: pass "debug" to the boot options when running in verbose mode

arighi:
Some distro / kernels require the "debug" boot option to properly show the kernel messages to the console.

Add this parameter to correctly show the kernel log when "-v" is used.

This allows to properly show the kernel log in Arch Linux, CachyOS (and probably most of the Arch-derivative distro).

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1063,7 +1063,7 @@ def do_it() -> int:
             qemuargs.extend(["-chardev", "file,path=/proc/self/fd/2,id=dmesg"])
             qemuargs.extend(["-device", arch.virtio_dev_type("serial")])
             qemuargs.extend(["-device", "virtconsole,chardev=dmesg"])
-            kernelargs.extend(["console=hvc0"])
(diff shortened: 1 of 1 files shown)


virtme/commands/run.py
```diff
@@ -1063,7 +1063,7 @@ def do_it() -> int:
             qemuargs.extend(["-chardev", "file,path=/proc/self/fd/2,id=dmesg"])
             qemuargs.extend(["-device", arch.virtio_dev_type("serial")])
             qemuargs.extend(["-device", "virtconsole,chardev=dmesg"])
-            kernelargs.extend(["console=hvc0"])
+            kernelargs.extend(["console=ttyS0", "debug"])
```
matttbe, line 1066: @arighi just to be sure, was the `console=` modification expected? I don't think that's an issue to switch to `ttyS0`, just surprised :)

arighi:
@matttbe oops, it's totally a mistake, thanks for noticing it! Feel free to send a PR to restore it back to `hvc0`. Otherwise I'll fix it later.

PR #145, opened 2024-07-29
virtme: switch back to hypervisor console

matttbe:
It looks like it was accidentally modified [1].

The hypervisor console is supposed to be better, except to debug early console stuff. But that's OK, another serial is used for that: ttyS0 or ttyAMA0, depending on the arch.

Fixes: 4bf382a ("virtme: pass "debug" to the boot options when running in verbose mode")
Link: https://github.com/arighi/virtme-ng/pull/144 [1]


arighi (approved):

arighi:
Applied, thanks @matttbe !

Issue #141, opened 2024-07-23
Unable to insert kernel modules while running kernel in a separate chroot


comeillfoo:
I've encountered a broken symbolic link for `/lib/modules/$kver` if the `--root` points to a rootfs directory that is not the parent of the directory where the kernel sources are located.
E.g. kernel sources at directory `linux` and rootfs is sibling directory to sources:
```
.
|-- linux
`-- rootfs
```

So, when running `vng --root ../rootfs` being at `linux` I have modules at `linux/.virtme_mods/lib/modules/0.0.0` and `virtme_link_mods` will be set to `/../linux/.virtme_mods/lib/modules/0.0.0` because of `os.path.relpath(kernel.moddir, args.root)` ([virtme/commands/run.py#L991](https://github.com/arighi/virtme-ng/blob/main/virtme/commands/run.py#L991)).

Hence, `virtme-init` tries to symlink `/lib/modules/$kver` outside of the provided rootfs.

As a workaround I add a disk image (using `--disk`) with the missing directory, mount it to `/tmp` and manually relink `/lib/modules/$kver` to `/tmp/my_disk/.virtme_mods/lib/modules/0.0.0`. But are there other solutions besides this?

jimc:
hey, it seems from your investigations, youre willing to hack some.

vng uses qemu to run the kernel in a vm, 
the "--kdir ./" arg to it makes module-loading path relative to $PWD

If this isnt the cause itself, there could be other path-relative issues.

comeillfoo:
> hey, it seems from your investigations, youre willing to hack some.
> 
> vng uses qemu to run the kernel in a vm, the "--kdir ./" arg to it makes module-loading path relative to $PWD
> 
> If this isnt the cause itself, there could be other path-relative issues.

hi, in fact, I'm willing to test linux kernel on arm64. That's why I have to use not a root folder as rootfs.

I haven't checked it but it seems that varying `--kdir` won't work for me.

[...]

PR #148, opened 2024-08-01
Outdir

jimc:
hi Andrea,

this is a rework of the previous patchset, some of which you committed already,
then I closed the PR.   Things are basically unchanged, with more confident commit-msgs.

Its all about O=outdir, 1st in config-kernel, then in vng, which uses it.

NB:  O=outdir is implicitly path relative, as it translates to --kdir ./  qemu arg.
This may relate at least peripherally to PR #141


arighi:
Looks good & merged. Thanks for working at this!

PR #154, opened 2024-08-23
Two fixes for small issues encountered with "--build-host"

brenns10:
I encountered two issues while using `vng --build --build-host` and thought I'd fix them up. One is due to my use of [git worktrees](https://git-scm.com/docs/git-worktree/en) (so the `.git` is not a directory), and the other was an issue that came up due to the first-time `git push` to my build host.

Thanks for this project, it's really convenient :)


arighi (approved):
LGTM, thanks!

PR #155, opened 2024-08-27
virtme-configkernel: disable nvram support

arighi:
We don't need to enable nvram support by default in the guest and having this disabled can save a few millisec during boot.

### DIFF
--- a/virtme/commands/configkernel.py
+++ b/virtme/commands/configkernel.py
@@ -277,6 +277,7 @@ _GENERIC_CONFIG_OPTIONAL = [
     "# CONFIG_MSDOS_FS is not set",
     "# CONFIG_AUTOFS4_FS is not set",
     "# CONFIG_AUTOFS_FS is not set",
+    "# CONFIG_NVRAM is not set",
 ]
 
 


matttbe (approved):
Good idea!

To accelerate the build, I also disable some options: I added a few a long time ago, also because we don't need sound, or DRM. But I don't think people using virtme needs LED drivers for example :)

https://github.com/multipath-tcp/mptcp-upstream-virtme-docker/blob/da60c60ccec923fde7c7e31dd07ea8ce2c2e6ba7/entrypoint.sh#L462-L466

matttbe:
Without `NVRAM`, it now takes 0.5 second to boot on my side!

```
[    0.233247] Run /usr/local/lib/python3.12/dist-packages/virtme/guest/bin/virtme-ng-init as init process
(...)
[    0.239021] virtme-ng-init: mount devtmpfs -> /dev: EBUSY: Device or resource busy
[    0.324961] systemd-tmpfile (53) used greatest stack depth: 13280 bytes left
[    0.330561] ip (55) used greatest stack depth: 12968 bytes left
[    0.330941] virtme-ng-init: setting up network device eth0
[    0.334850] virtme-ng-init: Starting systemd-udevd version 255.4-1ubuntu8.4
[    0.334931] virtme-ng-init: triggering udev coldplug
[    0.364034] ip (63) used greatest stack depth: 12792 bytes left
[    0.386586] virtme-ng-init: waiting for udev to settle
[    0.476824] virtme-ng-init: udhcpc: started, v1.36.1
[    0.476824] udhcpc: broadcasting discover
[    0.476824] udhcpc: broadcasting select for 10.0.2.15, server 10.0.2.2
[    0.476824] udhcpc: lease of 10.0.2.15 obtained from 10.0.2.2, lease time 86400
[    0.504215] virtme-ng-init: udev is done
[    0.507673] virtme-ng-init: initialization done
```

1/4 of second to reach the init script!

arighi:
> Good idea!
> 
> To accelerate the build, I also disable some options: I added a few a long time ago, also because we don't need sound, or DRM. But I don't think people using virtme needs LED drivers for example :)

At some point we should do a review of the default configs, but don't underestimate how people are using virtme-ng. :smile: 

There are people using it to test GUIs or even webcams... and we have `--sound` and `--graphics`, so we need to keep sound and drm enabled by default. Maybe we should consider to have a `--config-minimal`, but at the end things like sound/drm don't impact much on boot time, and I think it's more important to make this part fast.

matttbe:
(With a bit of delay)

> Very nice! How about time vng -- uname -r?

Around one second, not bad!

```shell
$ cd <kernel build dir>
$ time vng -- uname -r
6.11.0-rc5+
vng -- uname -r  1.17s user 1.43s system 203% cpu 1.281 total
```

arighi:
Nice! Thanks! :)

PR #160, opened 2024-09-14
Avoid prepending the version with `0.0+dev` all the time

matttbe:
A recent change has modified how the version was printed. I think the issue was due to the git repo not having any tags, causing an unexpected string, not looking like a version.

Instead, a fallback to the package version is now done when there are no git tags.


arighi (approved):
Makes sense to me, thanks for fixing it!

likewhatevs:
Thanks!

PR #161, opened 2024-09-24
Enable ppc64le to work with "vng -r" argument

marcosps:
Currently is also fails on s390x, but because we fail to detect the kernel version. But this work will also to bring s390x support in the future.


arighi (approved):
LGTM, thanks!

PR #162, opened 2024-09-25
S390x enablement

marcosps:
These two commits enabled us to run ./vng -r on our s390x SLE machines. I'm not sure if there are s390x experts around to double check the "image" naming of the kernel installed on s390x machines, but this is true for SLE.


arighi (approved):
I also don't have access to an s390 system at the moment, but the change looks safe, so let's just merge it and if there are issues with other distros we can fix them later, thanks!

PR #165, opened 2024-09-30
NVIDIA gpu support

ianm-nv:
This PR enables gpu passthrough with virtme-ng.  It adds a new paramemeter "--nvgpu' which is used to provide the PCI address of the gpu.  There are included instructions in the README for enabling VFIO passthrough support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,26 @@ Examples
    See also: `.github/workflows/run.yml` as a practical example on how to use
    virtme-ng inside docker.
 
+ - Run virtme-ng with gpu passthrough:
+```
+   # Confirm host kernel has VFIO and IOMMU support
+   # Check if NVIDIA module is installed on the host
+   $ modinfo nvidia
+   # If the nvidia module is installed, blacklist the nvidia modules
+   $ sudo bash -c 'echo -e "blacklist nvidia\nblacklist nvidia-drm\nblacklist nvidia-modeset\nblacklist nvidia-peermem\nblacklist nvidia-uvm" > /etc/modprobe.d/blacklist-nvidia.conf'
+   # Host will need to be rebooted for blacklist to take effect.
+   # Get GPU device ID
+   $ lspci -nn | grep NVIDIA
+     0000:01:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD104GLM [RTX 3500 Ada Generation Laptop GPU] [10de:27bb] (rev a1)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -358,6 +358,26 @@ Examples
...
+   # Get GPU device ID
+   $ lspci -nn | grep NVIDIA
+     0000:01:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD104GLM [RTX 3500 Ada Generation Laptop GPU] [10de:27bb] (rev a1)
+     0000:01:00.1 Audio device [0403]: NVIDIA Corporation Device [10de:22bc] (rev a1))
+   # Configure VFIO for device passthrough
+   $ sudo bash -c 'options vfio-pci ids=10de:27bb,10de:22bc' > /etc/modprobe.d/vfio.conf
+   # Load VFIO module
+   $ sudo modprobe vfio-pci
+   # Pass PCI address to virtme-ng
+   $ sudo vng --nvgpu "01:00.0" -r linux
```
arighi, line 378: Do we really need sudo permission here to run vng? That might be the case because we need to access pci directly I guess... do you get an permission error if you try to run vng as regular user?

arighi (approved):
Looks great overall, thanks for working on this!

I just left a small comment about running vng as regular user, but we can definitely merge it and address this later in-tree.

SPYFF:
While this is a good addition, I wonder if it can be generalised. I used to pass-through the network interface with VFIO and the steps were almost identical. I'm afraid there are driver and device specific quirks, but the basic steps were the same as in this case:
1. Find the PCIe address and device ID of the device
2a. Unbind it from its driver (on a running system)
2b. Blacklist the driver (if runtime unbind failed)
3. Load the vfio-pci driver
4. Map the device to the vfio-pci driver
5. May or may not be visible and usable in QEMU

Sure, success is highly device and host dependent (e.g. the same i225 NIC worked in the VM on an Intel host, but not on an AMD, etc.), but a generic PCIe passthrough feature might be worth considering.

arighi:
Absolutely, I totally agree on generalizing the concept of PCI passthrough and I think we can improve this in-tree. At this stage even if we break the API (command options) is not so critical as long as we don't break any previous behavior or option (maybe we should have a way to mark new options as "unstable"?). But I think it's good to have "something" merged, a lot of people have been asking me this feature for a while, so at least now people can start using it and improving it. 🙂

That said, I'll think about a proper interface for a generic PCI passthrough and maybe we can discuss it here or better create a separate issue.

Thanks!

Issue #173, opened 2024-10-04
virtme-ng kernel downloader should use native arch without having to pass --arch option


anisse:
On all platforms, the kernel downloader defaults to `amd64` if no architecture is passed on the command line. It should just select the current native architecture instead.

Here is an example, on arm64 machine:

```
> ./vng --run v6.6.50 -v
use v6.6.50/amd64 pre-compiled kernel from https://kernel.ubuntu.com/mainline/v6.6.50/amd64
```
It will then fail to start it, and it's even impossible to specify x86_64 as an architecture in order to start it (but that might be another issue).

arighi:
@anisse can you double check if #176 fixes this? Thanks!

anisse:
I commented on the PR, and I confirmed it worked and fixed this issue.

PR #176, opened 2024-10-05
virtme-ng: auto-detect host architecture

arighi:
Instead of implicitly assuming amd64 as the host architecture, detect it automatically so the user doesn't need to manually specify --arch on non-amd64 systems.

This fixes issue #173.


arighi:
@anisse hahaha sorry for wasting your time, but I'm considering this a success. :D

PR #178, opened 2024-10-08
Modernize build system

tamird:
Remove `requirements.txt`; runtime dependencies are inlined in
`install_requires` in `setup.py` and build dependencies are moved to
`pyproject.toml`.

This allows building with `uv` to succeed.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -11,17 +11,26 @@ jobs:
       matrix:
         python-version: ["3.8", "3.9", "3.10"]
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
+        set -euxo pipefail
+
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
-        pip install pylint
-        pip install flake8
+
+        # Both build and runtime deps. This is the price of using pip.
+        pip install 'argparse-manpage[setuptools]' argcomplete requests
+
+        pip install  pylint flake8
     - name: Analysing the code with pylint
       run: |
-        python setup.py lint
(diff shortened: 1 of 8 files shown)


tamird:
Context: I started hacking on this because I wanted to run this on my mac. In hindsight, that doesn't make any sense because my host filesystem contains a bunch of binaries that can't possibly run in the VM -- but I apparently had to get to the VM booting to realize that.

I'd be happy to send up the rest of the patches that made that work, if you can think of a generalization of this tool that would make sense on macOS.

arighi:
> Context: I started hacking on this because I wanted to run this on my mac. In hindsight, that doesn't make any sense because my host filesystem contains a bunch of binaries that can't possibly run in the VM -- but I apparently had to get to the VM booting to realize that.
> 
> I'd be happy to send up the rest of the patches that made that work, if you can think of a generalization of this tool that would make sense on macOS.

Absolutely! Feel free to send the other patches, people have been asking for this so I'm more than happy to help.

In theory if you prepare a separate chroot with all the linux binaries it should work. You could even `vng --root-release oraculr --root /path/to/empty/dir` and vng will automatically create an Ubuntu chroot for you, that should be enough to have a working linux kernel testing environment on macOS.

SPYFF:
With `uv` you can build the vng init as well?

tamird:
> > Context: I started hacking on this because I wanted to run this on my mac. In hindsight, that doesn't make any sense because my host filesystem contains a bunch of binaries that can't possibly run in the VM -- but I apparently had to get to the VM booting to realize that.
> > I'd be happy to send up the rest of the patches that made that work, if you can think of a generalization of this tool that would make sense on macOS.
> 
> Absolutely! Feel free to send the other patches, people have been asking for this so I'm more than happy to help.

Will do! The changes are mostly about running qemu correctly and compiling the init program for Linux (rather than host).

> In theory if you prepare a separate chroot with all the linux binaries it should work. You could even `vng --root-release oraculr --root /path/to/empty/dir` and vng will automatically create an Ubuntu chroot for you, that should be enough to have a working linux kernel testing environment on macOS.

Cool! I'll try it out. 

@SPYFF Yes, this builds the init program as it always has.

PR #179, opened 2024-10-09
Get (some) things working on macOS

tamird:
See individual commits.

With this I'm able to run `vng --root-release oraculr --root ../linux-chroot`
but the result is still unsatisfying: QEMU runs, but produces no output while
eating 9 CPU cores :(.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,6 @@ INSTALL_ARGS :=	# todo: add --break-system-packages if ubuntu
 # Get git version information for make install
 GIT_DESCRIBE := $(shell git describe --always --long --dirty)
 
-.PHONY: init
-init:
-	cd virtme_ng_init && cargo install --path . --root ../virtme/guest
-
 # see README.md '* Install from source'
 install: install_from_source
 install_from_source:
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import os
+import platform
 import sys
 import sysconfig
 from subprocess import check_call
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -3,9 +3,11 @@ INSTALL_ARGS :=	# todo: add --break-system-packages if ubuntu
 # Get git version information for make install
 GIT_DESCRIBE := $(shell git describe --always --long --dirty)
 
+TARGET := $(shell rustc -vV | grep 'host:' | cut -d' ' -f2 | sed 's/apple-darwin/unknown-linux-musl/')
+
 .PHONY: init
 init:
-	cd virtme_ng_init && cargo install --path . --root ../virtme/guest
+	cd virtme_ng_init && cargo install --path . --root ../virtme/guest --target $(TARGET) --config "target.$(TARGET).linker = \"rust-lld\""
```
arighi, line 10: This seems to break the build on Linux, since rust-lld isn't available. Maybe set the linker only if the binary exists in the PATH?

setup.py
```diff
@@ -39,10 +39,6 @@ def run(self):
         # Build virtme-ng-init
         if build_virtme_ng_init:
             check_call(["make", "init"])
-            check_call(
```
arighi, line 42: Can we make this conditional and execute it only when `platform.system() == "Linux"`  instead of removing it?

virtme/architectures.py
```diff
@@ -256,9 +256,7 @@ class Arch_aarch64(Arch):
     def __init__(self):
         Arch.__init__(self, "aarch64")
 
-        self.qemuname = "aarch64"
         self.linuxname = "arm64"
-        self.gccname = "aarch64"
```
arighi: I'm wondering if there's a better way to avoid removing gccname and qemuname, my concern is that we may break something if we drop these definitions...

arighi (commented):
Looks like a really good start!

I left a few comments, mostly to address potential breakage. Let me know if they make sense to you. Thanks for working on this!

Makefile
arighi: Hm... I don't see it in my system (tested both on Ubuntu 24.04 and CachyOS). Checking if it exists in the path shouldn't make the Makefile logic too complex, something like:
```
ifeq ($(shell command -v rustc-lld 2> /dev/null),)
    INSTALL_CMD = cd virtme_ng_init && cargo install --path . --root ../virtme/guest --target $(TARGET)
else
    INSTALL_CMD = cd virtme_ng_init && cargo install --path . --root ../virtme/guest --target $(TARGET) --config "target.$(TARGET).linker = \"rust-lld\""
endif
```

setup.py
arighi: Ok, I understand it's irrelevant in 99.999% of the cases, but what's the benefit of having an unstripped virtme-ng-init? Do you need this change because strip is not available on macOS? I'm just trying to understand the reason why you dropped this command.

arighi: Oh sorry, I didn't notice you already explained in the commit message. Alright, if it helps for portability let's just get rid of the strip command.

matttbe: On my side, this binary is 3.9M unstripped. Can we not simply drop `-s`?

Without `-s`, I have the exact same file that is produced on my side (556K, same checksum).

tamird: Sure, I can do that. I wonder if the difference is that mine is compiled for linux-musl rather than linux-gnu.

arighi (approved):
Looks good to me. I'll wait a bit before merging this, because I'd like to cut a new version before. I should be able to work on this tomorrow. Then I'll merge this and we can improve/fix in-tree if needed. Thanks!

PR #185, opened 2024-10-18
virtme-init: Enable lvm usage

marcosps:
Current /etc/lvm/ directories are restricted to root only (700), so just create an empty directory and bind mount over. This is enough to make commands like pvcreate to succeed.

It's the same commit as virtme-ng-init, but for virtme-init bash script. What do you think about this approach? Should we first check if the user is root before creating the bind mount?


arighi (approved):

arighi:
@marcosps now that I've merged the same change in virtme-ng-init, can you also update the submodule here in this PR, so the change will consistent? Thanks!

marcosps:
@arighi done, thanks lot!

PR #186, opened 2024-10-18
setup.py: Do not fail when argparse-manpage is not available

hramrach:
Building the man page is not strictly necesary for using the tool.

Add back the detection of presence argparse-manpage that makes it possible to build on older distributions that don't carry the module.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,10 @@ import subprocess
 import sys
 import sysconfig
 from argcomplete import shell_integration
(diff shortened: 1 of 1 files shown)


arighi (approved):

tamird:
This is a weird change. We have a build-time dependency on on argparse-manpage [here](https://github.com/arighi/virtme-ng/blob/39487f8fd61c20339e580c95f90fe9a9febe0f9a/pyproject.toml#L3) so this should be unreachable.

PR #194, opened 2024-11-27
 net: support multiple interfaces and different bridges

matttbe:
`virtme-run` supports multiple `--net` arguments to setup `virtio-net` devices, but the init program was only starting the DHCP client for one of them: the first one by alphabetical order.

Now, the DHCP client is started for each interface, so each of them can have an assigned IP address.

It is also now possible to specify the bridge interface, with `--net bridge=<br>` option, useful when there are multiple bridges that are created.

While at it, the udhcpc script no longer try to run a wrong command if no router is announced.

Note: this PR depends on https://github.com/arighi/virtme-ng-init/pull/12

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -121,8 +121,7 @@ def make_parser() -> argparse.ArgumentParser:
         action="append",
         const="user",
         nargs="?",
-        choices=["user", "bridge", "loop"],
-        help="Enable basic network access.",
+        help="Enable basic network access: user, bridge(=<br>), loop.",
     )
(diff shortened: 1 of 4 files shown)


matttbe:
Thank you for the review!

> Can you update the PR to include also the corresponding change in virtme-ng-init as well? (I've just merged it). Thanks!

I've just updated my branch.

PS: In the meantime, I managed to run `vng --snaps` as I mentioned in https://github.com/arighi/virtme-ng-init/pull/12, and it looks OK!

PR #195, opened 2024-11-27
net: option to force the MAC address

matttbe:
By default, QEMU sets the first NIC's MAC address to `52:54:00:12:34:56`, and increments the last octet for the next ones. When different VMs need to discuss between each others via the same bridge, that can cause conflicts.

A new option has been added to set a different MAC address, e.g.

    --net-mac-address 52:54:00:12:34:56

Which will assign 52:54:00:12:34:56 to the fist NIC, and increment the last octet for the next one, etc.

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -353,6 +353,13 @@ virtme-ng is based on virtme, written by Andy Lutomirski <luto@kernel.org>.
     )
 
     parser.add_argument(
+        "--net-mac-address",
+        action="store",
+        help="The MAC address to assign to the NIC interface, e.g. 52:54:00:12:34:56. "
+        + "The last octet will be incremented for the next network devices.",
+    )
+
+    parser.add_argument(
         "--disk",
         "-D",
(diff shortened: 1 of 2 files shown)


virtme_ng/run.py
```diff
@@ -352,6 +352,14 @@ def make_parser():
         help="Enable network access: user, bridge(=<br>), loop",
     )
 
+    parser.add_argument(
+        "--net-mac-address",
+        "-N",
```
arighi, line 357: Do we really need a short option here? Can we just use the long option like the virtme-run backend?

My concern is that we may want to use `-N` for something more "common" in the future (I'm not sure how many people will need to change the mac addr) and if we reserve this option here, then it's problematic to change it later.

arighi (approved):
Left a small comment, but overall LGTM!

virtme_ng/run.py
matttbe: Good point, I will remove this short option!

PR #197, opened 2024-11-29
feat: vsock support for remote console access

matttbe:
Having a remote console access can be helpful for different reasons, e.g. an easy way to have multiple terminals, not using a serial which can be slow to print a lot of text, etc.

With this new support, a user can simply use vng like before, and add the new `--vsock` option:

    vng --vsock (...)

Then in another terminal, it can connect to the existing VM:

    vng --vsock-connect

Advanced users can set a different CID with `--vsock-cid`, useful when multiple VMs are started in parallel.

It is also possible to change the command that is executed when connected inside the VM. By default, `bash -i` is used, but it is possible to pass something else, e.g.

    vng --vsock byobu

Or zsh, fish, tmux, etc. Or a script, for example to set env vars, change dir and set other dimensions, etc., e.g. starting the VM with:

    $ vng --vsock "${PWD}/console.sh"

... and connecting to it from another terminal with:

    $ read -r rows columns <<< "$(stty size)"
    $ cat <<-EOF > console.sh
        #! /bin/bash
        stty rows ${rows} columns ${columns}
        cd "\${virtme_chdir}"
        HOME=${HOME}
        byobu
      EOF
    $ chmod +x console.sh
    $ vng --vsock-connect

Note: on my side, 'bash' is a bit weird: I cannot see the command I'm typing. I didn't try to find a solution for a too long time as I'm usually not using bash.

Regarding the kernel config, two new config are now required: `VSOCKETS` and `VIRTIO_VSOCKETS`. They seem light enough.

While at it, fixed a typo in the README file with the --network option.

Depends-on: #195

Link: https://github.com/arighi/virtme-ng/discussions/151

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,6 +345,33 @@ Examples
    # vmlinux available in the system.
 ```
 
+ - Connect to a simple remote shell:
+```
+   # Start the vng instance with vsock support:
+   $ vng --vsock
+
+   # In a separate terminal run the following command to connect to a remote shell:
+   $ vng --vsock-connect
+```
+
+ - Connect to a remote shell with proper dimensions, env vars, and using 'Fish':
+```
+   # Start the vng instance with vsock support:
+   $ vng --vsock "${PWD}/console.sh"
+
+   # In a separate terminal run the following commands:
+   $ read -r rows columns <<< "$(stty size)"
+   $ cat <<-EOF > console.sh
+         #! /bin/bash
+         stty rows ${rows} columns ${columns}
+         cd "\${virtme_chdir}"
+         HOME=${HOME}
+         fish  # use use zsh, tmux, byobu, screen, etc.
+     EOF
+   $ chmod +x console.sh
+   $ vng --vsock-connect
+```
+
  - Run virtme-ng inside a docker container:
 ```
    $ docker run -it --privileged ubuntu:23.10 /bin/bash
@@ -449,7 +476,7 @@ Troubleshooting
   $ groups | grep "kvm\|libvirt"
 ```
 
- - When using `--net bridge` to create a bridged network in the guest you
+ - When using `--network bridge` to create a bridged network in the guest you
    may get the following error:
 ```
   ...
--- a/virtme/commands/configkernel.py
+++ b/virtme/commands/configkernel.py
@@ -159,6 +159,9 @@ _GENERIC_CONFIG = [
     "CONFIG_ZONE_DEVICE=y",
     "CONFIG_FUSE_FS=y",
     "CONFIG_VIRTIO_FS=y",
+    "##: vsock support",
+    "CONFIG_VSOCKETS=y",
+    "CONFIG_VIRTIO_VSOCKETS=y",
 ]
 
 _GENERIC_CONFIG_OPTIONAL = [
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -131,6 +131,25 @@ def make_parser() -> argparse.ArgumentParser:
         + "The last octet will be incremented for the next network devices.",
     )
     g.add_argument(
+        "--vsock",
+        action="store",
+        default=None,
+        help="Enable a VSock to communicate from the host to the device and "
+        + "execute the specified command.",
+    )
+    g.add_argument(
+        "--vsock-cid",
+        action="store",
+        type=int,
+        default=3,
+        help="CID for the VSock.",
+    )
+    g.add_argument(
+        "--vsock-connect",
+        action="store_true",
+        help="Connect to a VM using VSock.",
+    )
+    g.add_argument(
         "--balloon",
         action="store_true",
         help="Allow the host to ask the guest to release memory.",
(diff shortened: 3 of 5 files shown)


matttbe:
Note: because I added the options after others from #195, this PR depends on that. But I can change if needed.

README.md
```diff
@@ -345,6 +345,33 @@ Examples
...
+   $ vng --vsock "${PWD}/console.sh"
+
+   # In a separate terminal run the following commands:
+   $ read -r rows columns <<< "$(stty size)"
+   $ cat <<-EOF > console.sh
+         #! /bin/bash
+         stty rows ${rows} columns ${columns}
+         cd "\${virtme_chdir}"
+         HOME=${HOME}
+         fish  # use use zsh, tmux, byobu, screen, etc.
```
arighi, line 369: I was wondering if we should provide a helper script in `virtme/guest`, so when `--vsock` is specified without any argument we can use this default shell wrapper. Something like `virtme/guest/virtme-vsock-console`.

And maybe this wrapper should look like the following:

```
#!/bin/bash

read -r rows columns <<< "$(stty size)"
stty rows ${rows} columns ${columns}
HOME=$(getent passwd ${virtme_user} | cut -d: -f6)
cd ${virtme_chdir}
stty echo
exec su - ${virtme_user}
```

What do you think?

arighi (approved):
I just left a comment about providing a helper script to initialize the console session, but overall it looks great, I love this!

Feel free to merge this as is, we can improve it later in tree.

README.md
matttbe: I think that would be a good idea.

I didn't try to do this for two main reasons:
- We need to make sure the script is mounted in the VM. I guess we can always add another mount point, or use the guest one (but the user might not have the right to modify it)
- We need a way to pass info about the terminal size (`stty size`) when `vng --vsock-connect` is launched: so after the VM has been started, and from a different terminal. For SSH, they have a control channel to pass all that, not here.

It is not impossible: I guess we could have a dedicated mount point, using a temp file on the host that would be populated when launching `vng --vsock-connect`. WDYT?

Should we eventually already merge this while we are discussing this, and improve this later in another PR?

About the wrapper:

> `read -r rows columns <<< "$(stty size)"`

This needs to be done from the host. Inside, I guess stty size will always return 80x24 (or the one of the terminal which launched the VM?), no?

> `HOME=$(getent passwd ${virtme_user} | cut -d: -f6)`

Good idea, but we should use: `${virtme_user:-root}` if root is used I suppose. Same later.

> `stty echo`

Nice, it looks like it is fixing the issue I had with `bash`! Is it why you added it?

> `exec su - ${virtme_user}`

We need to do the `cd ${virtme_chdir}` after that one. (And also use `${virtme_user:-root}`)

arighi: > I think that would be a good idea.
> 
> I didn't try to do this for two main reasons:
> 
> * We need to make sure the script is mounted in the VM. I guess we can always add another mount point, or use the guest one (but the user might not have the right to modify it)

If we put the script in the guest_tools path they'll always be available (see `virtme/guest`, it's the same dir where the init script/binary is also placed), so no need to add another mount point.

> * We need a way to pass info about the terminal size (`stty size`) when `vng --vsock-connect` is launched: so after the VM has been started, and from a different terminal. For SSH, they have a control channel to pass all that, not here.

Ah good point about the terminal size. I'm not sure how to fix this, I have to think about it.

> 
> It is not impossible: I guess we could have a dedicated mount point, using a temp file on the host that would be populated when launching `vng --vsock-connect`. WDYT?

Honestly I'd prefer to not add a dedicated mount point for this, in theory we already have a lot of shared dirs between host and guest. Or maybe we can introduce a shared rw mount point to exchange info between host and guest... and use it also for other purposes, something similar to the guest_tools path, for example guest_data? We can probably introduce this later and ignore the terminal size issue for now...

> 
> Should we eventually already merge this while we are discussing this, and improve this later in another PR?

Yes, sounds good to me. We can definitely merge as it and improve it later in-tree.

> 
> About the wrapper:
> 
> > `read -r rows columns <<< "$(stty size)"`
> 
> This needs to be done from the host. Inside, I guess stty size will always return 80x24 (or the one of the terminal which launched the VM?), no?

Yep, you're right.

> 
> > `HOME=$(getent passwd ${virtme_user} | cut -d: -f6)`
> 
> Good idea, but we should use: `${virtme_user:-root}` if root is used I suppose. Same later.

Correct.

> 
> > `stty echo`
> 
> Nice, it looks like it is fixing the issue I had with `bash`! Is it why you added it?

Yep, this fixed bash for me.

> 
> > `exec su - ${virtme_user}`
> 
> We need to do the `cd ${virtme_chdir}` after that one. (And also use `${virtme_user:-root}`)

Correct, and maybe it's fine to just switch to the user's $HOME. At the end it'd be like a separate login (similar to ssh), so we don't have to honor `${virtme_chdir}`.

matttbe: > > I think that would be a good idea.
> > I didn't try to do this for two main reasons:
> > 
> > * We need to make sure the script is mounted in the VM. I guess we can always add another mount point, or use the guest one (but the user might not have the right to modify it)
> 
> If we put the script in the guest_tools path they'll always be available (see `virtme/guest`, it's the same dir where the init script/binary is also placed), so no need to add another mount point.

Indeed, but if we need to modify this script on the host (for the terminal size), that will be an issue :-/

> > It is not impossible: I guess we could have a dedicated mount point, using a temp file on the host that would be populated when launching `vng --vsock-connect`. WDYT?
> 
> Honestly I'd prefer to not add a dedicated mount point for this, in theory we already have a lot of shared dirs between host and guest. Or maybe we can introduce a shared rw mount point to exchange info between host and guest... and use it also for other purposes, something similar to the guest_tools path, for example guest_data? We can probably introduce this later and ignore the terminal size issue for now...

Yes, that's a good idea. But we would need a way to avoid conflicts when multiple VMs are started in parallel, and a way to find this directory when `vng --vsock-connect` is launched.

During a break, I started to look for a solution. I can commit that. It introduces a new mount point if root is not mounted. Maybe easier to discuss that around the code.

matttbe: I just opened #198, just in case you still want to merge the first part now :)

PR #200, opened 2024-12-03
make: build virtme-ng-init by default

matttbe:
As mentioned in the README file, the `make` command should only build `virtme-ng-init`. Building is what `make` usually does, and not installing stuff.

Since commit 85a9b42 ("Make it build on macOS"), `make` was building `virtme-ng-init`, but it was also trying to install it, what `make install` does.

Instead of directly calling `cargo` like it was the case before, the `build` part of `setup.py` is called. This will generate files in the build dir that are probably not needed, but it is quick, and it will also build `virtme-ng-init` properly, what we want. Cc: @tamird.

While at it:

- Mention all targets that are not linked to files in `.PHONY`.

- Only generate the init file when needed: when the `.rs` files have been modified.

- Initialise the submodule if the `.rs` files are not present.

- A new `clean` target can help to force rebuilding (or just to clean, even if it looks like it missed many generated files...).


tamird:
> Since commit https://github.com/arighi/virtme-ng/commit/85a9b424c51a3d18d3b9b317f24984a09f935b87 ("Make it build on macOS"), make was building virtme-ng-init, but it was also trying to install it, what make install does.

Are you referring to the use of `cargo install`? Because that isn't "installing" in the way you're implying because of the `--root` argument which causes `cargo` to "install" into the package, not system-wide destination.

I'm not really sure what problem this PR is solving but this description is misleading. Please consider describing the true motivation.

matttbe:
Note: I used `python3 setup.py build`, but maybe there is a better command to use?

arighi:
> Note: I used `python3 setup.py build`, but maybe there is a better command to use?

We could detect python3 || python, but I wouldn't over-complicate things here, running python3 setup.py build seems good enough.

matttbe:
> > Note: I used `python3 setup.py build`, but maybe there is a better command to use?
> 
> We could detect python3 || python, but I wouldn't over-complicate things here, running python3 setup.py build seems good enough.

Good point. We can indeed change that later. Are there still some distributions not having a `python3` binary, only `python`? I remembered issues a few years ago when using `python`, but not sure about `python3`.

arighi:
> Good point. We can indeed change that later. Are there still some distributions not having a `python3` binary, only `python`? I remembered issues a few years ago when using `python`, but not sure about `python3`.

Not that I'm aware of, I think all the recent distro have python3.

PR #204, opened 2024-12-06
vsock: connect: support no tty, command with right user and no 'su'

matttbe:
Here are a few small improvements for the new `--vsock-connect` feature:
- support environments without tty, e.g. CI.
- execute commands with the right user, not root if the VM was launched with a different user.
- use `su` only if needed: to save ~0.5 sec when executing a command with root.
- `--user` is used if defined
- `--pwd` and `--cwd` are used if defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,7 +345,7 @@ Examples
    # vmlinux available in the system.
 ```
 
- - Connect to a simple remote shell:
+ - Connect to a simple remote shell (`socat` is required):
 ```
    # Start the vng instance with vsock support:
    $ vng --vsock
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -887,7 +887,10 @@ def do_it() -> int:
             socat_in = '-'
         socat_out = f'VSOCK-CONNECT:{args.vsock_cid}:1024'
 
-        cmd = args.vsock_connect if args.vsock_connect else 'su ${virtme_user:-root}'
+        cmd = ''
+        if args.vsock_connect:
+            exec_escaped = args.vsock_connect.replace('"', '\\"')
+            cmd = f' -c "{exec_escaped}"'
 
         with open(vsock_script_path, 'w', encoding="utf-8") as file:
             print((
@@ -896,7 +899,7 @@ def do_it() -> int:
                 f'{stty}\n'
                 'HOME=$(getent passwd ${virtme_user:-root} | cut -d: -f6)\n'
                 'cd ${virtme_chdir:+"${virtme_chdir}"}\n'
-                f'exec {cmd}\n'
+                'exec su ${virtme_user}' f'{cmd}\n'
                 '}\n'
                 'main'  # use a function to avoid issues when the script is modified
             ), file=file)
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -889,6 +889,13 @@ def do_it() -> int:
 
         user = args.user if args.user else '${virtme_user:-root}'
 
+        if args.pwd:
+            cwd = os.path.relpath(os.getcwd(), args.root)
+        elif args.cwd is not None:
+            cwd = os.path.relpath(args.cwd, args.root)
+        else:
+            cwd = '${virtme_chdir:+"${virtme_chdir}"}'
+
         # use 'su' only if needed: another use, or to get a prompt
         cmd = f'if [ "{user}" != "root" ]; then\n' + \
               f'  exec su "{user}"'
@@ -908,7 +915,7 @@ def do_it() -> int:
                 'main() {\n'
                 f'{stty}\n'
                 f'HOME=$(getent passwd "{user}" | cut -d: -f6)\n'
-                'cd ${virtme_chdir:+"${virtme_chdir}"}\n'
+                f'cd {cwd}\n'
                 f'{cmd}\n'
                 '}\n'
                 'main'  # use a function to avoid issues when the script is modified
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -878,9 +878,14 @@ def do_it() -> int:
 
     if args.vsock_connect is not None:
         try:
+            # with tty support
             (cols, rows) = os.get_terminal_size()
+            stty = f'stty rows {rows} cols {cols} iutf8 echo'
+            socat_in = f'file:{os.ttyname(sys.stdin.fileno())},raw,echo=0'
         except OSError:
-            cols, rows = (80, 24)
+            stty = ''
+            socat_in = '-'
+        socat_out = f'VSOCK-CONNECT:{args.vsock_cid}:1024'
 
         cmd = args.vsock_connect if args.vsock_connect else 'su ${virtme_user:-root}'
 
@@ -888,7 +893,7 @@ def do_it() -> int:
             print((
                 '#! /bin/bash\n'
                 'main() {\n'
-                f'stty rows {rows} cols {cols} iutf8 echo\n'
+                f'{stty}\n'
                 'HOME=$(getent passwd ${virtme_user:-root} | cut -d: -f6)\n'
                 'cd ${virtme_chdir:+"${virtme_chdir}"}\n'
                 f'exec {cmd}\n'
@@ -897,10 +902,7 @@ def do_it() -> int:
             ), file=file)
         os.chmod(vsock_script_path, 0o755)
 
-        tty = os.ttyname(sys.stdin.fileno())
-        command = ['socat', f'file:{tty},raw,echo=0',
-                   f'VSOCK-CONNECT:{args.vsock_cid}:1024']
-        os.execvp('socat', command)
+        os.execvp('socat', ['socat', socat_in, socat_out])
         sys.exit(0)
 
     arch = architectures.get(args.arch)
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -887,10 +887,18 @@ def do_it() -> int:
             socat_in = '-'
         socat_out = f'VSOCK-CONNECT:{args.vsock_cid}:1024'
 
-        cmd = ''
+        # use 'su' only if needed: another use, or to get a prompt
+        cmd = 'if [ "${virtme_user:-root}" != "root" ]; then\n' + \
+              '  exec su ${virtme_user}'
         if args.vsock_connect:
             exec_escaped = args.vsock_connect.replace('"', '\\"')
-            cmd = f' -c "{exec_escaped}"'
+            cmd += f' -c "{exec_escaped}"' + \
+                   '\nelse\n' + \
+                   f'  {args.vsock_connect}\n'
+        else:
+            cmd += '\nelse\n' + \
+                   '  exec su\n'
+        cmd += 'fi'
 
         with open(vsock_script_path, 'w', encoding="utf-8") as file:
             print((
@@ -899,7 +907,7 @@ def do_it() -> int:
                 f'{stty}\n'
                 'HOME=$(getent passwd ${virtme_user:-root} | cut -d: -f6)\n'
                 'cd ${virtme_chdir:+"${virtme_chdir}"}\n'
-                'exec su ${virtme_user}' f'{cmd}\n'
+                f'{cmd}\n'
                 '}\n'
                 'main'  # use a function to avoid issues when the script is modified
             ), file=file)
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -887,9 +887,11 @@ def do_it() -> int:
             socat_in = '-'
         socat_out = f'VSOCK-CONNECT:{args.vsock_cid}:1024'
 
+        user = args.user if args.user else '${virtme_user:-root}'
+
         # use 'su' only if needed: another use, or to get a prompt
-        cmd = 'if [ "${virtme_user:-root}" != "root" ]; then\n' + \
-              '  exec su ${virtme_user}'
+        cmd = f'if [ "{user}" != "root" ]; then\n' + \
+              f'  exec su "{user}"'
         if args.vsock_connect:
             exec_escaped = args.vsock_connect.replace('"', '\\"')
             cmd += f' -c "{exec_escaped}"' + \
@@ -905,7 +907,7 @@ def do_it() -> int:
                 '#! /bin/bash\n'
                 'main() {\n'
                 f'{stty}\n'
-                'HOME=$(getent passwd ${virtme_user:-root} | cut -d: -f6)\n'
+                f'HOME=$(getent passwd "{user}" | cut -d: -f6)\n'
                 'cd ${virtme_chdir:+"${virtme_chdir}"}\n'
                 f'{cmd}\n'
                 '}\n'
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1498,7 +1498,7 @@ def do_it() -> int:
             return 1
         kernelargs.append("virtme_chdir=%s" % rel_cwd)
 
-    if args.user:
+    if args.user and args.user != "root":
         kernelargs.append("virtme_user=%s" % args.user)
 
     if args.nvgpu:
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -136,7 +136,7 @@ def make_parser() -> argparse.ArgumentParser:
         nargs="?",
         metavar="COMMAND",
         const="",
-        help="Enable a VSock to communicate from the host to the device. "
+        help="Enable a VSock to communicate from the host to the device, 'socat' is required. "
         + "An argument can be optionally specified to start a different command.",
     )
     g.add_argument(
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -365,7 +365,7 @@ virtme-ng is based on virtme, written by Andy Lutomirski <luto@kernel.org>.
         nargs="?",
         metavar="COMMAND",
         const="",
-        help="Enable a VSock to communicate from the host to the device. "
+        help="Enable a VSock to communicate from the host to the device, 'socat' is required. "
         + "An argument can be optionally specified to start a different command.",
     )
 
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -852,16 +852,14 @@ class KernelSource:
         # sessions is current user.
         #
         # NOTE: graphic sessions are considered interactive.
+        self.virtme_param["user"] = ""
         if args.exec and not args.graphics:
             self.virtme_param["user"] = ""
         else:
             self.virtme_param["user"] = "--user " + get_username()
         # Override default user, if specified by the --user argument.
         if args.user is not None:
-            if args.user != "root":
-                self.virtme_param["user"] = "--user " + args.user
-            else:
-                self.virtme_param["user"] = ""
+            self.virtme_param["user"] = "--user " + args.user
 
     def _get_virtme_arch(self, args):
         if args.arch is not None:


arighi (approved):

arighi:
@matttbe side note (not strictly related to this PR), I was considering to add also a way to ssh into the vng instance (similar to the vsock connect), because some people that I know have a CI workflow where they need to ssh into the instance, run some commands and scp the final output on the host.

I already provided them an unofficial ad-hoc solution (with a script in guest_tools dir to start sshd), but I'd like to integrate this properly. So, I was wondering if we could change the syntax to have something like this (suggestions are welcome):
```
$ vng --server vsock|ssh [--server-id CID] [--server-cmd COMMAND]
$ vng --connect vsock|ssh [--connect-cmd COMMAND]
```

This would unify both the ssh and vsock way to connect to a vng instance, I'm not sure what the `--server-id` would be for ssh (maybe change the hostname? Or we can just ignore it for now and use it only with vsock).

What do you think?

matttbe:
@arighi thank you for the review!

I just added an extra commit to mention `socat` is required. Still OK for you to merge?

> I was considering to add also a way to ssh into the vng instance (similar to the vsock connect), because some people that I know have a CI workflow where they need to ssh into the instance, run some commands and scp the final output on the host.

Good idea!

> I already provided them an unofficial ad-hoc solution (with a script in guest_tools dir to start sshd), but I'd like to integrate this properly.

Out of curiosity, for this to work, what network connectivity do you put in place? Do you require a `bridge` netdev to be up, or a `user` one with port forwarding (`hostfwd`? then the port on the host might need to be adapted if multiple VMs are started), or something else?

VSOCK can also be used for the connection from the host to the VM. With the help of `socat`, it is possible to proxy the connection TCP to VSOCK and VSOCK to TCP. `systemd-ssh-proxy` could also be used.

> So, I was wondering if we could change the syntax to have something like this (suggestions are welcome):
> 
> ```
> $ vng --server vsock|ssh [--server-id CID] [--server-cmd COMMAND]
> $ vng --connect vsock|ssh [--connect-cmd COMMAND]
> ```
> 
> This would unify both the ssh and vsock way to connect to a vng instance, I'm not sure what the `--server-id` would be for ssh (maybe change the hostname? Or we can just ignore it for now and use it only with vsock).
> 
> What do you think?

Good idea! Maybe there is no need to call it `vsock`, it could be called `console`.

For the server and client sides, there is a need to specify a port (ssh, or a hostname?) or CID (vsock) if multiple VMs are used in parallel.

It might be good to use `console` by default, e.g. if only `--server` or `--client` is passed: with this mode, there are fewer dependences, and it should just work if `socat` is available.

With ssh, it might be could to have a way to print the ssh command that will be used, for advanced use on the client side (or to be used with scp, etc.)

arighi:
> @arighi thank you for the review!
> 
> I just added an extra commit to mention that `socat` is required. Still OK for you to merge?

Looks good, feel free to merge!

> 
> > I was considering to add also a way to ssh into the vng instance (similar to the vsock connect), because some people that I know have a CI workflow where they need to ssh into the instance, run some commands and scp the final output on the host.
> 
> Good idea!
> 
> > I already provided them an unofficial ad-hoc solution (with a script in guest_tools dir to start sshd), but I'd like to integrate this properly.
> 
> Out of curiosity, for this to work, what network connectivity do you put in place? Do you require a `bridge` netdev to be up, or a `user` one with port forwarding (`hostfwd`? Then the port on the host might need to be adapted if multiple VMs are started), or something else?

User networking with port forwarding. Basically I did this (needs some adjustment with `--disable-microvm`):
```
$ vng -r --qemu-opts="--netdev user,id=net0,hostfwd=tcp::8889-:22 -device virtio-net-device,netdev=net0"
```
Then in the guest I run this script:
```
#!/bin/bash

sudo ip link set ens2 up
sudo ip addr add 10.0.2.15/24 dev ens2
sudo ip route add default via 10.0.2.2
echo "nameserver 8.8.8.8" | sudo tee /etc/resolv.conf
cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
sudo mkdir -p /run/sshd
sudo rm -f /var/run/nologin
sudo /usr/sbin/sshd -h ~/.ssh/id_rsa
```

And from the host I can `ssh -p 8889 localhost`.

It works, but I wanted to make all this stuff a bit nicer, more accessible and properly integrated in vng.

> 
> VSOCK can also be used for the connection from the host to the VM. With the help of `socat`, it is possible to proxy the connection TCP to VSOCK and VSOCK to TCP. `systemd-ssh-proxy` could also be used.
> 
> > So, I was wondering if we could change the syntax to have something like this (suggestions are welcome):
> > ```
> > $ vng --server vsock|ssh [--server-id CID] [--server-cmd COMMAND]
> > $ vng --connect vsock|ssh [--connect-cmd COMMAND]
> > ```
> >   
> > This would unify both the ssh and vsock way to connect to a vng instance, I'm not sure what the `--server-id` would be for ssh (maybe change the hostname? Or we can just ignore it for now and use it only with vsock).
> > What do you think?
> 
> Good idea! Maybe there is no need to call it `vsock`, it could be called `console`.

Using `console` seems like  good idea, it's much more intuitive.

> 
> For the server and client sides, there might be a need to specify a port (ssh, or a hostname?) or CID (vsock) if multiple VMs are used in parallel.

Right with `--server CID` in the ssh case CID could be the port (maybe we can just call this ID). 

> 
> It might be good to use `console` (`vsock`) by default, e.g. if only `--server` or `--client` is passed: with this mode, there are fewer dependences, and it should just work if `socat` is available.

I agree, `--server` and `--connect` should use vsock by default.

> 
> With ssh, it might be could to have a way to print the ssh command that will be used, for advanced use on the client side (or to be used with scp, etc.)

Also good idea!

matttbe:
> > @arighi thank you for the review!
> > I just added an extra commit to mention that `socat` is required. Still OK for you to merge?
> 
> Looks good, feel free to merge!

Thank you!

> > > I was considering to add also a way to ssh into the vng instance (similar to the vsock connect), because some people that I know have a CI workflow where they need to ssh into the instance, run some commands and scp the final output on the host.
> > 
> > 
> > Good idea!
> > > I already provided them an unofficial ad-hoc solution (with a script in guest_tools dir to start sshd), but I'd like to integrate this properly.
> > 
> > 
> > Out of curiosity, for this to work, what network connectivity do you put in place? Do you require a `bridge` netdev to be up, or a `user` one with port forwarding (`hostfwd`? Then the port on the host might need to be adapted if multiple VMs are started), or something else?
> 
> User networking with port forwarding. Basically I did this (needs some adjustment with `--disable-microvm`):
> 
> ```
> $ vng -r --qemu-opts="--netdev user,id=net0,hostfwd=tcp::8889-:22 -device virtio-net-device,netdev=net0"
> ```
> 
> Then in the guest I run this script:
> 
> ```
> #!/bin/bash
> 
> sudo ip link set ens2 up
> sudo ip addr add 10.0.2.15/24 dev ens2
> sudo ip route add default via 10.0.2.2
> echo "nameserver 8.8.8.8" | sudo tee /etc/resolv.conf
> cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
> sudo mkdir -p /run/sshd
> sudo rm -f /var/run/nologin
> sudo /usr/sbin/sshd -h ~/.ssh/id_rsa
> ```

How did you make it run automatically?
It might be good to have an option to run something as a last step before giving the hand back to the user. Something like `--exec` / `--script-sh`, bit without the exit part. Could be good to experiment things, or to setup an environment before starting a script. But maybe such option already exists?

> And from the host I can `ssh -p 8889 localhost`.
> 
> It works, but I wanted to make all this stuff a bit nicer, more accessible and properly integrated in vng.

I see, looks like a good way for that.

I guess the 4 first lines of this script would be done by the init script simply by setting `virtme.dhcp`.

> > VSOCK can also be used for the connection from the host to the VM. With the help of `socat`, it is possible to proxy the connection TCP to VSOCK and VSOCK to TCP. `systemd-ssh-proxy` could also be used.
> > > So, I was wondering if we could change the syntax to have something like this (suggestions are welcome):
> > > ```
> > > $ vng --server vsock|ssh [--server-id CID] [--server-cmd COMMAND]
> > > $ vng --connect vsock|ssh [--connect-cmd COMMAND]
> > > ```
> > >
> > > This would unify both the ssh and vsock way to connect to a vng instance, I'm not sure what the `--server-id` would be for ssh (maybe change the hostname? Or we can just ignore it for now and use it only with vsock).
> > > What do you think?
> > 
> > 
> > Good idea! Maybe there is no need to call it `vsock`, it could be called `console`.
> 
> Using `console` seems like good idea, it's much more intuitive.
> 
> > For the server and client sides, there might be a need to specify a port (ssh, or a hostname?) or CID (vsock) if multiple VMs are used in parallel.
> 
> Right with `--server CID` in the ssh case CID could be the port (maybe we can just call this ID).

Yes, ID sounds good! (Or port, or both with an "alias"?)

> > It might be good to use `console` (`vsock`) by default, e.g. if only `--server` or `--client` is passed: with this mode, there are fewer dependences, and it should just work if `socat` is available.
> 
> I agree, `--server` and `--connect` should use vsock by default.
> 
> > With ssh, it might be could to have a way to print the ssh command that will be used, for advanced use on the client side (or to be used with scp, etc.)
> 
> Also good idea!

Great!

Do you plan to rename these `--vsock` options when introducing the SSH ones, or do you prefer if I do this part?

arighi:
> > > > I was considering to add also a way to ssh into the vng instance (similar to the vsock connect), because some people that I know have a CI workflow where they need to ssh into the instance, run some commands and scp the final output on the host.
> > > 
> > > 
> > > Good idea!
> > > > I already provided them an unofficial ad-hoc solution (with a script in guest_tools dir to start sshd), but I'd like to integrate this properly.
> > > 
> > > 
> > > Out of curiosity, for this to work, what network connectivity do you put in place? Do you require a `bridge` netdev to be up, or a `user` one with port forwarding (`hostfwd`? Then the port on the host might need to be adapted if multiple VMs are started), or something else?
> > 
> > 
> > User networking with port forwarding. Basically I did this (needs some adjustment with `--disable-microvm`):
> > ```
> > $ vng -r --qemu-opts="--netdev user,id=net0,hostfwd=tcp::8889-:22 -device virtio-net-device,netdev=net0"
> > ```
> >   
> > Then in the guest I run this script:
> > ```
> > #!/bin/bash
> > 
> > sudo ip link set ens2 up
> > sudo ip addr add 10.0.2.15/24 dev ens2
> > sudo ip route add default via 10.0.2.2
> > echo "nameserver 8.8.8.8" | sudo tee /etc/resolv.conf
> > cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
> > sudo mkdir -p /run/sshd
> > sudo rm -f /var/run/nologin
> > sudo /usr/sbin/sshd -h ~/.ssh/id_rsa
> > ```
> 
> How did you make it run automatically? It might be good to have an option to run something as a last step before giving the hand back to the user. Something like `--exec` / `--script-sh`, bit without the exit part. Could be good to experiment things, or to setup an environment before starting a script. But maybe such option already exists?

I'm just starting it manually for now. :)

About having a mechanism to automatically start init scripts at boot, it has been considered in the past. Personally I'm not a big fan of it, because it can slow down the boot time if abused (they need to placed in guest_tools_dir or a similar path exported via 9pfs, that's not really optimal in terms of performance).

I'd rather hard-code the logic in virtme-ng-init / virtme-init for now with a script in `guest_tools_path` that is executed only when the corresponding feature is enabled  (like `virtme-udhcpc-script`, `virtme-sound-script`, etc.), so that people don't abuse the init script mechanism.

But I don't have a strong opinion on this and I can change my mind if it makes our life easier (like if we start to have too many custom init scripts at some point).

> 
> > And from the host I can `ssh -p 8889 localhost`.
> > It works, but I wanted to make all this stuff a bit nicer, more accessible and properly integrated in vng.
> 
> I see, looks like a good way for that.
> 
> I guess the 4 first lines of this script would be done by the init script simply by setting `virtme.dhcp`.

Ah good point! That seems better.

> > > VSOCK can also be used for the connection from the host to the VM. With the help of `socat`, it is possible to proxy the connection TCP to VSOCK and VSOCK to TCP. `systemd-ssh-proxy` could also be used.
> > > > So, I was wondering if we could change the syntax to have something like this (suggestions are welcome):
> > > > ```
> > > > $ vng --server vsock|ssh [--server-id CID] [--server-cmd COMMAND]
> > > > $ vng --connect vsock|ssh [--connect-cmd COMMAND]
> > > > ```
> > > >   
> > > > This would unify both the ssh and vsock way to connect to a vng instance, I'm not sure what the `--server-id` would be for ssh (maybe change the hostname? Or we can just ignore it for now and use it only with vsock).
> > > > What do you think?
> > > 
> > > 
> > > Good idea! Maybe there is no need to call it `vsock`, it could be called `console`.
> > 
> > 
> > Using `console` seems like good idea, it's much more intuitive.
> > > For the server and client sides, there might be a need to specify a port (ssh, or a hostname?) or CID (vsock) if multiple VMs are used in parallel.
> > 
> > 
> > Right with `--server CID` in the ssh case CID could be the port (maybe we can just call this ID).
> 
> Yes, ID sounds good! (Or port, or both with an "alias"?)

PORT is probably a more familiar name (and in the vsock case it can be interpreted as a CID).

> > > It might be good to use `console` (`vsock`) by default, e.g. if only `--server` or `--client` is passed: with this mode, there are fewer dependences, and it should just work if `socat` is available.
> > 
> > 
> > I agree, `--server` and `--connect` should use vsock by default.
> > > With ssh, it might be could to have a way to print the ssh command that will be used, for advanced use on the client side (or to be used with scp, etc.)
> > 
> > 
> > Also good idea!
> 
> Great!
> 
> Do you plan to rename these `--vsock` options when introducing the SSH ones, or do you prefer if I do this part?

I think it'd be nice to rename these options before cutting a new release, so that we don't break the API in the "officially" released versions.

Feel free to go ahead and do the renaming if you have time, otherwise I'll take a look later this week and see if I can also send a PR for the sshd change (I just need to polish the code a little bit).

Thanks!

arighi:
> I think it is good to have virtme(-ng)-init preparing the environment for most use-cases. But it looks impossible to cover all use-cases, and executing a script / binary before giving the control to the user might be handy.
> 
> A recent example: I had to launch two VMs in parallel, communicating via the same bridge. Inside the VMs, at startup, and before doing anything else (e.g. launching tests using an external tests suite), I had to set the network up, something specific to my environment (so not handled by an external tests suite). I started with a simple script similar to yours, setting the IP manually, but it was "annoying" to launch it manually each time each VM started. I ended up starting a DHCP server with static leases on the host, and modifying virtme(-ng)-init to support that, but that's probably not something everybody will do :) It would have been useful and enough to run my small script, then letting me type the commands I wanted.
> 
> Even when everything is automated, I think it might be good to have a "generic" script that would be executed at startup, to set the environment up, before the rest.

I'm all for it, as long as we make it optional, so that we don't regress the boot time. Maybe we could have something like `--init-scripts DIR` and virtme-ng-init / virtme-init will just run all the scripts in `DIR` before giving control to the user session. If `--init-script DIR` isn't specified, we won't do anything, so we shouldn't regress the boot time.

> > I think it'd be nice to rename these options before cutting a new release, so that we don't break the API in the "officially" released versions.
> 
> Indeed.
> 
> > Feel free to go ahead and do the renaming if you have time, otherwise I'll take a look later this week and see if I can also send a PR for the sshd change (I just need to polish the code a little bit).
> 
> I will check if I can quickly do that.

Take your time. If you're willing to look at this I can wait for your change and focus at polishing the sshd change.

Thanks!

matttbe:
> > I think it is good to have virtme(-ng)-init preparing the environment for most use-cases. But it looks impossible to cover all use-cases, and executing a script / binary before giving the control to the user might be handy.
> > A recent example: I had to launch two VMs in parallel, communicating via the same bridge. Inside the VMs, at startup, and before doing anything else (e.g. launching tests using an external tests suite), I had to set the network up, something specific to my environment (so not handled by an external tests suite). I started with a simple script similar to yours, setting the IP manually, but it was "annoying" to launch it manually each time each VM started. I ended up starting a DHCP server with static leases on the host, and modifying virtme(-ng)-init to support that, but that's probably not something everybody will do :) It would have been useful and enough to run my small script, then letting me type the commands I wanted.
> > Even when everything is automated, I think it might be good to have a "generic" script that would be executed at startup, to set the environment up, before the rest.
> 
> I'm all for it, as long as we make it optional, so that we don't regress the boot time. Maybe we could have something like `--init-scripts DIR` and virtme-ng-init / virtme-init will just run all the scripts in `DIR` before giving control to the user session. If `--init-script DIR` isn't specified, we won't do anything, so we shouldn't regress the boot time.

Yes, optional is what I had in mind. We could indeed pass a directory or a file, and in this case, the init script will execute everything before continuing like before.

> > > I think it'd be nice to rename these options before cutting a new release, so that we don't break the API in the "officially" released versions.
> > 
> > 
> > Indeed.
> > > Feel free to go ahead and do the renaming if you have time, otherwise I'll take a look later this week and see if I can also send a PR for the sshd change (I just need to polish the code a little bit).
> > 
> > 
> > I will check if I can quickly do that.
> 
> Take your time. If you're willing to look at this I can wait for your change and focus at polishing the sshd change.
> 
> Thanks!

Thanks! Renaming should not be difficult, and I can already adapt my other scripts using them! :)

matttbe:
> So, I was wondering if we could change the syntax to have something like this (suggestions are welcome):
> 
> ```
> $ vng --server vsock|ssh [--server-id CID] [--server-cmd COMMAND]
> $ vng --connect vsock|ssh [--connect-cmd COMMAND]
> ```

I was looking at this new version:

```
$ vng --server [console|ssh] [--server-port CID] [--server-cmd COMMAND]
$ vng --connect [console|ssh] [--client-port CID] [--client-cmd COMMAND]
```

Because we need the same options for the client and server (port/cmd) and they cannot be used at the same time, should we use `--port` and `--XXX-cmd`? Any idea for a good name for the cmd one? :)

matttbe:
> Any idea for a good name for the cmd one? :)

`--remote-cmd`? `--server-client-cmd`?

arighi:
> > So, I was wondering if we could change the syntax to have something like this (suggestions are welcome):
> > ```
> > $ vng --server vsock|ssh [--server-id CID] [--server-cmd COMMAND]
> > $ vng --connect vsock|ssh [--connect-cmd COMMAND]
> > ```
> 
> I was looking at this new version:
> 
> ```
> $ vng --server [console|ssh] [--server-port CID] [--server-cmd COMMAND]
> $ vng --connect [console|ssh] [--client-port CID] [--client-cmd COMMAND]

Oh I like this more, console sounds more intuitive (and we can mention that it's implemented using vsock in the command description).

> ```
> 
> Because we need the same options for the client and server (port/cmd) and they cannot be used at the same time, should we use `--port` and `--XXX-cmd`? Any idea for a good name for the cmd one? :)

Hm... what if we make `--server` and `--connect` mutually exclusive and then we can just have `--port` for both (we just need to check if `--server` or `--connect` has been specified).

About the client command I'd do something similar, we could just have a `--command CMD`, even if it's a bit misleading as it may be interpreted as the command to run in the vng session, so maybe have something like `--init-comand CMD`?

matttbe:
> > Because we need the same options for the client and server (port/cmd) and they cannot be used at the same time, should we use `--port` and `--XXX-cmd`? Any idea for a good name for the cmd one? :)
> 
> Hm... what if we make `--server` and `--connect` mutually exclusive and then we can just have `--port` for both (we just need to check if `--server` or `--connect` has been specified).

Sorry, yes, that's what I meant to say.

> About the client command I'd do something similar, we could just have a `--command CMD`, even if it's a bit misleading as it may be interpreted as the command to run in the vng session, so maybe have something like `--init-comand CMD`?

I think `--command` is indeed misleading, too generic. I just created a new group called `Remote Console` and the new option `--remote-cmd`. I can send that in a new PR in a few minutes.

arighi:
That's great, I like `--remote-cmd`!

PR #205, opened 2024-12-09
vsock: rename options to --client/--server

matttbe:
The idea is to have other types of remote connections later on, e.g. SSH. See the discussion on #204.

So to support future types later on, the syntax is now:

    vng --server|--client [console|ssh] [--port PORT] [--remote-cmd COMMAND]

The 'ssh' option is not available yet, but by default, 'console' will be used, so users can just do:

    $ vng --server
    $ vng --client  # in another terminal

A new group for the parser has been added to avoid confusions: `--port` and `--remote-cmd` are linked to `--client` / `--server`.

While at it, the code for the console client/server has been moved to dedicated functions, grouped together. It will also be easy to support a new type later on, simply by calling a different function.

Note that this should only be a refactoring, the behaviour, apart from different option name, should not be modified.

The variables names have been renamed too. 'vsock' is now only used in `virtme(-ng)-init` and in the `socat` command.

+ a few minor changes linked to these options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,9 @@ Requirements
  * Optionally, you may need virtiofsd 1.7.0 (or higher) for better filesystem
    performance inside the virtme-ng guests.
 
+ * Optionally, you may need `socat` for the `--server` and `--client` options,
+   and the host's kernel should support VSOCK (`CONFIG_VHOST_VSOCK`).
+
 Examples
 ========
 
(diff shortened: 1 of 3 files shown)


arighi (approved):
Sorry for the late review. This is great, thank you so much for working at this @matttbe !!! I left a small comment but overall it looks great.

matttbe:
> Sorry for the late review.

Less than 48h, that's not a late review ;-)

PR #206, opened 2024-12-12
numa: introduce --numa-distance

arighi:
Introduce a new option `--numa-distance SRC,DST=VAL` to define the distance cost between two NUMA nodes and simulate non-uniform memory access.

Example usage:
```
      $ vng -r --cpu 8 -m 4G \
      >     --numa 1G,cpus=0-1 --numa 1G,cpus=2-3 \
      >     --numa 1G,cpus=4-5 --numa 1G,cpus=6-7 \
      >     --numa-distance 0,1=51 --numa-distance 0,2=31 --numa-distance 0,3=41 \
      >     --numa-distance 1,2=21 --numa-distance 1,3=61 \
      >     --numa-distance 2,3=11 -- numactl -H
      available: 4 nodes (0-3)
      node 0 cpus: 0 1
      node 0 size: 1006 MB
      node 0 free: 974 MB
      node 1 cpus: 2 3
      node 1 size: 953 MB
      node 1 free: 919 MB
      node 2 cpus: 4 5
      node 2 size: 943 MB
      node 2 free: 894 MB
      node 3 cpus: 6 7
      node 3 size: 1006 MB
      node 3 free: 965 MB
      node distances:
      node     0    1    2    3
         0:   10   51   31   41
         1:   51   10   21   61
         2:   31   21   10   11
         3:   41   61   11   10
```

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -340,6 +340,15 @@ virtme-ng is based on virtme, written by Andy Lutomirski <luto@kernel.org>.
     )
 
     parser.add_argument(
+        "--numa-distance",
+        metavar="SRC,DST=VAL",
+        action="append",
+        help="Set a distance of VAL between NUMA node SRC_NODE and DST_NODE. "
+        + "Use this option multiple times to define multiple distances between NUMA nodes. "
+        + "This option is used only together with --numa."
+    )
+
+    parser.add_argument(
         "--balloon",
         action="store_true",
         help="Allow the host to ask the guest to release memory",
@@ -1084,6 +1093,25 @@ class KernelSource:
         else:
             self.virtme_param["numa"] = ""
 
+    def _get_virtme_numa_distance(self, args):
+        if args.numa_distance is not None:
+            if not args.numa:
+                arg_fail("error: --numa-distance can be used only with --numa", show_usage=False)
+            numa_dist_str = ""
+            for arg in args.numa_distance:
+                try:
+                    nodes = arg.split('=')
+                    src, dst = nodes[0].split(',')
+                    val = nodes[1]
+                    numa_dist_str += f" --numa-distance src={src},dst={dst},val={val}"
+                except ValueError:
+                    err_msg = f"error: invalid distance '{arg}', " + \
+                               "NUMA distance string must be in the format SRC,DST=VAL"
+                    arg_fail(err_msg, show_usage=False)
+            self.virtme_param["numa_distance"] = numa_dist_str
+        else:
+            self.virtme_param["numa_distance"] = ""
+
     def _get_virtme_balloon(self, args):
         if args.balloon:
             self.virtme_param["balloon"] = "--balloon"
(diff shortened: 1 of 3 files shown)


virtme_ng/run.py
```diff
@@ -1084,6 +1093,25 @@ def _get_virtme_numa(self, args):
...
+                try:
+                    nodes = arg.split('=')
+                    src, dst = nodes[0].split(',')
+                    val = nodes[1]
+                    numa_dist_str += f" --numa-distance src={src},dst={dst},val={val}"
+                except ValueError:
+                    err_msg = f"error: invalid distance '{arg}', " + \
+                               "NUMA distance string must be in the format SRC,DST=VAL"
+                    arg_fail(err_msg, show_usage=False)
+            self.virtme_param["numa_distance"] = numa_dist_str
```
matttbe, line 1111: Out of curiosity, why not doing the parsing and moving this "complexity" to `virtme-run`?

virtme_ng/run.py
```diff
@@ -1084,6 +1093,25 @@ def _get_virtme_numa(self, args):
         else:
             self.virtme_param["numa"] = ""
 
+    def _get_virtme_numa_distance(self, args):
+        if args.numa_distance is not None:
+            if not args.numa:
+                arg_fail("error: --numa-distance can be used only with --numa", show_usage=False)
```
matttbe, line 1099: Because the issue is due to a wrong using, maybe `show_usable=False` is not needed?
But up to you, it is also easy to run `vng -h` :)

(Same below for the parsing issue).

matttbe (approved):
thank you for this new feature! I have two small questions for you, but maybe no need to change anything.

virtme_ng/run.py
arighi: > Out of curiosity, why not doing the parsing and moving this "complexity" to `virtme-run`?

At some point in the future I'd be nice to just drop virtme-run and have only vng. The virtme-run interface is maintained only for backward compatibility with the old virtme tool and vng is like a front-end to the old interface. But ultimately it'd be nice to drop the old interface and just have a front-end to qemu. That's the reason why I usually prefer to add complexity and new abstractions to vng directly.

virtme_ng/run.py
arighi: Yeah, initially I was showing the help, but I thought it was a bit too verbose and it could hide the actual error, so I added `show_usage=False`. :)

virtme_ng/run.py
matttbe: > > Out of curiosity, why not doing the parsing and moving this "complexity" to `virtme-run`?
> 
> At some point in the future I'd be nice to just drop virtme-run and have only vng. The virtme-run interface is maintained only for backward compatibility with the old virtme tool and vng is like a front-end to the old interface. But ultimately it'd be nice to drop the old interface and just have a front-end to qemu. That's the reason why I usually prefer to add complexity and new abstractions to vng directly.

Thank you, I see! And that's not what I did in my previous PR :)

Please note that I'm still using `virtme-run` directly in my automated scripts, mainly because it works, and I don't need `vng` to build the kernel (I already had that, and when I switched, `vng` could not build the kernel in a different build dir :) ).

PR #209, opened 2024-12-23
virtme-init: Start udev after mounting all filesystems

arighi:
Ensure udev is started only after all system filesystems have been mounted.

This improves the reliability of virtme-init, as the performance benefit from parallelizing filesystem mounts was negligible. For faster boot times, virtme-ng-init should be preferred, while virtme-init serves as a reliable fallback in case of issues, so prioritizing reliability over performance aligns with virtme-init's role.

This change also resolves networking failures that can occur when using virtme-init.


matttbe (approved):
Good idea, it makes sense to play it safe here with this init script.

PR #210, opened 2024-12-23
virtme: enable ssh support

arighi:
Add support for specifying `--server ssh`, to automatically start and configure sshd in the virtme-ng guest, and `--client ssh` to connect to a virtme-ng guest via SSH.

The server's port can be customized with the `--port NUM` option (default is 2222).

Example usage (start a vng instance with sshd enabled):

 $ vng --server ssh --port 2222

In another shell session on the host, connect to the running instance as a client:

 $ vng --client ssh --port 2222

Note that it's also possible to use ssh or scp directly, the --client option is provided for convenience and to be consistent with the vsock option.

Edit: additionally, introduce the following shortcuts for convenience, to quickly enable remote connection (either vsock or ssh):

 $ vng --console PORT
 $ vng --ssh PORT

Example:
```
 $ vng --ssh

 # In another terminal
 $ ssh -p 2222 localhost
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,6 +388,27 @@ Examples
    $ vng --client
 ```
 
+ - Enable ssh in the vng guest:
+```
+   # Start the vng instance with ssh server support:
+   $ vng --server ssh
+
+   # Connect to the vng guest from the host via ssh:
+   $ vng --client ssh
+```
+
+ - Generate some results inside the vng guest and copy them back to the
+   host using scp:
+```
+   # Start the vng instance with SSH server support:
+   arighi@host~> vng --server ssh
+   ...
+   arighi@virtme-ng~> ./run.sh > result.txt
+
+   # In another terminal, copy result.txt from the guest to the host using scp:
+   arighi@gpd3~> scp -P 2222 localhost:~/result.txt .
+```
+
  - Run virtme-ng inside a docker container:
 ```
    $ docker run -it --privileged ubuntu:23.10 /bin/bash
--- a/README.md
+++ b/README.md
@@ -129,8 +129,12 @@ Requirements
  * Optionally, you may need virtiofsd 1.7.0 (or higher) for better filesystem
    performance inside the virtme-ng guests.
 
- * Optionally, you may need `socat` for the `--server` and `--client` options,
-   and the host's kernel should support VSOCK (`CONFIG_VHOST_VSOCK`).
+ * Optionally, you may need `socat` for the `--console` and
+   `--console-client` options, and the host's kernel should support VSOCK
+   (`CONFIG_VHOST_VSOCK`).
+
+ * Optionally, you may need `sshd` installed for the `--ssh` and
+   `--ssh-client` options.
 
 Examples
 ========
@@ -382,26 +386,26 @@ Examples
  - Connect to a simple remote shell (`socat` is required, VSOCK will be used):
 ```
    # Start the vng instance with server support:
-   $ vng --server
+   $ vng --console
 
    # In a separate terminal run the following command to connect to a remote shell:
-   $ vng --client
+   $ vng --console-client
 ```
 
  - Enable ssh in the vng guest:
 ```
    # Start the vng instance with ssh server support:
-   $ vng --server ssh
+   $ vng --ssh
 
    # Connect to the vng guest from the host via ssh:
-   $ vng --client ssh
+   $ vng --ssh-client
 ```
 
  - Generate some results inside the vng guest and copy them back to the
    host using scp:
 ```
    # Start the vng instance with SSH server support:
-   arighi@host~> vng --server ssh
+   arighi@host~> vng --ssh
    ...
    arighi@virtme-ng~> ./run.sh > result.txt
 
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -24,6 +24,7 @@ import termios
 from shutil import which
 from time import sleep
 from base64 import b64encode
+from virtme_ng.utils import CACHE_DIR
 from .. import virtmods
 from .. import modfinder
 from .. import mkinitramfs
@@ -330,7 +331,7 @@ def make_parser() -> argparse.ArgumentParser:
     )
 
     g = parser.add_argument_group(title="Remote Console")
-    cli_srv_choices = ["console"]
+    cli_srv_choices = ["console", "ssh"]
     g.add_argument(
         "--server",
         action="store",
@@ -968,6 +969,42 @@ def console_server(args, qemu, arch, qemuargs, kernelargs):
     qemuargs.extend(["-device", "vhost-vsock-pci,guest-cid=%d" % args.port])
 
 
+def ssh_client(args):
+    if args.remote_cmd is not None:
+        exec_escaped = args.remote_cmd.replace('"', '\\"')
+        remote_cmd = ["bash", "-c", exec_escaped]
+    else:
+        remote_cmd = []
+
+    cmd = ["ssh", "-p", str(args.port), "localhost"] + remote_cmd
+
+    if args.dry_run:
+        print(" ".join(cmd))
+    else:
+        os.execvp("ssh", cmd)
+
+
+def ssh_server(args, arch, qemuargs, kernelargs):
+    # Check if we need to generate the ssh host keys for the guest.
+    ssh_key_dir = f"{CACHE_DIR}/.ssh"
+    os.makedirs(f"{ssh_key_dir}/etc/ssh", exist_ok=True)
+    os.system(f"ssh-keygen -A -f {ssh_key_dir}")
+
+    # Implicitly enable dhcp to automatically get an IP on the network
+    # interface and prevent interface renaming.
+    kernelargs.extend(["virtme.dhcp", "net.ifnames=0", "biosdevname=0"])
+
+    # Tell virtme-ng-init / virtme-init to start sshd and use the current
+    # username keys/credentials.
+    username = get_username()
+    kernelargs.extend(["virtme.ssh"])
+    kernelargs.extend([f"virtme_ssh_user={username}"])
+
+    # Setup a port forward network interface for the guest.
+    qemuargs.extend(["-device", "%s,netdev=ssh" % (arch.virtio_dev_type("net"))])
+    qemuargs.extend(["-netdev", "user,id=ssh,hostfwd=tcp::%d-:22" % args.port])
+
+
 # Allowed characters in mount paths.  We can extend this over time if needed.
 _SAFE_PATH_PATTERN = "[a-zA-Z0-9_+ /.-]+"
 _RWDIR_RE = re.compile("^(%s)(?:=(%s))?$" % (_SAFE_PATH_PATTERN, _SAFE_PATH_PATTERN))
@@ -980,7 +1017,10 @@ def do_it() -> int:
         if args.server is not None:
             arg_fail('--client cannot be used with --server.')
 
-        console_client(args)
+        if args.client == 'console':
+            console_client(args)
+        elif args.client == 'ssh':
+            ssh_client(args)
         sys.exit(0)
 
     arch = architectures.get(args.arch)
@@ -1526,7 +1566,10 @@ def do_it() -> int:
         )
 
     if args.server is not None:
-        console_server(args, qemu, arch, qemuargs, kernelargs)
+        if args.server == "console":
+            console_server(args, qemu, arch, qemuargs, kernelargs)
+        elif args.server == "ssh":
+            ssh_server(args, arch, qemuargs, kernelargs)
 
     if args.pwd:
         rel_pwd = os.path.relpath(os.getcwd(), args.root)
--- a/virtme/guest/virtme-sshd-script
+++ b/virtme/guest/virtme-sshd-script
@@ -1,0 +1,39 @@
+#!/bin/bash
+#
+# Initialize ssh server for remote connections (option `--server ssh`)
+
+if [ -z "${virtme_ssh_user}" ]; then
+    echo "ssh: virtme_ssh_user is not defined" >&2
+    exit 1
+fi
+
+SSH_HOME=$(getent passwd "${virtme_ssh_user}" | cut -d: -f6)
+if [ ! -e "${SSH_HOME}" ]; then
+    # Setup an arbitrary ssh location, just to be able to start sshd.
+    SSH_HOME=/run/ssh
+fi
+
+# Update authorized_keys by adding the user's public keys, but only if the
+# changes are confined to the guest (no modifications made to the host).
+#
+# Overwriting authorized_keys is considered safe only when the guest rootfs
+# is mounted as read-only, with an overlayfs on top to handle writes within
+# the guest environment (`--rw` not specified as argument).
+if grep ' / ' /proc/mounts | grep -q ' ro,'; then
+    cat "${SSH_HOME}"/.ssh/id_*.pub >> "${SSH_HOME}/.ssh/authorized_keys" 2>/dev/null
+    chown "${virtme_ssh_user}" "${SSH_HOME}/.ssh/authorized_keys" 2>/dev/null
+fi
+
+# Generate ssh host keys (if they don't exist already).
+CACHE_DIR=${SSH_HOME}/.cache/virtme-ng/.ssh
+mkdir -p "${CACHE_DIR}/etc/ssh"
+ssh-keygen -A -f "${CACHE_DIR}"
+ARGS=()
+for key in "${CACHE_DIR}"/etc/ssh/ssh_host_*_key; do
+    ARGS+=(-h "${key}")
+done
+
+# Start sshd.
+mkdir -p /run/sshd
+rm -f /var/run/nologin
+/usr/sbin/sshd "${ARGS[@]}"
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -487,32 +487,46 @@ virtme-ng is based on virtme, written by Andy Lutomirski <luto@kernel.org>.
     )
 
     g_remote = parser.add_argument_group(title="Remote Console")
-    cli_srv_choices = ["console", "ssh"]
 
     g_remote.add_argument(
-        "--server",
+        "--console",
         action="store",
-        const=cli_srv_choices[0],
         nargs="?",
-        choices=cli_srv_choices,
-        help="Enable a server to communicate later from the host to the device using '--client'. "
+        type=int,
+        const=2222,
+        metavar="PORT",
+        help="Enable a server to communicate later from the host using '--console-client'. "
         + "By default, a simple console will be offered using a VSOCK connection, and 'socat' for the proxy."
     )
 
     g_remote.add_argument(
-        "--client",
+        "--console-client",
         action="store",
-        const=cli_srv_choices[0],
         nargs="?",
-        choices=cli_srv_choices,
-        help="Connect to a VM launched with the '--server' option for a remote control.",
+        type=int,
+        const=2222,
+        metavar="PORT",
+        help="Connect to a VM launched with the '--console' option for a remote control.",
     )
 
     g_remote.add_argument(
-        "--port",
+        "--ssh",
         action="store",
+        nargs="?",
         type=int,
-        help="Unique port to communicate with a VM.",
+        const=2222,
+        metavar="PORT",
+        help="Enable SSH server to communicate later from the host to using '--ssh-client'."
+    )
+
+    g_remote.add_argument(
+        "--ssh-client",
+        action="store",
+        nargs="?",
+        type=int,
+        const=2222,
+        metavar="PORT",
+        help="Connect to a VM launched with the '--ssh' option for a remote control.",
     )
 
     g_remote.add_argument(
@@ -992,23 +1006,44 @@ class KernelSource:
         else:
             self.virtme_param["net_mac_address"] = ""
 
-    def _get_virtme_server(self, args):
-        if args.server is not None:
-            self.virtme_param["server"] = "--server " + args.server
+    def _get_virtme_console(self, args):
+        if args.console is not None:
+            self.virtme_param["console"] = f"--server console --port {args.console}"
         else:
-            self.virtme_param["server"] = ""
+            self.virtme_param["console"] = ""
 
-    def _get_virtme_client(self, args):
-        if args.client is not None:
-            self.virtme_param["client"] = "--client " + args.client
-        else:
-            self.virtme_param["client"] = ""
+    def _get_virtme_console_client(self, args):
+        if args.console is not None and args.console_client is not None:
+            arg_fail('--console cannot be used with --console-client', show_usage=False)
 
-    def _get_virtme_port(self, args):
-        if args.port is not None:
-            self.virtme_param["port"] = "--port " + str(args.port)
+        if args.console_client is not None:
+            self.virtme_param["console_client"] = f"--client console --port {args.console_client}"
         else:
-            self.virtme_param["port"] = ""
+            self.virtme_param["console_client"] = ""
+
+    def _get_virtme_ssh(self, args):
+        if args.console is not None and args.ssh is not None:
+            arg_fail('--console cannot be used with --ssh', show_usage=False)
+
+        if args.ssh is not None:
+            self.virtme_param["ssh"] = f"--server ssh --port {args.ssh}"
+        else:
+            self.virtme_param["ssh"] = ""
+
+    def _get_virtme_ssh_client(self, args):
+        if args.console_client is not None and args.ssh_client is not None:
+            arg_fail('--console-client cannot be used with --ssh-client', show_usage=False)
+
+        if args.ssh is not None and args.ssh_client is not None:
+            arg_fail('--ssh cannot be used with --ssh-client', show_usage=False)
+
+        if args.console is not None and args.ssh_client is not None:
+            arg_fail('--console cannot be used with --ssh-client', show_usage=False)
+
+        if args.ssh_client is not None:
+            self.virtme_param["ssh_client"] = f"--client ssh --port {args.ssh_client}"
+        else:
+            self.virtme_param["ssh_client"] = ""
 
     def _get_virtme_remote_cmd(self, args):
         if args.remote_cmd is not None:
@@ -1190,9 +1225,10 @@ class KernelSource:
         self._get_virtme_mods(args)
         self._get_virtme_network(args)
         self._get_virtme_net_mac_address(args)
-        self._get_virtme_server(args)
-        self._get_virtme_client(args)
-        self._get_virtme_port(args)
+        self._get_virtme_console(args)
+        self._get_virtme_console_client(args)
+        self._get_virtme_ssh(args)
+        self._get_virtme_ssh_client(args)
         self._get_virtme_remote_cmd(args)
         self._get_virtme_disk(args)
         self._get_virtme_sound(args)
@@ -1234,9 +1270,10 @@ class KernelSource:
             + f'{self.virtme_param["mods"]} '
             + f'{self.virtme_param["network"]} '
             + f'{self.virtme_param["net_mac_address"]} '
-            + f'{self.virtme_param["server"]} '
-            + f'{self.virtme_param["client"]} '
-            + f'{self.virtme_param["port"]} '
+            + f'{self.virtme_param["console"]} '
+            + f'{self.virtme_param["console_client"]} '
+            + f'{self.virtme_param["ssh"]} '
+            + f'{self.virtme_param["ssh_client"]} '
             + f'{self.virtme_param["remote_cmd"]} '
             + f'{self.virtme_param["disk"]} '
             + f'{self.virtme_param["sound"]} '
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -487,7 +487,7 @@ virtme-ng is based on virtme, written by Andy Lutomirski <luto@kernel.org>.
     )
 
     g_remote = parser.add_argument_group(title="Remote Console")
-    cli_srv_choices = ["console"]
+    cli_srv_choices = ["console", "ssh"]
 
     g_remote.add_argument(
         "--server",
--- a/setup.py
+++ b/setup.py
@@ -98,6 +98,7 @@ packages = [
 package_files = [
     "virtme-init",
     "virtme-udhcpc-script",
+    "virtme-sshd-script",
     "virtme-snapd-script",
     "virtme-sound-script",
 ]
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -257,6 +257,10 @@ if cat /proc/cmdline |grep -q -E '(^| )virtme.dhcp($| )'; then
     wait
 fi
 
+if cat /proc/cmdline |grep -q -E '(^| )virtme.ssh($| )'; then
+    $(dirname $0)/virtme-sshd-script
+fi
+
 if cat /proc/cmdline |grep -q -E '(^| )virtme.snapd($| )'; then
     # If snapd is present in the system try to start it, to properly support snaps.
     snapd_bin="/usr/lib/snapd/snapd";


virtme/guest/virtme-sshd-script
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+#
+# Initialize ssh server for remote connections (option `--server ssh`)
+
+HOME=$(getent passwd "${virtme_user}" | cut -d: -f6)
```
matttbe, line 5: If I'm not mistaken, `virtme_user` is not set when `root` is used.

Maybe safer to use `${virtme_user:-root}`?

virtme/guest/virtme-sshd-script
```diff
@@ -0,0 +1,10 @@
...
+cat ${HOME}/.ssh/id_rsa.pub >> ${HOME}/.ssh/authorized_keys
+chown ${virtme_user} ${HOME}/.ssh/authorized_keys
```
matttbe, line 7: same here: `${virtme_user:-root}`?

virtme/guest/virtme-sshd-script
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+#
+# Initialize ssh server for remote connections (option `--server ssh`)
+
+HOME=$(getent passwd "${virtme_user}" | cut -d: -f6)
+cat ${HOME}/.ssh/id_rsa.pub >> ${HOME}/.ssh/authorized_keys
```
matttbe, line 6: Maybe safer to cat all `id_*.pub` files? e.g. what if the default key is a `id_ed25519` (is it not the new default algo?)

virtme/guest/virtme-sshd-script
```diff
@@ -0,0 +1,10 @@
...
+chown ${virtme_user} ${HOME}/.ssh/authorized_keys
+mkdir -p /run/sshd
+rm -f /var/run/nologin
+/usr/sbin/sshd -h ~/.ssh/id_rsa
```
matttbe, line 10: Do you need `-h`? Is this script not going to be executed as root? If yes, then no need to specify `-h`, no?

If you need `-h`:
- can you use `~` instead of `${HOME}` here?
- also this RSA key might not be available if another algorithm is used

virtme/commands/run.py
```diff
@@ -968,6 +968,28 @@ def cleanup_console_script():
     qemuargs.extend(["-device", "vhost-vsock-pci,guest-cid=%d" % args.port])
 
 
+def ssh_client(args):
+    remote_cmd = args.remote_cmd if args.remote_cmd else ""
+    cmd = "ssh -p %d localhost %s" % (args.port, remote_cmd)
+    if args.dry_run:
+        print(cmd)
+    else:
+        os.system(cmd)
```
matttbe, line 977: (out of curiosity, why `os.system()`, and not `os.execvp()`? Should we change what is done in `console_client()`)

virtme/commands/run.py
```diff
@@ -968,6 +968,28 @@ def cleanup_console_script():
     qemuargs.extend(["-device", "vhost-vsock-pci,guest-cid=%d" % args.port])
 
 
+def ssh_client(args):
+    remote_cmd = args.remote_cmd if args.remote_cmd else ""
+    cmd = "ssh -p %d localhost %s" % (args.port, remote_cmd)
```
matttbe, line 973: Should you specify the user as well if `args.user` is set? (if it is not set, should it be `root`?)

virtme_ng/run.py
```diff
@@ -523,6 +523,26 @@ def make_parser():
...
+
+    g_remote.add_argument(
+        "--ssh",
+        action="store",
+        nargs="?",
+        type=int,
+        const=2222,
+        metavar="PORT",
+        help="Shortcut for `--server ssh --port PORT`"
+    )
```
matttbe, line 544: Why not dropping `--server` then?

Or have `--{console,ssh}-{client,server}` and drop `--port`?

`--client` and `--server` are maybe too generic?

matttbe (changes requested):
Thank you for looking at this! It looks good, just a few questions, mainly around the new script and options.

virtme_ng/run.py
matttbe: or `--{console,ssh}` for the server side (eventually with aliases to support `-server`), and ` --{console,ssh}-client` for the client part.

virtme/guest/virtme-sshd-script
arighi: We need to pass `-h`, otherwise sshd doesn't start and it complains about:
```
sshd: no hostkeys available -- exiting
```

I'll change the script to use `$HOME` (and also a bit smarter at finding the first private key available).

virtme/commands/run.py
arighi: `os.system()` was just easier, I was planning to change it to `os.execvp()` but I forgot, I'll fix it. :)

virtme/guest/virtme-sshd-script
arighi: The problem with root is that we won't have access to the real root's ssh keys, so that won't work.

Maybe we can always use the user that runs vng on the host (independently on `${virtme_user}`) to make sure that we will always be able to configure sshd as long as the user has some ssh keys available. But for this we need to pass a separate variable, like `${host_user}`.

virtme_ng/run.py
arighi: Hm.. probably that's a better interface, I think we've already covered all the possible "servers" with console and ssh. And from a user perspective it seems a lot more convenient to just do `vng --console ...` or `vng --ssh ...` to start the server.

virtme/guest/virtme-sshd-script
matttbe: > We need to pass `-h`, otherwise sshd doesn't start and it complains about:
> 
> ```
> sshd: no hostkeys available -- exiting
> ```

Strange, same here. I guess it is because it cannot use the keys that have been generated on the host. I can reproduce the issue on my side:

```
root@virtme-ng:/# /usr/sbin/sshd -dD
debug1: sshd version OpenSSH_9.9, OpenSSL 3.4.0 22 Oct 2024
debug1: Unable to load host key: /etc/ssh/ssh_host_rsa_key
debug1: Unable to load host key: /etc/ssh/ssh_host_ecdsa_key
debug1: Unable to load host key: /etc/ssh/ssh_host_ed25519_key
sshd: no hostkeys available -- exiting.
```

Here is a simple solution: replacing the hosts keys by new ones:

```
root@virtme-ng:/# rm /etc/ssh/ssh_host_*_key*
root@virtme-ng:/# /usr/bin/ssh-keygen -A
ssh-keygen: generating new host keys: RSA ECDSA ED25519 
root@virtme-ng:/# /usr/sbin/sshd -dD
debug1: sshd version OpenSSH_9.9, OpenSSL 3.4.0 22 Oct 2024
debug1: private host key #0: ssh-rsa SHA256:T4r1s6H3SUHnrr5ZNKRmIhoyoog8sTKbmia1W5Fw9PU
debug1: private host key #1: ecdsa-sha2-nistp256 SHA256:e0n9z3ot4KJZlSMQXV+EN0ZitMp5NhRmLpX+wx7q3po
debug1: private host key #2: ssh-ed25519 SHA256:++wwB2hFHUh8vBXKZ4S1DH17Yf0sAK6LeagcVIuHPVU
debug1: rexec_argv[1]='-dD'
debug1: Set /proc/self/oom_score_adj from 0 to -1000
debug1: Bind to port 22 on 0.0.0.0.
Server listening on 0.0.0.0 port 22.
debug1: Bind to port 22 on ::.
Server listening on :: port 22.
```

I guess instead of removing the keys, `ssh-keygen -A` should be used with `-f`, just in case `vng --rw` (or something else) is used :)

```
root@virtme-ng:/# TMP=$(mktemp -d)
root@virtme-ng:/# mkdir -p "${TMP}/etc/ssh"
root@virtme-ng:/# ssh-keygen -A -f "${TMP}"
ssh-keygen: generating new host keys: RSA ECDSA ED25519 
root@virtme-ng:/# ARGS=(); for i in "${TMP}"/etc/ssh/ssh_host_*_key; do ARGS+=(-h ${i}); done
root@virtme-ng:/# /usr/sbin/sshd "${ARGS[@]}"
```

> I'll change the script to use `$HOME` (and also a bit smarter at finding the first private key available).

Thanks!

matttbe: I forgot to add: on my side, I could not use `/usr/sbin/sshd -h ~/.ssh/id_XXX`, because my keys have a password.

virtme_ng/run.py
matttbe: Indeed, I understand the modification you did here. Then better not to confuse people with `--client` and `--server`. (then the README file needs to be adapted)

virtme/guest/virtme-sshd-script
matttbe: > The problem with root is that we won't have access to the real root's ssh keys, so that won't work.

It looks OK on my side:

```
$ sudo ls /root/.ssh/
authorized_keys
$ sudo ssh-keygen
(...)
$ sudo ls /root/.ssh/
authorized_keys  id_ed25519  id_ed25519.pub
$ sudo vng -r --user root -- ls /root/.ssh/
authorized_keys
id_ed25519
id_ed25519.pub
## and ${virtme_user} is not set
```

> Maybe we can always use the user that runs vng on the host (independently on `${virtme_user}`) to make sure that we will always be able to configure sshd as long as the user has some ssh keys available. But for this we need to pass a separate variable, like `${host_user}`.

Would it not be enough to description the limitations in the README or the `--help` menu? e.g. `sshd` is required, the host keys will be generated, only the current user (or `--user` if set) can access it, the private keys of this user will be added to the `authorized_keys` files (then not recommended with `--rw`), etc.?

virtme/guest/virtme-sshd-script
arighi: > > We need to pass `-h`, otherwise sshd doesn't start and it complains about:
> > ```
> > sshd: no hostkeys available -- exiting
> > ```
> 
> Strange, same here. I guess it is because it cannot use the keys that have been generated on the host. I can reproduce the issue on my side:

In my box it doesn't work, because as regular user I don't have access to my host's keys:
```
$ cat /etc/ssh/ssh_host_rsa_key
cat: /etc/ssh/ssh_host_rsa_key: Permission denied
```
 
> Here is a simple solution: replacing the hosts keys by new ones:

Yes, I was thinking about that, but the boot time will be slower, maybe we can try to use the user's keys if they exist and they're accessible, otherwise generate them on the fly.

matttbe: > > > We need to pass `-h`, otherwise sshd doesn't start and it complains about:
> > > ```
> > > sshd: no hostkeys available -- exiting
> > > ```
> > 
> > 
> > Strange, same here. I guess it is because it cannot use the keys that have been generated on the host. I can reproduce the issue on my side:
> 
> In my box it doesn't work, because as regular user I don't have access to my host's keys:
> 
> ```
> $ cat /etc/ssh/ssh_host_rsa_key
> cat: /etc/ssh/ssh_host_rsa_key: Permission denied
> ```

Yes, but the init script will be executed as root, no? (also needed to listen on port 22 by default)

> > Here is a simple solution: replacing the hosts keys by new ones:
> 
> Yes, I was thinking about that, but the boot time will be slower, maybe we can try to use the user's keys if they exist and they're accessible, otherwise generate them on the fly.

Loading the user's keys seems fragile: Which one to pick? What if there is a passphrase?

Generating host keys at boot time doesn't seem to be an issue, at least on my side:

``` shell
$ time sudo vng -r --user root -- 'mkdir -p /tmp/etc/ssh; ssh-keygen -A -f /tmp; /usr/sbin/sshd -h /tmp/etc/ssh/ssh_host_ed25519_key'
ssh-keygen: generating new host keys: RSA ECDSA ED25519 
sudo vng -r --user root --   0.00s user 0.01s system 0% cpu 2.384 total
$ time sudo vng -r --user root -- '/usr/sbin/sshd -h /root/.ssh/id_ed25519'
sudo vng -r --user root -- '/usr/sbin/sshd -h /root/.ssh/id_ed25519'  0.00s user 0.01s system 0% cpu 2.158 total
```

Also, to me, this SSH option is heavier than the Console one: it is normal if it takes longer, and it is more design to CI workflows, where 0.1 seconds more to be ready will not change much, no?

virtme/commands/run.py
matttbe, line 973: for the remote command, would it not be better to execute it in a subshell, e.g. with `bash -c "..."`? By doing that, it will be possible to pass longer commands, e.g. separated by semicolons.

So we could have something similar to the console mode:

```python
exec_escaped = args.remote_cmd.replace('"', '\\"')
cmd = f'bash -c "{exec_escaped}"'
```

virtme/guest/virtme-sshd-script
arighi: > > > > We need to pass `-h`, otherwise sshd doesn't start and it complains about:
> > > > ```
> > > > sshd: no hostkeys available -- exiting
> > > > ```
> > > 
> > > 
> > > Strange, same here. I guess it is because it cannot use the keys that have been generated on the host. I can reproduce the issue on my side:
> > 
> > 
> > In my box it doesn't work, because as regular user I don't have access to my host's keys:
> > ```
> > $ cat /etc/ssh/ssh_host_rsa_key
> > cat: /etc/ssh/ssh_host_rsa_key: Permission denied
> > ```
> 
> Yes, but the init script will be executed as root, no? (also needed to listen on port 22 by default)

It is root in the guest, but on the host virtiofsd runs as regular user, so in practice we can't access the file, even if we're root in the guest. And that's good, otherwise we'd provide a way to bypass file permissions, which won't be really nice. :)

> 
> > > Here is a simple solution: replacing the hosts keys by new ones:
> > 
> > 
> > Yes, I was thinking about that, but the boot time will be slower, maybe we can try to use the user's keys if they exist and they're accessible, otherwise generate them on the fly.
> 
> Loading the user's keys seems fragile: Which one to pick? What if there is a passphrase?
> 
> Generating host keys at boot time doesn't seem to be an issue, at least on my side:

Yeah, the passphrase is a good point and generating them on-the-fly is probably more reliable. However, the annoying part about changing the keys every time is that ssh will complain about seeing different keys in `~/.known_hosts`.

Maybe we could pre-generate the guest keys and store them in `~/.cache/virtme-ng/.ssh/`? Then we pass the keys to the guest and we'll have safe ssh keys to use that we can access and we won't have known_hosts issues.

> Also, to me, this SSH option is heavier than the Console one: it is normal if it takes longer, and it is more design to CI workflows, where 0.1 seconds more to be ready will not change much, no?

Yes, ssh is certainly heavier than console, however some users have  workflows that involve transferring data to/from the VM via scp, so having an easy way to enable ssh access in the guest would be highly convenient for them and they could easily migrate their CI to use vng.

matttbe: > > > > > We need to pass `-h`, otherwise sshd doesn't start and it complains about:
> > > > > ```
> > > > > sshd: no hostkeys available -- exiting
> > > > > ```
> > > > 
> > > > 
> > > > Strange, same here. I guess it is because it cannot use the keys that have been generated on the host. I can reproduce the issue on my side:
> > > 
> > > 
> > > In my box it doesn't work, because as regular user I don't have access to my host's keys:
> > > ```
> > > $ cat /etc/ssh/ssh_host_rsa_key
> > > cat: /etc/ssh/ssh_host_rsa_key: Permission denied
> > > ```
> > 
> > 
> > Yes, but the init script will be executed as root, no? (also needed to listen on port 22 by default)
> 
> It is root in the guest, but on the host virtiofsd runs as regular user, so in practice we can't access the file, even if we're root in the guest. And that's good, otherwise we'd provide a way to bypass file permissions, which won't be really nice. :)

OK I see, in my case I launch vng from a docker container, with the root user. So both the host and the guest are root :)

Anyway, from the VM, best not to use the same host keys as the ones of the host, even if we are root and have access to them.

> > > > Here is a simple solution: replacing the hosts keys by new ones:
> > > 
> > > 
> > > Yes, I was thinking about that, but the boot time will be slower, maybe we can try to use the user's keys if they exist and they're accessible, otherwise generate them on the fly.
> > 
> > 
> > Loading the user's keys seems fragile: Which one to pick? What if there is a passphrase?
> > Generating host keys at boot time doesn't seem to be an issue, at least on my side:
> 
> Yeah, the passphrase is a good point and generating them on-the-fly is probably more reliable. However, the annoying part about changing the keys every time is that ssh will complain about seeing different keys in `~/.known_hosts`.
> 
> Maybe we could pre-generate the guest keys and store them in `~/.cache/virtme-ng/.ssh/`? Then we pass the keys to the guest and we'll have safe ssh keys to use that we can access and we won't have known_hosts issues.

Good idea!

> > Also, to me, this SSH option is heavier than the Console one: it is normal if it takes longer, and it is more design to CI workflows, where 0.1 seconds more to be ready will not change much, no?
> 
> Yes, ssh is certainly heavier than console, however some users have workflows that involve transferring data to/from the VM via scp, so having an easy way to enable ssh access in the guest would be highly convenient for them and they could easily migrate their CI to use vng.

I understand. So more an option to be compatible with older workflows, but not recommended for new ones. (Except if there is really a need for a remote host to communicate with the VM).

virtme/commands/run.py
```diff
@@ -968,6 +969,46 @@ def cleanup_console_script():
...
+        os.mkdir(ssh_key_dir)
+        os.system(f"ssh-keygen -t rsa -b 2048 -f {ssh_key_dir}/id_rsa -N '' -q")
+
+    # Implicitly enable dhcp to automatically get an IP on the network
+    # interface and prevent interface renaming.
+    kernelargs.extend(["virtme.dhcp", "net.ifnames=0", "biosdevname=0"])
+
+    # Tell virtme-ng-init / virtme-init to start sshd and use the current
+    # username keys/credentials.
+    username = get_username()
```
matttbe, line 1003: What if `--user` is passed? Should it eventually be:

    username = args.user if args.user else get_username()

Or maybe you do want the current user, because if `--user root` is used, but `vng` is not launched as root, that will be an issue?

virtme/guest/virtme-sshd-script
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+#
+# Initialize ssh server for remote connections (option `--server ssh`)
+
+# Generate authorized_keys adding the host's user public keys.
+HOME=$(getent passwd "${ssh_user}" | cut -d: -f6)
```
matttbe, line 6: (see above: maybe `virtme_ssh_user`? Same below with `chown`)

virtme/guest/virtme-sshd-script
```diff
@@ -0,0 +1,20 @@
...
+cat ${HOME}/.ssh/id_*.pub >> ${HOME}/.ssh/authorized_keys
```
matttbe, line 7: (do you think we should add double quotes around `HOME` just to be on the safe side? same below)

virtme/guest/virtme-sshd-script
```diff
@@ -0,0 +1,20 @@
...
+chown ${ssh_user} ${HOME}/.ssh/authorized_keys
+
+# Try to use the virtme-ng host keys if available, otherwise generate a new
+# key on the fly.
+HOST_KEY=${HOME}/.cache/virtme-ng/.ssh/id_rsa
+if [ ! -e "${HOST_KEY}" ]; then
+    ssh-keygen -t rsa -b 2048 -f "${HOST_KEY}" -N "" -q
```
matttbe, line 14: `mkdir -p` will be needed here.

But as mentioned above, I think it would be safer to do this here instead:

```
CACHE_DIR="{HOME}/.cache/virtme-ng/ssh"  ## or /virtme/?
mkdir -p "${CACHE_DIR}/etc/ssh"
ssh-keygen -A -f "${CACHE_DIR}" ## will not do anything if the keys exist
ARGS=(); for i in "${CACHE_DIR}"/etc/ssh/ssh_host_*_key; do ARGS+=(-h ${i}); done

(...)

/usr/sbin/sshd "${ARGS[@]}"

File: virtme/commands/run.py [hunk 12]
```diff
@@ -968,6 +969,46 @@ def cleanup_console_script():
...
+    )
+    if args.dry_run:
+        print(" ".join(cmd))
+    else:
+        os.execvp("ssh", cmd)
+
+
+def ssh_server(args, arch, qemuargs, kernelargs):
+    # Check if we need to generate the ssh host keys for the guest.
+    ssh_key_dir = f"{CACHE_DIR}/.ssh"
```
matttbe, line 992: (maybe no need to add the `.` in `.ssh`?)

matttbe, line 992: you might want to prefix it with `args.root` if defined, no?

matttbe (changes requested):
Thank you for the new version!

It looks good to me, apart a few missing `mkdir -p`. I also added a few questions and suggestions, but they might not require changes.

README.md
```diff
@@ -382,26 +383,26 @@ Examples
...
    # In a separate terminal run the following command to connect to a remote shell:
-   $ vng --client
+   $ vng --console-client
 ```
 
  - Enable ssh in the vng guest:
 ```
    # Start the vng instance with ssh server support:
-   $ vng --server ssh
+   $ vng --ssh
```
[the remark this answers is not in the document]
arighi: I'm not sure about introducing the `--scp` option, it doesn't seem to provide much compared to just running scp, it can help just to figure out the port, but I don't think it's worth to maintain an option just for that. I think we can simply show an example in the README on how to use scp and be happy with it. :)

virtme_ng/run.py
```diff
@@ -992,23 +1006,38 @@ def _get_virtme_net_mac_address(self, args):
         else:
             self.virtme_param["net_mac_address"] = ""
 
-    def _get_virtme_server(self, args):
-        if args.server is not None:
-            self.virtme_param["server"] = "--server " + args.server
+    def _get_virtme_console(self, args):
+        if args.console is not None and args.ssh is not None:
+            arg_fail('--console cannot be used with --ssh', show_usage=False)
```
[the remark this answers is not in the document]
arighi: Ah no, that's a good point! We definitely want to arg_fail() also when `--ssh-client` is used together with `--console-client`, the vng UI needs to be consistent. I'll add a check also for this.

virtme_ng/run.py
```diff
@@ -992,23 +1006,38 @@ def _get_virtme_net_mac_address(self, args):
...
+            self.virtme_param["console_client"] = f"--client console --port {args.console_client}"
+        else:
+            self.virtme_param["console_client"] = ""
+
+    def _get_virtme_ssh(self, args):
+        if args.console is not None and args.ssh is not None:
+            arg_fail('--console cannot be used with --ssh', show_usage=False)
+
+        if args.ssh is not None and args.root is not None:
+            arg_fail('--ssh is not currently compatible with --root', show_usage=False)
```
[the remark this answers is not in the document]
arighi: So, I added this check because we don't know if the user that runs vng has write access to the rootfs and in that case it'd fail to generate the ssh keys in the cache dir.

We could export/mount a separate fs using 9p for the cache dir, like we do for the guest tools, but I didn't want to add too much stuff with this PR, so I was thinking to keep it simple and do not allow `--ssh` with `--root` for now.

Later (as a separate PR) we should probably introduce a generic shared cachedir between host and guest, where we can put data generated on-the-fly, like the ssh keys, or temporary scripts generated on the host that must be executed inside the guest (see `console_script_path`), etc.

WDYT?

virtme_ng/run.py
```diff
@@ -992,23 +1006,38 @@ def _get_virtme_net_mac_address(self, args):
...
```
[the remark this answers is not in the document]
arighi: Oh we definitely want to fail if `--rw` is provided. I really hate `--rw`, I wish we could just drop that at some point. :D

virtme/commands/run.py
```diff
@@ -24,6 +24,7 @@
 from shutil import which
 from time import sleep
 from base64 import b64encode
+from virtme_ng.utils import CACHE_DIR
```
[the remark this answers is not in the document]
arighi: Yes, we should probably move all this logic in vng and pass cache dir as a parameter to virtme-run.

virtme/commands/run.py
```diff
@@ -968,6 +969,46 @@ def cleanup_console_script():
...
+    else:
+        os.execvp("ssh", cmd)
+
+
+def ssh_server(args, arch, qemuargs, kernelargs):
+    # Check if we need to generate the ssh host keys for the guest.
+    ssh_key_dir = f"{CACHE_DIR}/.ssh"
+    if not os.path.exists(ssh_key_dir):
+        os.mkdir(ssh_key_dir)
+        os.system(f"ssh-keygen -t rsa -b 2048 -f {ssh_key_dir}/id_rsa -N '' -q")
```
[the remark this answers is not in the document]
arighi: Hm... I'd like to keep vng self-consistent and not add any dependency on libvirt or similar tools, so I'd prefer to simply auto-generate the host ssh keys somewhere (maybe provide an option that allows the user to pass the keys as a future improvement).

Again, the args.root incompatibility will be solved once we'll have a shared cachedir between host and guest. And the virtme-ng dependency can also be solved by moving the logic to virtme-ng and passing the cachedir as a separate option to virtme-run.

virtme/commands/run.py
arighi: I think ssh should be allowed only for the user that runs vng, independently on the `--user` option. Otherwise it's a bit too tricky to figure out file permissions, etc. and we may have more problems than benefits.... In fact I think I may also ignore `--user` in `--ssh-client` and just allow the current user to log into the vng instance for now. It seems safer and simpler and we can improve the logic later if needed. :)

virtme/guest/virtme-sshd-script
arighi: About the cache dir, we already have `~/.cache/virtme-ng` (used to cache the precompiled mainline kernels), so we should use the same location.

virtme/commands/run.py
arighi: If the cache dir was only used by ssh, yes, but considering that it's a generic cache dir (use to cache the mainline precompiled kernels) I think it's good to have a separate dot dir for ssh.

arighi: In the future definitely yes, but for now `--root` can't be used with `--ssh` (for the potential permissions limitations). This should also be solved once we introduce the generic shared cache dir between host and guest.

README.md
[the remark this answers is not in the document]
matttbe: It makes sense. I suppose people interested in using `scp` will look at the `README` file, not just the man/`--help`.

In this case, no need to switch to `--ssh[d]` and `--console[d]` I suppose.

virtme/commands/run.py
matttbe: OK, I thought there was no risk of conflicts with the kernel versions, so it was fine to call it `ssh` and keep it "visible". But fine for me to keep it hidden.

virtme/commands/run.py
matttbe: Fine for me to keep it simple for the moment. Should we have this "limitation" written somewhere?

virtme/commands/run.py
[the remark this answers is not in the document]
matttbe: > > I wonder if you can force the SSH host keys in libvirt and others.
>
> Hm... I'd like to keep vng self-consistent and not add any dependency on libvirt or similar tools, so I'd prefer to simply auto-generate the host ssh keys somewhere (maybe provide an option that allows the user to pass the keys as a future improvement).

Sorry, I didn't express it correctly, let me rephrase: here, we keep the host keys so each time a VM is started, it re-uses the same keys. Do you know how people are generally dealing with SSH hosts keys for test VMs? Maybe it is normal to have different host keys, and the verification is bypassed? Maybe we don't even need to keep the same host keys? Or maybe it is common to set up a new test VM with existing host keys, so the verification is not bypassed? (I don't know what can easily be done with libvirt, vmware, etc.)

> Again, the args.root incompatibility will be solved once we'll have a shared cachedir between host and guest. And the virtme-ng dependency can also be solved by moving the logic to virtme-ng and passing the cachedir as a separate option to virtme-run.

:+1:

virtme/guest/virtme-sshd-script
matttbe: > About the cache dir, we already have `~/.cache/virtme-ng` (used to cache the precompiled mainline kernels), so we should use the same location.

Yes, I was not sure if we wanted a different cache for `virtme-run` and `virtme-ng`, to avoid any conflicts.

But fine to keep the same location.

matttbe (changes requested):
Sorry, a few more comments :)

But only one thing to change I think to properly fill `authorized_keys` file.

virtme/guest/virtme-sshd-script
```diff
@@ -0,0 +1,24 @@
...
+# Initialize ssh server for remote connections (option `--server ssh`)
+
+if [ ! -n "${virtme_ssh_user}" ]; then
+    echo "ssh: virtme_ssh_user is not defined" >&2
+    exit 1
+fi
+
+# Generate authorized_keys adding the host's user public keys.
+HOME=$(getent passwd "${virtme_ssh_user}" | cut -d: -f6)
+cat "${HOME}/.ssh/id_*.pub" >> "${HOME}/.ssh/authorized_keys"
```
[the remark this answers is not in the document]
arighi: ah good catch! will fix that, thanks for noticing!

virtme/commands/run.py
```diff
@@ -968,6 +969,43 @@ def cleanup_console_script():
...
+    if args.dry_run:
+        print(" ".join(cmd))
+    else:
+        os.execvp("ssh", cmd)
+
+
+def ssh_server(args, arch, qemuargs, kernelargs):
+    # Check if we need to generate the ssh host keys for the guest.
+    ssh_key_dir = f"{CACHE_DIR}/.ssh"
+    if not os.path.exists(ssh_key_dir):
```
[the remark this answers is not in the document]
arighi: About `args.root` you're right, the host keys should be generated at runtime, so we should be able to start sshd in any case.

We won't be able to add the user's keys to `authorized_keys` (unless the ssh keys are also available in the rootfs), and maybe that's fine, as the user can still be able to login using the password. This can probably be detailed later in the README (in the troubleshooting section).

About `--rw` maybe we should simply skip the step that tries to update `authorized_keys`, this should be safe and, again, users can either configure the `authorized_keys` manually or they can still be able to ssh using the password.

I'll do some tests with `--rw` and `--root`. Thanks!

arighi:
About the host keys, thanks for clarifying. I think most VM managers automatically generate new keys, but they also give the option to pass some pre-generated keys (using cloud-init for example) with special config options. We can probably follow the same approach, auto-generate them for now and in the future add an option to pass some pre-generate host keys if it's really needed.

virtme/guest/virtme-sshd-script
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+# Generate authorized_keys adding the host's user public keys, but only if
+# the rootfs is mounted in read-only.
+SSH_HOME=$(getent passwd "${virtme_ssh_user}" | cut -d: -f6)
+if [ ! -e "${SSH_HOME}" ]; then
+    # Setup an arbitrary ssh location, just to be able to start sshd.
+    SSH_HOME=/run/ssh
+fi
+if grep ' / ' /proc/mounts | grep -q ' ro,'; then
```
matttbe, line 17: It's a detail, but it feels strange to modify files only if the rootfs is in read-only :-)

Would it not make more sense to grep on `/home overlay rw,`?

virtme/guest/virtme-sshd-script
```diff
@@ -0,0 +1,31 @@
...
+    cat ${SSH_HOME}/.ssh/id_*.pub >> "${SSH_HOME}/.ssh/authorized_keys" 2>/dev/null
```
matttbe, line 18: I guess `SSH_HOME` will not contain spaces. Just in case this patch fixes warnings reported by shellcheck:

```diff
diff --git a/virtme/guest/virtme-sshd-script b/virtme/guest/virtme-sshd-script
index 334ae24..5ef2915 100755
--- a/virtme/guest/virtme-sshd-script
+++ b/virtme/guest/virtme-sshd-script
@@ -2,7 +2,7 @@
 #
 # Initialize ssh server for remote connections (option `--server ssh`)
 
-if [ ! -n "${virtme_ssh_user}" ]; then
+if [ -z "${virtme_ssh_user}" ]; then
     echo "ssh: virtme_ssh_user is not defined" >&2
     exit 1
 fi
@@ -15,7 +15,7 @@ if [ ! -e "${SSH_HOME}" ]; then
     SSH_HOME=/run/ssh
 fi
 if grep ' / ' /proc/mounts | grep -q ' ro,'; then
-    cat ${SSH_HOME}/.ssh/id_*.pub >> "${SSH_HOME}/.ssh/authorized_keys" 2>/dev/null
+    cat "${SSH_HOME}"/.ssh/id_*.pub >> "${SSH_HOME}/.ssh/authorized_keys" 2>/dev/null
     chown "${virtme_ssh_user}" "${SSH_HOME}/.ssh/authorized_keys" 2>/dev/null
 fi
 
@@ -23,7 +23,10 @@ fi
 CACHE_DIR=${SSH_HOME}/.cache/virtme-ng/.ssh
 mkdir -p "${CACHE_DIR}/etc/ssh"
 ssh-keygen -A -f "${CACHE_DIR}"
-ARGS=(); for i in "${CACHE_DIR}"/etc/ssh/ssh_host_*_key; do ARGS+=(-h ${i}); done
+ARGS=()
+for key in "${CACHE_DIR}"/etc/ssh/ssh_host_*_key; do
+    ARGS+=(-h "${key}")
+done
 
 # Start sshd.
 mkdir -p /run/sshd
```

(while at it, I split the for-loop on different lines ; shellcheck was only asking to add double quotes around `${i}`)

matttbe (approved):
Thank you for the last update!

This last version looks good to me, and I tested it.

I have just two minors comments on the sshd script, but you can ignore them and merge directly if you prefer.

virtme/guest/virtme-sshd-script
arighi: Haha yes, it's quite obscure if you're not familiar with the filesystem layout. :smile:  I'll change the logic to check for the rw overlay.

arighi: Actually, I take it back... trying to figure out if the cache dir is in an overlay isn't trivial, because potentially it can be outside of /home. Basically what I care about is preventing any changes to the host fs, and that happens only when `--rw` is passed, and the most effective way from the guest to check if `--rw` was specified is to check if the rootfs has been mounted in `rw` or `ro`.

Maybe I can keep this check and add more details to explain why we're doing this check.

arighi:
@matttbe I've applied your patch to `virtme-sshd-script` and added more details to the comment about the check for the read-only rootfs.

I think this PR is decent enough for a merge at this point and we can improve things in-tree later. Moreover, after this PR is merged I was considering to cut a new release, because we have some really cool features. :)

Thank you so much for all the help on this!

virtme/guest/virtme-sshd-script
matttbe: Thank you for the update. I think the current version is good, no need to do other modifications.

Just in case, if all you want to check is if overlayfs is used, you can use:

```shell
OVERLAYFS="794c7630"
if [ "$(stat -f -c "%t" "${SSH_HOME}")" = "${OVERLAYFS}" ]; then
```

(you could use `%T` which should print `overlayfs`, but it doesn't work with busybox apparently -- you can use [`OVERLAYFS_SUPER_MAGIC`](https://elixir.bootlin.com/linux/v6.12.6/source/include/uapi/linux/magic.h#L38) instead)

matttbe:
> @matttbe I've applied your patch to `virtme-sshd-script` and added more details to the comment about the check for the read-only rootfs.

Thank you!

> I think this PR is decent enough for a merge at this point and we can improve things in-tree later.

Sounds good to me!

> Moreover, after this PR is merged I was considering to cut a new release, because we have some really cool features. :)

Good idea!

Once released, I will use it in my script to use the console instead of the serial!

> Thank you so much for all the help on this!

You are welcome! Thank you for the modifications and the different updates!

virtme/guest/virtme-sshd-script
arighi: Looks like a good change for a separate PR. :)

matttbe: done :-)

https://github.com/arighi/virtme-ng/pull/211

marcosps:
Thanks a lot for working on this feature! I just tried to test it, and the `vng -r --ssh` works, but when I try running a new instance of vng as client with `vng --ssh-client`, it returns this error message:

```
kex_exchange_identification: read: Connection reset by peer
Connection reset by 127.0.0.1 port 2222
```

The same happens when trying to connect with plain ssh:

```
$ ssh -p 2222 localhost
kex_exchange_identification: read: Connection reset by peer
Connection reset by 127.0.0.1 port 2222
```

Is it something wrong with my machine? I mean, I never really playing with portfwd or anything like that, so I'm not sure what can be wrong in this case: if it's related to vng or something on my machine's ssh.

Also, I installed virtme-ng using pipx:

```
$ pipx install --force .
```

Inside the virtme-ng clonned dir.

arighi:
> Thanks a lot for working on this feature! I just tried to test it, and the `vng -r --ssh` works, but when I try running a new instance of vng as client with `vng --ssh-client`, it returns this error message:
> 
> ```
> kex_exchange_identification: read: Connection reset by peer
> Connection reset by 127.0.0.1 port 2222
> ```

Hi @marcosps, to make sure I understand, when you say it works, you mean that you're able to ssh into the first instance, but other (new) instances are not working? Or even the first one is not working?

marcosps:
> > Thanks a lot for working on this feature! I just tried to test it, and the `vng -r --ssh` works, but when I try running a new instance of vng as client with `vng --ssh-client`, it returns this error message:
> > ```
> > kex_exchange_identification: read: Connection reset by peer
> > Connection reset by 127.0.0.1 port 2222
> > ```
> 
> Hi @marcosps, to make sure I understand, when you say it works, you mean that you're able to ssh into the first instance, but other (new) instances are not working? Or even the first one is not working?

@arighi it "works"  because I see the VM started (the server), and QEMU is running.

IIUC, by using --ssh on the "server" I would still see a bash session on the "server" as the same as before, and only the --ssh-client on a different session would allow me to "join" in the same VM, right?


> @marcosps thank you for testing it!
> 
> Can you `ssh localhost` from inside the vm?

@matttbe 

Nope: `ssh: connect to host localhost port 22: Connection refused`

matttbe:
> > Can you `ssh localhost` from inside the vm?
> 
> @matttbe: Nope: `ssh: connect to host localhost port 22: Connection refused`

This feature has a few assumptions: the user has a "normal" key at the traditional place (`~/.ssh/id_*.pub`), overlayfs is used in the home dir, only the current user executing `vng` can log in, sshd is installed on the host (or chroot), etc.

Maybe best to check inside the VM what's there: sshd running? `.authorized_keys` OK? etc.? Maybe run `ssh` with `-vvv`.

arighi:
@marcosps you can also try to run `virtme-sshd-script` (from the guest tools dir) manually inside the guest to see if you get an explicit error, it looks like sshd failed to start in your case.

marcosps:
> > > Can you `ssh localhost` from inside the vm?
> > 
> > 
> > @matttbe: Nope: `ssh: connect to host localhost port 22: Connection refused`
> 
> This feature has a few assumptions: the user has a "normal" key at the traditional place (`~/.ssh/id_*.pub`), overlayfs is used in the home dir, only the current user executing `vng` can log in, sshd is installed on the host (or chroot), etc.

Yep, sshd is installed, I'm trying to ssh into the VM with the same user, by "overlayfs in the home dir", you mean, having overlayfs support on the VM? if yes, than I'm covered. I also have the "normal" keys in my ~/.ssh/

> 
> Maybe best to check inside the VM what's there: sshd running? `.authorized_keys` OK? etc.? Maybe run `ssh` with `-vvv`.


> @marcosps you can also try to run `virtme-sshd-script` (from the guest tools dir) manually inside the guest to see if you get an explicit error, it looks like sshd failed to start in your case.

Thanks @matttbe and @arighi it seems that my sshd is failing to start:

```
mpdesouza@virtme-ng:~/git/linux> /home/mpdesouza/git/virtme-ng/virtme/guest/virtme-sshd-script
rm: cannot remove '/var/run/nologin': Operation not permitted
/usr/etc/ssh/sshd_config: Permission denied
mpdesouza@virtme-ng:~/git/linux> ls -l /usr/etc/ssh/sshd_config
-rw-r----- 1 root root 3727 Oct 28 08:18 /usr/etc/ssh/sshd_config
```

I'm using openSUSE Tumbleweed, which already uses the usrmerge (either way, I'm running as a normal user, which doesn't allow me to start sshd either way)....

arighi:
...
 
> Thanks @matttbe and @arighi it seems that my sshd is failing to start:
> 
> ```
> mpdesouza@virtme-ng:~/git/linux> /home/mpdesouza/git/virtme-ng/virtme/guest/virtme-sshd-script
> rm: cannot remove '/var/run/nologin': Operation not permitted
> /usr/etc/ssh/sshd_config: Permission denied
> mpdesouza@virtme-ng:~/git/linux> ls -l /usr/etc/ssh/sshd_config
> -rw-r----- 1 root root 3727 Oct 28 08:18 /usr/etc/ssh/sshd_config
> ```
> 
> I'm using openSUSE Tumbleweed, which already uses the usrmerge (either way, I'm running as a normal user, which doesn't allow me to start sshd either way)....

You need to run the sshd script as root (in the guest), and you probably need to `export virtme_ssh_user=YOUR_USER`, sorry I forgot to mention that.

marcosps:
> ...
> 
> > Thanks @matttbe and @arighi it seems that my sshd is failing to start:
> > ```
> > mpdesouza@virtme-ng:~/git/linux> /home/mpdesouza/git/virtme-ng/virtme/guest/virtme-sshd-script
> > rm: cannot remove '/var/run/nologin': Operation not permitted
> > /usr/etc/ssh/sshd_config: Permission denied
> > mpdesouza@virtme-ng:~/git/linux> ls -l /usr/etc/ssh/sshd_config
> > -rw-r----- 1 root root 3727 Oct 28 08:18 /usr/etc/ssh/sshd_config
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm using openSUSE Tumbleweed, which already uses the usrmerge (either way, I'm running as a normal user, which doesn't allow me to start sshd either way)....
> 
> You need to run the sshd script as root (in the guest), and you probably need to `export virtme_ssh_user=YOUR_USER`, sorry I forgot to mention that.

Ah, ok ok, makes sense. Sorry about it. I'm now running as root, but I'm still seeing the EACCESS error:

```
# /home/mpdesouza/git/virtme-ng/virtme/guest/virtme-sshd-script 
/usr/etc/ssh/sshd_config: Permission denied

# id
'uid=0(root) gid=0(root) groups=0(root)

# /home/mpdesouza/git/virtme-ng/virtme/guest/virtme-sshd-script 
/usr/etc/ssh/sshd_config: Permission denied

# echo $virtme_ssh_user 
mpdesouza
```

I tried to `strace` it, and it seems that's still triggering the EACCES, even tough I'm root:

```
# strace --follow-forks /home/mpdesouza/git/virtme-ng/virtme/guest/virtme-sshd-scr
ipt

.... <lots and lots of output>
[pid   415] newfstatat(AT_FDCWD, "/usr/etc/ssh/sshd_config", {st_mode=S_IFREG|0640, st_size=3727, ...}, 0) = 0
[pid   415] newfstatat(AT_FDCWD, "/etc/ssh/sshd_config", 0x7fffe0f8ac00, 0) = -1 ENOENT (No such file or directory)
[pid   415] openat(AT_FDCWD, "/usr/etc/ssh/sshd_config", O_RDONLY) = -1 EACCES (Permission denied)
[pid   415] dup(2)                      = 7
[pid   415] fcntl(7, F_GETFL)           = 0x2 (flags O_RDWR)
[pid   415] fstat(7, {st_mode=S_IFCHR|0600, st_rdev=makedev(0x88, 0), ...}) = 0
[pid   415] write(7, "/usr/etc/ssh/sshd_config: Permis"..., 44/usr/etc/ssh/sshd_config: Permission denied
) = 44
[pid   415] close(7)                    = 0
[pid   415] exit_group(1)               = ?
[pid   415] +++ exited with 1 +++
```

So I'm not sure what's going on... starting sshd on my host system as root works, so I'm not sure why this doesn't work...

arighi:
> I tried to `strace` it, and it seems that's still triggering the EACCES, even tough I'm root:
> 
> ```
> # strace --follow-forks /home/mpdesouza/git/virtme-ng/virtme/guest/virtme-sshd-scr
> ipt
> 
> .... <lots and lots of output>
> [pid   415] newfstatat(AT_FDCWD, "/usr/etc/ssh/sshd_config", {st_mode=S_IFREG|0640, st_size=3727, ...}, 0) = 0
> [pid   415] newfstatat(AT_FDCWD, "/etc/ssh/sshd_config", 0x7fffe0f8ac00, 0) = -1 ENOENT (No such file or directory)
> [pid   415] openat(AT_FDCWD, "/usr/etc/ssh/sshd_config", O_RDONLY) = -1 EACCES (Permission denied)
> [pid   415] dup(2)                      = 7
> [pid   415] fcntl(7, F_GETFL)           = 0x2 (flags O_RDWR)
> [pid   415] fstat(7, {st_mode=S_IFCHR|0600, st_rdev=makedev(0x88, 0), ...}) = 0
> [pid   415] write(7, "/usr/etc/ssh/sshd_config: Permis"..., 44/usr/etc/ssh/sshd_config: Permission denied
> ) = 44
> [pid   415] close(7)                    = 0
> [pid   415] exit_group(1)               = ?
> [pid   415] +++ exited with 1 +++
> ```
> 
> So I'm not sure what's going on... starting sshd on my host system as root works, so I'm not sure why this doesn't work...

Ah ok, makes sense, that's because sshd is trying to access `/usr/etc/ssh/sshd_config` on the host via virtiofsd, but virtiofsd (that runs as your regular user) probably doesn't have access to that file, hence EACCESS.

In my case I have the config in `/etc/ssh/sshd_config`, that is world readable. Maybe we can fix this by generating a custom sshd_config in tmp and bind mount to the real sshd_config if we don't have read access.

matttbe:
> > > I'm using openSUSE Tumbleweed, which already uses the usrmerge (either way, I'm running as a normal user, which doesn't allow me to start sshd either way)....
> > 
> > You need to run the sshd script as root (in the guest), and you probably need to `export virtme_ssh_user=YOUR_USER`, sorry I forgot to mention that.
> 
> Ah, ok ok, makes sense. Sorry about it. I'm now running as root, but I'm still seeing the EACCESS error:

From the host, can you access `/usr/etc/ssh/sshd_config`?
If not, and if `vng` is started by your user, I guess it is normal not to have access to this file from the guest.

matttbe:
> Maybe we can fix this by generating a custom sshd_config in tmp and bind mount to the real sshd_config if we don't have read access.

Or could it be enough to write something in the troubleshooting section in the README file for the moment? 
Or maybe it is easy to maintain a small config file? Can we easily find where this file should be located?

arighi:
> > Maybe we can fix this by generating a custom sshd_config in tmp and bind mount to the real sshd_config if we don't have read access.
> 
> Or could it be enough to write something in the troubleshooting section in the README file for the moment? Or maybe it is easy to maintain a small config file? Can we easily find where this file should be located?

Yeah, finding the proper location for the confiig file can be tricky, I'd hope that the first path that sshd checks is `/etc/ssh//sshd_config`, if that's the case we may just place a custom config there and be happy...

matttbe:
> > > Maybe we can fix this by generating a custom sshd_config in tmp and bind mount to the real sshd_config if we don't have read access.
> > 
> > 
> > Or could it be enough to write something in the troubleshooting section in the README file for the moment? Or maybe it is easy to maintain a small config file? Can we easily find where this file should be located?
> 
> Yeah, finding the proper location for the confiig file can be tricky, I'd hope that the first path that sshd checks is `/etc/ssh//sshd_config`

I guess it is hardcoded at build time, using info passed to the `configure` script.

> if that's the case we may just place a custom config there and be happy...

Maybe it is easier to maintain and always use a minimal custom config. At least we would be sure the port is always 22 for example.

arighi:
I think all distro are configuring sshd to use the config from m /etc/ssh/sshd_config, if we maintain a small file there we can probably solve 90% of the issues (or worst case we can check a couple of known paths depending on the distro)

arighi:
@marcosps @matttbe #212 should fix the EACESS issue.

Moreover, I was thinking, why don't we just allow empty password for all users (see https://github.com/arighi/virtme-ng-init/pull/16) and set `PermitEmptyPasswords yes` in the custom sshd_config? In this way we can completely get rid of the `authorized_keys` logic, simplify the sshd setup and make everything more robust.

At the end we don't care about security inside the guest, because it's for testing purposes only, and from a security standpoint it's also safe, because all the filesystem operations are still performed by virtiofsd, that runs as the host user, independently on the user in the guest.

virtme/commands/run.py
```diff
@@ -968,6 +969,42 @@ def cleanup_console_script():
...
+
+    # Tell virtme-ng-init / virtme-init to start sshd and use the current
+    # username keys/credentials.
+    username = get_username()
+    kernelargs.extend(["virtme.ssh"])
+    kernelargs.extend([f"virtme_ssh_user={username}"])
+
+    # Setup a port forward network interface for the guest.
+    qemuargs.extend(["-device", "%s,netdev=ssh" % (arch.virtio_dev_type("net"))])
+    qemuargs.extend(["-netdev", "user,id=ssh,hostfwd=tcp::%d-:22" % args.port])
```
matttbe, line 1005: As mentioned on https://github.com/arighi/virtme-ng-init/pull/16#pullrequestreview-2526363654, this should be modified:

    "hostfwd=tcp:127.0.0.1:%d-:22" % args.port

marcosps:
> > I tried to `strace` it, and it seems that's still triggering the EACCES, even tough I'm root:
> > ```
> > # strace --follow-forks /home/mpdesouza/git/virtme-ng/virtme/guest/virtme-sshd-scr
> > ipt
> > 
> > .... <lots and lots of output>
> > [pid   415] newfstatat(AT_FDCWD, "/usr/etc/ssh/sshd_config", {st_mode=S_IFREG|0640, st_size=3727, ...}, 0) = 0
> > [pid   415] newfstatat(AT_FDCWD, "/etc/ssh/sshd_config", 0x7fffe0f8ac00, 0) = -1 ENOENT (No such file or directory)
> > [pid   415] openat(AT_FDCWD, "/usr/etc/ssh/sshd_config", O_RDONLY) = -1 EACCES (Permission denied)
> > [pid   415] dup(2)                      = 7
> > [pid   415] fcntl(7, F_GETFL)           = 0x2 (flags O_RDWR)
> > [pid   415] fstat(7, {st_mode=S_IFCHR|0600, st_rdev=makedev(0x88, 0), ...}) = 0
> > [pid   415] write(7, "/usr/etc/ssh/sshd_config: Permis"..., 44/usr/etc/ssh/sshd_config: Permission denied
> > ) = 44
> > [pid   415] close(7)                    = 0
> > [pid   415] exit_group(1)               = ?
> > [pid   415] +++ exited with 1 +++
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So I'm not sure what's going on... starting sshd on my host system as root works, so I'm not sure why this doesn't work...
> 
> Ah ok, makes sense, that's because sshd is trying to access `/usr/etc/ssh/sshd_config` on the host via virtiofsd, but virtiofsd (that runs as your regular user) probably doesn't have access to that file, hence EACCESS.
> 
> In my case I have the config in `/etc/ssh/sshd_config`, that is world readable. Maybe we can fix this by generating a custom sshd_config in tmp and bind mount to the real sshd_config if we don't have read access.

ahhh, ok, yes, it makes sense that virtiofsd runs as a normal user... thanks a lot for spotting this problem! I'll test your PR to check if that fixes my issue!

PR #212, opened 2024-12-31
sshd: generate a custom sshd_config

arighi:
Instead of relying on the host's sshd configuration, generate a custom sshd_config to make sure that all the options required by virtme-ng are enabled.


virtme/guest/virtme-sshd-script
```diff
@@ -24,18 +24,37 @@ SSH_AUTH_KEYS="${SSH_HOME}/.ssh/authorized_keys"
 if [ "$(stat -f -c "%t" "${SSH_AUTH_KEYS}")" = "${OVERLAYFS}" ]; then
     cat "${SSH_HOME}"/.ssh/id_*.pub >> "${SSH_AUTH_KEYS}" 2>/dev/null
     chown "${virtme_ssh_user}" "${SSH_AUTH_KEYS}" 2>/dev/null
+    chmod 600 "${SSH_HOME}/.ssh/authorized_keys" 2>/dev/null
```
matttbe, line 27: (detail: you can use `${SSH_AUTH_KEYS}`)

```suggestion
    chmod 600 "${SSH_AUTH_KEYS}" 2>/dev/null
```

virtme/guest/virtme-sshd-script
```diff
@@ -24,18 +24,37 @@ SSH_AUTH_KEYS="${SSH_HOME}/.ssh/authorized_keys"
...
+PermitRootLogin yes
+AuthorizedKeysFile .ssh/authorized_keys
+PubkeyAuthentication yes
+UsePAM yes
+PrintMotd no
+EOF
+fi
+
+# Start sshd.
 ARGS=()
```
matttbe, line 52: (detail: you can also modify this variable)

```suggestion
ARGS=(-f "${SSH_CONFIG}")
```

virtme/guest/virtme-sshd-script
```diff
@@ -24,18 +24,37 @@ SSH_AUTH_KEYS="${SSH_HOME}/.ssh/authorized_keys"
...
+    mkdir -p "${ssh_dir}"
+    cat << EOF > "${SSH_CONFIG}"
+# This file is automatically generated by virtme-ng.
+Port 22
+PermitRootLogin yes
+AuthorizedKeysFile .ssh/authorized_keys
+PubkeyAuthentication yes
+UsePAM yes
+PrintMotd no
+EOF
```
matttbe, line 48: detail: you can also use `-EOF` to keep the indentation after the `if` statement

```suggestion
    cat << -EOF > "${SSH_CONFIG}"
        # This file is automatically generated by virtme-ng.
        Port 22
        PermitRootLogin yes
        AuthorizedKeysFile .ssh/authorized_keys
        PubkeyAuthentication yes
        UsePAM yes
        PrintMotd no
    EOF
```

virtme/guest/virtme-sshd-script
```diff
@@ -24,18 +24,37 @@ SSH_AUTH_KEYS="${SSH_HOME}/.ssh/authorized_keys"
...
+fi
+
+# Start sshd.
 ARGS=()
 for key in "${CACHE_DIR}"/etc/ssh/ssh_host_*_key; do
     ARGS+=(-h "${key}")
 done
```
matttbe, line 55: (detail: this code could be kept next to `ssh-keygen -A`)

virtme/guest/virtme-sshd-script
```diff
@@ -24,18 +24,37 @@ SSH_AUTH_KEYS="${SSH_HOME}/.ssh/authorized_keys"
...
+if [ "$(stat -f -c "%t" "${SSH_CONFIG}")" = "${OVERLAYFS}" ]; then
+    ssh_dir=$(dirname "${SSH_CONFIG}")
+    mkdir -p "${ssh_dir}"
+    cat << EOF > "${SSH_CONFIG}"
+# This file is automatically generated by virtme-ng.
+Port 22
+PermitRootLogin yes
+AuthorizedKeysFile .ssh/authorized_keys
+PubkeyAuthentication yes
+UsePAM yes
```
matttbe, line 46: I wonder if this option can be removed if there are no passwords (same for the two above). I guess all Linux distributions have PAM support.

virtme/guest/virtme-sshd-script
```diff
@@ -24,18 +24,37 @@ SSH_AUTH_KEYS="${SSH_HOME}/.ssh/authorized_keys"
...
 fi
 
 # Generate ssh host keys (if they don't exist already).
 CACHE_DIR=${SSH_HOME}/.cache/virtme-ng/.ssh
 mkdir -p "${CACHE_DIR}/etc/ssh"
 ssh-keygen -A -f "${CACHE_DIR}"
+
+# Generate a minimal sshd config.
+SSH_CONFIG=/etc/ssh/sshd_config
+if [ "$(stat -f -c "%t" "${SSH_CONFIG}")" = "${OVERLAYFS}" ]; then
```
matttbe, line 37: Why not putting this file elsewhere? e.g. in `/run` or similar? Maybe an issue with the permissions?

matttbe (approved):
Good idea! That will be easier to maintain than having to deal with the different config options!

I have a few suggestions, and questions, but globally it looks good to me.

marcosps:
I'm always messing with the virtme-ng-init (somehow when using `pipx install --force` doesn't recompile or reinstall it), but with the fix that I mentioned above (dirname), it works for me!

One thing is that I needed to have my local ssh to the authorized_keys (and I didn't had this file in my machine because I don't connect to it to work, so ssh-copy-id solved this to me). With this fixed, I was able to connect to it!

marcosps (changes requested):
With the /etc/ssh/sshd_config it works for me!

matttbe:
> I'm always messing with the virtme-ng-init (somehow when using `pipx install --force` doesn't recompile or reinstall it)

Personally, I'm simply using `make` from the source directory, and execute the `vng` file from there to run the development versions. On my system, I only install stable versions with `pip`.

marcosps:
> > I'm always messing with the virtme-ng-init (somehow when using `pipx install --force` doesn't recompile or reinstall it)
> 
> Personally, I'm simply using `make` from the source directory, and execute the `vng` file from there to run the development versions. On my system, I only install stable versions with `pip`.

Makes sense. I'll do the same.

virtme/guest/virtme-sshd-script
arighi: oh nice! TIL

arighi: Hm... this doesn't seem to work for me. Is -EOF only removing tabs maybe?

virtme/guest/virtme-sshd-script
arighi: I think we need `UsePAM yes` for the pubkey authentication, at least in my case if I remove that it's asking me the password and it doesn't work...

virtme/guest/virtme-sshd-script
matttbe: > Hm... this doesn't seem to work for me. Is -EOF only removing tabs maybe?

oops yes, sorry, I'm usually using tabs :)

(also there should not be a space before the `-EOF`)

virtme/guest/virtme-sshd-script
matttbe: > I think we need `UsePAM yes` for the pubkey authentication, at least in my case if I remove that it's asking me the password and it doesn't work...

Yes indeed, we need it for the moment. But maybe we don't need it with https://github.com/arighi/virtme-ng-init/pull/16 ?

virtme/guest/virtme-sshd-script
matttbe: A workaround is to use:

```
  printf "%s\n" \
      "line 1" \
      "line 2" \
      ...
```

or `sed`. But not as well integrated than this `cat <<-EOF` :)

virtme/guest/virtme-sshd-script
arighi: > Why not putting this file elsewhere? e.g. in `/run` or similar? Maybe an issue with the permissions?

I was also thinking about that and also move authorized_keys in /run as well, so we don't have to deal about potentially overwriting the real authorized_keys on the host.

However, sshd is complaining if I try to move authorized_keys in /run:
```
Authentication refused: bad ownership or modes for directory /run
```

And we don't want to change the permissions of /run otherwise we may break other stuff. But I think it's possible to move sshd_config to /run, will do some tests.

arighi: Maybe we can move authorized_keys into the cache dir (where we store the host keys), so even if we rewrite authorize_keys on the host it won't be that bad, as it's basically a file generated in the virtme-ng cache.

marcosps (approved):
It worked for me, thanks a lot!

marcosps:
> Collaborator

As usual, it was

> > I'm always messing with the virtme-ng-init (somehow when using `pipx install --force` doesn't recompile or reinstall it)
> 
> Personally, I'm simply using `make` from the source directory, and execute the `vng` file from there to run the development versions. On my system, I only install stable versions with `pip`.

As usual, it was my fault. I should have updated the submodule (I'm not very used to work with submodules...), so once it was updates, make and "pipx install" was able to install the latest binary, and now I don't need to run it in the same dir :)

matttbe (approved):
Thank you for the update

matttbe:
And happy new year 🎉😊

arighi:
First merge of the year! :)

Issue #213, opened 2025-01-04
Simplify packaging


septatrix:
There are a few issues with the way virtme-ng-init is currently built, bundled and used. As a result many distros only ship virtme without the rust-based virtme-ng-init component.
The main problem is the submodule/git-fetch usage and the cargo invocation in setup.py. As a result offline compilation is made difficult, HTTP proxies fail to work, and simple control over cargo flags is not possible. These are all things common for package building and enterprise environments.

This situation could be improved by merging both repos, wrapping the init repo in a simple dummy python package and adding a dependency on that, or adding another search path for the init-ng binary outside of the package resources such that it could be build separately. Ideally this would also allow us to completely get rid of the setup.py file.

Please let me know your thoughts about these approaches. I myself have not yet completely explored the options so I am not yet sure which one would best suit packaging but they would all be an improvement over the status quo

matttbe:
Hello,

Thank you for this bug report!

I already wanted to suggest merging both repos to help with the development. If it can also help with the packaging (if I understood correctly), why not doing that!

To merge the two repo, we can use `git subtree add` to include the history as well.

SPYFF:
Right now virtme-ng has a dependency on a statically built busybox. I can imagine something like that on a statically built virtme-ng-init. That could be a separate package, one binary. Similarly to virtiofsd: if not found, fallback to 9p. However I have no idea how hard a static build of virtme-ng-inig would be, probably not that hard since its a relatively small program without complex dependencies.

A more extreme version would be a github CI wokflow, which produces reproducible builds of static virtme-ng-init binaries, and the python part would download & verify that binary if available. Sure, it would be much better to make available the init as a package on all distro, but maintainers might not find enough justification for that, since its only useful virtme-ng (and why not make part of it then, as you proposed).

arighi:
Yeah, the virtme-ng-init submodule is bugging me. At some point I'd like to merge the code directly into virtme-ng. It'd be nice to keep the history somehow (like re-applying all the commits with some scripting maybe).

Then there's the packaging, which I'm not sure how to solve honestly. Having a separate package for virtme-ng-init seems complicated, since it'll be used only virtme-ng. I like the idea of having a dummy python module, maybe this can help? https://pypi.org/project/setuptools-rust/

septatrix:
[...]

PR #216, opened 2025-01-06
Drop `virtme-ng-init` submodule and import commits instead

matttbe:
As discussed on #213, the `virtme-ng-init` submodule has been dropped, and the commits have been imported in the `virtme_ng_init` directory instead.

3 commits have been "manually" created here:
- 89e8fa1 ("virtme-ng-init: remove the submodule"), using:
  ```shell
  git rm virtme_ng_init && git commit -s
  ```
- d2b8a2b ("Add 'virtme_ng_init/' from commit '9b1e02a0deb094a36741b6172fb7ea8dc7dd8ddb'"), using:
  ```shell
  git subtree -P virtme_ng_init add git@github.com:arighi/virtme-ng-init.git main
  ```
- 2c4fd14 ("gh: import and adapt rust workflow")

The `git subtree add` command automatically created as many commits as there are in the `virtme-ng-init` repo, making this PR a bit hard to review. I guess only the commands used above, and the last commit needs to be reviewed.

Note: Because I re-used the same directory as the one previously used by the submodule, I suspect people syncing with the new version might have to do an additional step (to be verified), e.g. this one?

    git submodule deinit virtme_ng_init


arighi:
Hi @matttbe thanks for looking at this! I would be even more aggressive, import/rewrite all the commits and forget about the external virtme-ng-init repo.

I've tried to do some experiments and I came up with this:
https://github.com/arighi/virtme-ng/tree/arighi/vng-init-merge

This is what I've done, I cherry picked your first commit, then:
```
$ git clone git@github.com:arighi/virtme-ng-init.git temp
$ cd temp
$ git filter-repo --to-subdirectory-filter virtme_ng_init
$ cd ..
$ git remote add temp temp
$ git merge --allow-unrelated-histories temp/main
$ rm -rf temp
```

The thing that I like a lot about this is that we can `git log virtme_ng_init` and see the history of virtme-ng-init, like the real history, in particular we can see the real contributor of each commit (instead of the author that updated the submodule), which I think is eally nice. Also, in this way you last commit `gh: import and adapt rust workflow` is not needed (IIUC), we can archive/close virtme-ng-init and do all the development here.

What do you think?

matttbe:
> Hi @matttbe thanks for looking at this! I would be even more aggressive, import/rewrite all the commits and forget about the external virtme-ng-init repo.
> 
> I've tried to do some experiments and I came up with this: https://github.com/arighi/virtme-ng/tree/arighi/vng-init-merge
> 
> This is what I've done, I cherry picked your first commit, then:
> 
> ```
> $ git clone git@github.com:arighi/virtme-ng-init.git temp
> $ cd temp
> $ git filter-repo --to-subdirectory-filter virtme_ng_init
> $ cd ..
> $ git remote add temp temp
> $ git merge --allow-unrelated-histories temp/main
> $ rm -rf temp
> ```
> 
> The thing that I like a lot about this is that we can `git log virtme_ng_init` and see the history of virtme-ng-init, like the real history, in particular we can see the real contributor of each commit (instead of the author that updated the submodule), which I think is eally nice.

Oh, I'm surprised the result is not the same for the `git log virtme_ng_init` command.

With `git subtree`, the good thing is that we can keep the same SHA to follow the differences that were done on the submodule. But I understand we might not need that, and we might prefer to look at history with just `git log virtme_ng_init`. Also, we will not need to do any sync later on, so we don't need the exact same SHA. It is of course fine for me to use your technique!

> Also, in this way you last commit `gh: import and adapt rust workflow` is not needed (IIUC), we can archive/close virtme-ng-init and do all the development here.

I think it is still needed to move the workflow file from `.virtme_ng_init/.github` to `.github`, no? Or can you keep it in a subdirectory (`virtme_ng_init`) as it is in your [version](https://github.com/arighi/virtme-ng/blob/arighi/vng-init-merge/virtme_ng_init/.github/workflows/rust.yml)?

Do you plan to cherry-pick the last commit and open a PR replacing this one? (you can also override my branch if you prefer)

arighi:
> > Also, in this way you last commit `gh: import and adapt rust workflow` is not needed (IIUC), we can archive/close virtme-ng-init and do all the development here.
> 
> I think it is still needed to move the workflow file from `.virtme_ng_init/.github` to `.github`, no? Or can you keep it in a subdirectory (`virtme_ng_init`) as it is in your [version](https://github.com/arighi/virtme-ng/blob/arighi/vng-init-merge/virtme_ng_init/.github/workflows/rust.yml)?
> 
> Do you plan to cherry-pick the last commit and open a PR replacing this one? (you can also override my branch if you prefer)

Ah! You are absolutely right! I cherry-picked your last commit and updated my branch.

arighi:
I really like that we can preserve the git history with the original author and it'd be nice to keep this PR instead of opening a new one, maybe it's easier if I just force push to your branch instead of opening a new PR?

matttbe:
> maybe it's easier if I just force push to your branch?

Yes, that sounds better. You should have the right to modify my branch on my repo, no?

arighi:
> > maybe it's easier if I just force push to your branch?
> 
> Yes, that sounds better. You should have the right to modify my branch on my repo, no?

Ah no, I don't think so, because the branch is in your workspace.

BTW maybe at some point we should also move the repo to https://github.com/virtme/ so that we can all create branches and stuff in the same workspace.

arighi:
@matttbe maybe you can force push my branch to your branch and then we can merge :)

arighi:
oh wait, nevermind, I just tried and it worked, I just forced pushed to your branch... what? :-o

matttbe:
It looks like it worked 🙂

> BTW maybe at some point we should also move the repo to https://github.com/virtme/ so that we can all create branches and stuff in the same workspace.

Yes, good idea! (Or `virtme-ng`)
With the migration option in the repo, people will be redirected to the new repo, so it sounds safe.

matttbe:
> oh wait, nevermind, I just tried and it worked, I just forced pushed to your branch... what? :-o

Yes, that's linked to the "Allow edits and access to secrets by maintainers" option 🙂

arighi:
> > oh wait, nevermind, I just tried and it worked, I just forced pushed to your branch... what? :-o
> 
> Yes, that's linked to the "Allow edits and access to secrets by maintainers" option 🙂

I was a bit surprised to see that I could just destroy your work lol.

About moving the project to a virtme-ng organization, I didn't know there was a migrate option, that's very interesting! I think it'd make more sense to migrate the project outside of my personal workspace, considering that we have multiple contributors now. I'll take a look!

arighi (approved):

matttbe:
> I was a bit surprised to see that I could just destroy your work lol.

I still have a local copy 🙂

@arighi: before merging, did you have issues to switch to my branch when you had the submodule still configured? Just to know if we have to tell people what to do? 🙂

arighi:
> @arighi: before merging, did you have issues to switch to my branch when you had the submodule still configured? Just to know if we have to tell people what to do? 🙂

Ah yes, good point, I had to manually remove the `virtme_ng_init` dir and reset to your branch. Probably people will need to do the same with this PR. I'm not sure if there's a clean way to prevent this... I guess people will just have to fix this manually (or clone the repo again).

matttbe:
> > @arighi: before merging, did you have issues to switch to my branch when you had the submodule still configured? Just to know if we have to tell people what to do? 🙂
> 
> Ah yes, good point, I had to manually remove the `virtme_ng_init` dir and reset to your branch. Probably people will need to do the same with this PR. I'm not sure if there's a clean way to prevent this... I guess people will just have to fix this manually (or clone the repo again).

Should I then just merge and see? 🙂

If someone has the issue, maybe the proper solution is to execute:

    git submodule deinit virtme_ng_init

arighi:
Let's go ahead and merge, people will figure out what to do. If not they can ask. :)

PR #218, opened 2025-01-23
virtme_ng_init: use uzers instead of users (which is unmaintained)

WOnder93:
See:
- https://rustsec.org/advisories/RUSTSEC-2023-0040.html
- https://bugzilla.redhat.com/show_bug.cgi?id=2212194


matttbe (approved):
Thank you for the modification!

@arighi: is there not something that can be turned on in the settings to automatically check the dependencies? See: https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories

arighi:
> Thank you for the modification!
> 
> @arighi: is there not something that can be turned on in the settings to automatically check the dependencies? See: https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories

That seems interesting, will take a look!

PR #220, opened 2025-01-23
virtme-ng-init: always check return code when executing bash

arighi:
Fix the following warning:

warning: unused return value of `exec` that must be used
   --> src/main.rs:978:13
    |
978 |             Command::new("bash").arg("-l").exec();
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default

Moreover, check the return code in case exec() fails and log an explicit error before exiting.


matttbe (approved):
thank you for this fix!

PR #222, opened 2025-01-24
properly quoting exec string to avoid breaking virtme-run invocation …

### LINKED ISSUE #221: command passed through --exec to --script-sh needs escaping
the command string passed through --exec to virtme-ng is enclosed in double quotes before being passed to virtme-run, but also needs to be quote-escaped. Otherwise, quote characters contained in the command will break the constructed virtme-run command, and may even lead to commands destined for execution in the guest instead being injected on the host, if the command string additionally contains semicolons or other shell control characters.

example:

```
$ vng [...] --exec "false\"; uname -n\""
You must specify a kernel to use.
[hostname of the host]
/bin/sh: line 1: : command not found
```

in this example, the virtme-run command will fail because it is terminated early by the injected semicolon, then uname is executed on the _host_, and then sh will complain about the rest of the virtme command line which it percieves as garbage.

Beyond this undesired side effect, not escaping the exec parameter in virtme_ng makes it harder to correctly quote the command string to be run in the guest, and consequently unnecessarily difficult to specify more complex guest commands.

The fix is simple:

```
From ecd69ab3d86f09d9a52837b30f0b75c7bd8e7d4e Mon Sep 17 00:00:00 2001
From: Andreas Grapentin <andreas.grapentin@ibm.com>
Date: Fri, 24 Jan 2025 08:24:16 +0100
Subject: [PATCH] properly quoting exec string to avoid breaking virtme-run
 invocation on more complex guest calls

---
 virtme_ng/run.py | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

index ec7aa63..10966ce 100644
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -22,6 +22,7 @@ from subprocess import (
 )
 from select import select
 from pathlib import Path
+import shlex
 try:
     from argcomplete import autocomplete
 except ModuleNotFoundError:
@@ -784,7 +785,7 @@ class KernelSource:
         if args.envs:
             args.exec = " ".join(args.envs)
         if args.exec is not None:
-            self.virtme_param["exec"] = f'--script-sh "{args.exec}"'
+            self.virtme_param["exec"] = f'--script-sh {shlex.quote(args.exec)}'
         else:
             self.virtme_param["exec"] = ""

--
2.47.0
```

virtme-run itself also uses `shlex.quote` to escape arguments passed to qemu.

Note: It's possible that other parameters of virtme_ng need the same treatment, but I haven't checked in detail, which is why this is not a PR.

andreasgrapentin:
as requested, this PR fixes #221 by replacing the double quotes in the command string passed through --exec with a call to `shlex.quote`

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -23,6 +23,7 @@ from subprocess import (
 )
 from select import select
 from pathlib import Path
+import shlex
 
 import argcomplete
 
@@ -876,7 +877,7 @@ class KernelSource:
         if args.envs:
             args.exec = " ".join(args.envs)
         if args.exec is not None:
-            self.virtme_param["exec"] = f'--script-sh "{args.exec}"'
+            self.virtme_param["exec"] = f'--script-sh {shlex.quote(args.exec)}'
         else:
             self.virtme_param["exec"] = ""
 


arighi (approved):
LGTM, thanks!

PR #224, opened 2025-01-27
Code simplifications

devnexen:


devnexen:
2 separate commits in case you disagree with the second change.

devnexen:
I knew that would provoke this kind of reaction :) it s just make the intention clearer but again do not mind removing these changes.

arighi (commented):
Can we rephrase saying that is_err() is more idiomatic? It's not really a simplification.

arighi (commented):
Why do we need this? Can you motivate more in the commit description?

Issue #223, opened 2025-01-27
vng does not support --arch x86_64 for running on non-x86 machines


ctmarinas:
I normally run Linux on arm64 machines and it turns out that vng does not support compiling/running x86 targets. My hack was simple, just add the hunk below to `ARCH_MAPPING` in `virtme_ng/run.py`:

```
    "x86_64": {
        "qemu_name": "x86_64",
        "linux_name": "x86_64",
        "cross_compile": "x86_64-linux-gnu-",
        "kernel_target": "bzImage",
        "kernel_image": "bzImage",
    },
    "i386": {
        "qemu_name": "i386",
        "linux_name": "i386",
        "cross_compile": "i686-linux-gnu-",
        "kernel_target": "bzImage",
        "kernel_image": "bzImage",
    },
```

I just picked the Linux target names, `x86_64` and `i386`. I haven't tested `--arch=i386` and I suspect virtme-ng doesn't support clang either.

Thanks.

arighi:
@ctmarinas thanks for opening this, #225 seems to work for me (at least on arm64).

I'm not sure about introducing i386 support, as we might be missing features, like initializing a rootfs, since Ubuntu cloud images don't provide the i386 variants anymore and probably other stuff might be broken as well.

matttbe:
#225 is now merged: it is now possible to use `--arch=amd64` on non-x86 platforms. We can then close this issue.

PR #225, opened 2025-01-29
virtme-ng: enable amd64 architecture support on non-x86 platforms

arighi:
As mentioned in issue #223, there is no reason to forbid the use of x86_64 as a non-native architecture, so add the proper definition to support x86_64 (building kernel and starting instances) on non-x86 machines.

For consistency with Debian/Ubuntu naming conventions used for other architectures, use amd64 to identify x86_64.

Link: https://lore.kernel.org/all/426d4476-e3b4-4a95-84a1-850015651ee6@kernel.org/
Suggested-by: Catalin Marinas <catalin.marinas@arm.com>

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -552,6 +552,13 @@ def arg_fail(message, show_usage=True):
 
 
 ARCH_MAPPING = {
+    "amd64": {
+        "qemu_name": "x86_64",
+        "linux_name": "x86_64",
+        "cross_compile": "x86_64-linux-gnu-",
+        "kernel_target": "bzImage",
+        "kernel_image": "bzImage",
+    },
     "arm64": {
         "qemu_name": "aarch64",
         "linux_name": "arm64",


arighi:
Successfully tested only on arm64 for now.

matttbe:
> > Thank you for having looked at that, and for having tested it!
> > > For consistency with Debian/Ubuntu naming conventions used for other architectures, use amd64 to identify x86_64.
> > 
> > 
> > Should we eventually support both, or only `x86_64`? I mean, I understand it is called `amd64` for historical reasons, but this feels wrong today. Especially in the Linux kernel development where `x86_64` is used and not `amd64` :)
> > Or do you need this name to be able to get Ubuntu cloud images?
> 
> I'm a bit conflicted about the naming. Historically virtme follows the Debian naming schema, so from this perspective using amd64 makes more sense. But it also makes sense to use x86_64, which is the kernel naming.
> 
> Probably the best is to keep using amd64 internally for now (so that we don't break anything) and provide an x86_64 "alias" that can be simply translated into amd64 internally.

Fine by me. Using `amd64` is a good first step.

> > > Successfully tested only on arm64 for now.
> > 
> > 
> > Not directly related to ↑, but do you think there is a need to run the tests on arm64 as well? It is possible with [GitHub Actions](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories), apparently, but it doesn't seem critical to have (it's not like the current tests were already covering a wide range of features :) ).
> 
> Yeah, we should write better unit tests in general, but It wouldn't be bad to try booting at least a precompiled kernel also on arm64, considering that quite a few people are now using vng on this arch.

Indeed. But there are also so many ways of using `virtme`, it is hard to cover everything :)
The good thing is that this tool is used by multiple CIs, so "stable" versions should be quite well tested :-D

Issue #217, opened 2025-01-10
vng silently hangs if boot fails


bjackman:
I have a system failing like `[    2.073575] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00007f00`.

I can see this if I pass `--verbose`, but if I just run `vng`, I get no error messages.

My initial feeling was that the kernel console should be included in the default output regardless of `--verbose` but I suspect there's a good reason why it isn't.

Another idea I had would be to combine the kernel's `panic=-1` with QEMU's `-no-reboot`. But then that breaks the usecase of people who genuinely want to reboot (is that a valid usecase? I guess so?).

Basically, I think the ideal behaviour is that if the kernel panics before we hit `virtme-init` , we spit out the console and exit with an error.

So... any ideas?

SPYFF:
Version?

matttbe:
Hello,

@bjackman : which mode are you using? Interactive or script mode?
In script mode, the VM should stop and not panic, see #70. (with `panic=-1` and `-no-reboot`).

In interactive mode, I agree it is unclear what can be done when it crashes before the init script. Maybe a timeout for the init script? In case of timeout, a message could be printed to use the verbose mode to see what's wrong?

bjackman:
> Version?

```
❯❯  vng --version
virtme-ng 1.31+43.g67cac60
```

> @bjackman : which mode are you using? Interactive or script mode?

I'm guessing script mode means when you pass a command after `--`? I'm not doing that, so I guess it's interactive mode.

> Maybe a timeout for the init script? In case of timeout, a message could be printed to use the verbose mode to see what's wrong?

Yeah - although I think before adding that complexity it would be good to confirm my assumption that setting `-no-reboot` would break a usecase that we care about.

I guess the ideal would be if there was a way to start QEMU with the `-no-reboot` behaviour and then have the init script communicate via the vsock or whatever it is that QGA uses, to tell QEMU "OK, now you're allowed to reboot instead of exit". But that's probably more complicated than just having the timeout + print message.

matttbe:
> > Version?
> 
> ```
> ❯❯  vng --version
> virtme-ng 1.31+43.g67cac60
> ```
> 
> > [@bjackman](https://github.com/bjackman) : which mode are you using? Interactive or script mode?
> 
> I'm guessing script mode means when you pass a command after `--`? I'm not doing that, so I guess it's interactive mode.

Yes sorry, everything passed after `--` or using `--exec`

[...]

PR #227, opened 2025-01-30
virtme: always print kernel panic / oops in interactive mode

arighi:
As mentioned in #217 it would be nice to always print the output of critical kernel errors (oops / panic), instead of suppressing all the kernel logs completely by default.

Therefore, keep suppressing the boot kernel log, but always dump panic/oops to stderr by default when running in interactive mode.

Example with this change applied:

```
arighi@virtme-ng~/s/linux (master)> vng
          _      _
                                                |___/
   kernel version: 6.13.0-virtme x86_64
   (CTRL+d to exit)

arighi@virtme-ng~/s/linux (master)> echo c | sudo tee /proc/sysrq-trigger
[    8.923672] sysrq: Trigger a crash
[    8.923980] Kernel panic - not syncing: sysrq triggered crash
[    8.924183] CPU: 0 UID: 0 PID: 198 Comm: tee Not tainted 6.13.0-virtme #2
[    8.924632] Call Trace:
[    8.924704]  <TASK>
[    8.924783]  panic+0x349/0x3b0
[    8.925055]  sysrq_handle_crash+0x36/0x80
[    8.925181]  __handle_sysrq+0xed/0x270
[    8.925274]  write_sysrq_trigger+0x6a/0x90
[    8.925380]  proc_reg_write+0x56/0xa0
[    8.925489]  vfs_write+0x105/0x590
[    8.925600]  ksys_write+0x74/0xf0
[    8.925682]  do_syscall_64+0xbb/0x1d0
[    8.925767]  entry_SYSCALL_64_after_hwframe+0x77/0x7f
[    8.925891] RIP: 0033:0x7fdfed54ba84
```

Closes: #217

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1282,25 +1282,28 @@ def do_it() -> int:
     if args.graphics is None and not args.script_sh and not args.script_exec:
         qemuargs.extend(["-echr", "1"])
 
-        if args.verbose:
-            # Check if we have permission to access the current stderr.
-            if not can_access_file("/proc/self/fd/2"):
-                print(
-                    "ERROR: not a valid pts, try to run vng inside tmux or screen",
-                    file=sys.stderr,
-                )
-                sys.exit(1)
-
-            # Redirect kernel messages to stderr, creating a separate console
+        # Redirect kernel errors to stderr, creating a separate console.
+        #
+        # If we don't have access to stderr via procfs (for example when
+        # running inside a container), print a warning and implicitly
+        # suppress the kernel errors redirection.
+        if can_access_file("/proc/self/fd/2"):
             qemuargs.extend(["-chardev", "file,path=/proc/self/fd/2,id=dmesg"])
             qemuargs.extend(["-device", arch.virtio_dev_type("serial")])
             qemuargs.extend(["-device", "virtconsole,chardev=dmesg"])
             kernelargs.extend(["console=hvc0"])
+        else:
+            print(
+                "WARNING: unable to write kernel messages, try to run vng with a valid PTS "
+                "(e.g., inside tmux or screen)",
+                file=sys.stderr,
+            )
 
-            # Unfortunately we can't use hvc0 to redirect early console
-            # messages to stderr, so just send them to the main console, in
-            # this way we don't lose early printk's in verbose mode and we can
-            # catch potential boot issues.
+        # Unfortunately we can't use hvc0 to redirect early console
+        # messages to stderr, so just send them to the main console, in
+        # this way we don't lose early printk's in verbose mode and we can
+        # catch potential boot issues.
+        if args.verbose:
             kernelargs.extend(arch.earlyconsole_args())
 
         qemuargs.extend(["-chardev", "stdio,id=console,signal=off,mux=on"])
@@ -1407,7 +1410,8 @@ def do_it() -> int:
            not can_access_file("/proc/self/fd/1") or \
            not can_access_file("/proc/self/fd/2"):
             print(
-                "ERROR: not a valid pts, try to run vng inside tmux or screen",
+                "ERROR: not a valid pts, try to run vng with a valid PTS "
+                "(e.g., inside tmux or screen)",
                 file=sys.stderr,
             )
             sys.exit(1)
@@ -1671,11 +1675,11 @@ def do_it() -> int:
         )
         initrdpath = None
 
-    if not args.verbose:
-        kernelargs.append("quiet")
-        kernelargs.append("loglevel=0")
-    else:
+    if args.verbose:
         kernelargs.append("debug")
+    else:
+        kernelargs.append("quiet")
+        kernelargs.append("loglevel=1")
 
     # Now that we're done setting up kernelargs, append user-specified args
     # and then initargs


virtme/commands/run.py
```diff
@@ -1282,25 +1282,25 @@ def do_it() -> int:
...
-        if args.verbose:
-            # Check if we have permission to access the current stderr.
-            if not can_access_file("/proc/self/fd/2"):
-                print(
-                    "ERROR: not a valid pts, try to run vng inside tmux or screen",
-                    file=sys.stderr,
-                )
-                sys.exit(1)
+        # Check if we have permission to access the current stderr.
+        if not can_access_file("/proc/self/fd/2"):
```
matttbe, line 1286: Do you know when this error could happen?

I mean: if it can happen in "normal" situations, maybe no need to exit except in verbose mode? (so the previous `if args.verbose` could be replaced by `if can_access_file("/proc/self/fd/2")` followed by a `elif args.verbose: print(); exit();`

matttbe (commented):
Good idea to change the log level in non-verbose mode!

Just one question to avoid exiting if stderr is not accessible.

virtme/commands/run.py
arighi: > Do you know when this error could happen?

Oh that's a good point, yes, it can happen when running vng in certain containers.

> 
> I mean: if it can happen in "normal" situations, maybe no need to exit except in verbose mode? (so the previous `if args.verbose` could be replaced by `if can_access_file("/proc/self/fd/2")` followed by a `elif args.verbose: print(); exit();`

Hm... but in this way we would implicitly suppress the errors when stderr isn't available. Maybe we can do that and also print a warning to stdout that we're implicitly suppressing errors because stderr isn't accessible?

matttbe: > > Do you know when this error could happen?
> 
> Oh that's a good point, yes, it can happen when running vng in certain containers.

OK, then probably best not to break this.

> > I mean: if it can happen in "normal" situations, maybe no need to exit except in verbose mode? (so the previous `if args.verbose` could be replaced by `if can_access_file("/proc/self/fd/2")` followed by a `elif args.verbose: print(); exit();`
> 
> Hm... but in this way we would implicitly suppress the errors when stderr isn't available. Maybe we can do that and also print a warning to stdout that we're implicitly suppressing errors because stderr isn't accessible?

Good point!
Or maybe in this case, it is fine to print the panic messages to stdout? These messages should be exceptional, no?

arighi: > > Hm... but in this way we would implicitly suppress the errors when stderr isn't available. Maybe we can do that and also print a warning to stdout that we're implicitly suppressing errors because stderr isn't accessible?
> 
> Good point! Or maybe in this case, it is fine to print the panic messages to stdout? These messages should be exceptional, no?

Alright, I've updated the PR, now we're never exiting, if we don't have access to stderr via procfs (e.g., when we run vng inside a container with restricted permissions), we print a warning, instead of an error, and suppress the kernel messages redirection. Otherwise, if we have access to stderr via procfs, we redirect only the panic/oops trace to stderr by default and with verbose we redirect all the kernel log messages to stderrr.

I think in this way we won't break anything, we can get the panic/oops when we run interactively and with `--verbose` everything remains the same.

virtme/commands/run.py
```diff
@@ -1282,25 +1282,27 @@ def do_it() -> int:
...
+        # supress the kernel errors redirection.
+        if can_access_file("/proc/self/fd/2"):
             qemuargs.extend(["-chardev", "file,path=/proc/self/fd/2,id=dmesg"])
             qemuargs.extend(["-device", arch.virtio_dev_type("serial")])
             qemuargs.extend(["-device", "virtconsole,chardev=dmesg"])
             kernelargs.extend(["console=hvc0"])
+        else:
+            print(
+                "WARNING: not a valid pts, try to run vng inside tmux or screen",
+                file=sys.stderr,
```
matttbe, line 1298: just to be sure: can we still write to stderr here if we cannot access `/proc/self/fd/2`?

Also, compared to before, this warning will always be printed in such environment, when stderr is not accessible. I guess it will not be an issue, right?

virtme/commands/run.py
```diff
@@ -1282,25 +1282,27 @@ def do_it() -> int:
...
-                )
-                sys.exit(1)
-
-            # Redirect kernel messages to stderr, creating a separate console
+        # Redirect kernel errors to stderr, creating a separate console.
+        #
+        # If we don't have access to stderr via procfs (for example when
+        # running inside a container), print a warning and implicitly
+        # supress the kernel errors redirection.
+        if can_access_file("/proc/self/fd/2"):
```
matttbe, line 1290: Just to be sure: if stderr is not available, should we not try to use stdout? Or do you think it can break some environments reading stdout? (if only panic are displayed, it should eventually be fine?)

virtme/commands/run.py
```diff
@@ -1282,25 +1282,27 @@ def do_it() -> int:
...
-                    file=sys.stderr,
-                )
-                sys.exit(1)
-
-            # Redirect kernel messages to stderr, creating a separate console
+        # Redirect kernel errors to stderr, creating a separate console.
+        #
+        # If we don't have access to stderr via procfs (for example when
+        # running inside a container), print a warning and implicitly
+        # supress the kernel errors redirection.
```
matttbe, line 1289: (just in case you have something else to modify)

```suggestion
        # suppress the kernel errors redirection.
```

matttbe (approved):
Thank you for the new version!

It looks good to me! I just have a couple of questions, just to be sure we don't want to look at alternatives. If not, feel free to merge :)

virtme/commands/run.py
arighi: > Just to be sure: if stderr is not available, should we not try to use stdout? Or do you think it can break some environments reading stdout? (if only panic are displayed, it should eventually be fine?)

Ok, now I remember exactly what was the issue. TL;DR the warning is printed correctly to stderr even when stderr is not accessible via procfs/devfs. :)

Details below.

Let's say you run as `userA`, then you `sudo su - userB` and `userB` is not in the `tty` group. A new pts is created but `userB` doesn't have access to it directly (since the new pts has `root:tty` ownership). Now if you open a tmux / screen session another pts is created setting the right ownership to `userB:tty`.

See for example:
```
[greeter@gpd3 /]$ groups
greeter video

# This is right after `sudo su - greeter`

[greeter@gpd3 /]$ ls -l /proc/self/fd/2
lrwx------ 1 greeter greeter 64 30 gen 15.03 /proc/self/fd/2 -> /dev/pts/7
[greeter@gpd3 /]$ ls -l /dev/pts/7
crw--w---- 1 root tty 136, 7 30 gen 15.03 /dev/pts/7

# This is inside tmux

greeter@gpd3 /]$ ls -l /proc/self/fd/2
lrwx------ 1 greeter greeter 64 30 gen 15.02 /proc/self/fd/2 -> /dev/pts/9
[greeter@gpd3 /]$ ls -l /dev/pts/9
crw--w---- 1 greeter tty 136, 9 30 gen 15.02 /dev/pts/9
```

So, in the first case I get the following (the warning is printed correctly and kernel errors are suppressed):
```
[greeter@gpd3 virtme-ng]$ ./vng -r
WARNING: not a valid pts, try to run vng inside tmux or screen
          _      _
                                                |___/
   kernel version: 6.13.0-1-cachyos-eevdf x86_64
   (CTRL+d to exit)
[greeter@virtme-ng virtme-ng]$ echo c | sudo tee /proc/sysrq-trigger
c
... hang....
```

Inside tmux:
```
[greeter@gpd3 virtme-ng]$ ./vng -r                          15:10:35 [37/37]
          _      _
                                                |___/
   kernel version: 6.13.0-1-cachyos-eevdf x86_64
   (CTRL+d to exit)

[greeter@virtme-ng virtme-ng]$ echo c | sudo tee /proc/sysrq-trigger
[    7.231129] sysrq: Trigger a crash
[    7.231222] Kernel panic - not syncing: sysrq triggered crash
...
```

virtme/commands/run.py
arighi: Fixed, thanks! :)

And I also slightly changed the warning to clarify that kernel errors are suppressed if we can't find a valid pts.

virtme/commands/run.py
matttbe: Thank you for having checked this and for the detailed answer! OK so if the access is not possible, it is still OK to write a warning without disturbing script monitoring stdout!

virtme/commands/run.py
```diff
@@ -1282,25 +1282,27 @@ def do_it() -> int:
...
+        # running inside a container), print a warning and implicitly
+        # suppress the kernel errors redirection.
+        if can_access_file("/proc/self/fd/2"):
             qemuargs.extend(["-chardev", "file,path=/proc/self/fd/2,id=dmesg"])
             qemuargs.extend(["-device", arch.virtio_dev_type("serial")])
             qemuargs.extend(["-device", "virtconsole,chardev=dmesg"])
             kernelargs.extend(["console=hvc0"])
+        else:
+            print(
+                "WARNING: kernel errors suppressed, try to run vng inside tmux or screen",
```
matttbe, line 1297: (A small suggestion, but also fine without → Feel free to merge with or without it ;-) )
```suggestion
                "WARNING: unable to write kernel messages, try to run vng with a valid PTS, e.g. inside tmux or screen",
```

matttbe, line 1297: (A small suggestion, but also fine without → Feel free to merge with or without it ;-) )
```suggestion
                "WARNING: unable to write kernel messages, try to run vng with a valid PTS, e.g. inside tmux or screen",
```

matttbe (approved):

virtme/commands/run.py
arighi: Ok, we also have a similar error in script mode (and in that case I think it's fine to error). I'll update both messages.

arighi: Ok, we also have a similar error in script mode (and in that case I think it's fine to error). I'll update both messages.

matttbe (approved):
thanks for the update!

matttbe (approved):
thanks for the update!

PR #229, opened 2025-02-06
Support `vng O=` in interactive mode

matttbe:
When looking at the discussion #228, I noticed `vng O=my_dir/` was not working as expected: with `-v`, I could see the VM booting, then stopping right after.

That's because `virtme-run` was called with `--script-sh "O=my_dir/" --kdir=./my_dir/`, so simply setting an env var and exiting.

Now `virtme-run` will be called with `--kdir=my_dir/`, so without `--script-sh` and without the `./` prefix to allow full paths.


arighi (approved):
Looks good to me, thanks!

Issue #230, opened 2025-02-07
`--debug` tampers with `--qemu-opts`


manufactory:
**tl;dr**
When running with the `--debug` flag one space character is missing in the `--qemu-opts` passed on the command line

**Versions**
I tested this with virtme-ne `1.32` *and* `virtme-ng 1.32+133.g8cb14a8` (which should be the latest commit at time of writing)

**How to reproduce**
```
vng --debug  --qemu-opts="-device vhost-vsock-pci,guest-cid=123"
```
fails with
```
qemu-system-x86_64: vmcoreinfo: Could not open 'vmcoreinfo': No such file or directory
```

**Confirming with `--dry-run`**

`--dry-run` shows the problem (I shortened the full command for clarity and to keep me precious file paths non-public)

* *without* extra space the next `-device` command gets concatenated to `123-device`, i.e.,
```
-device vhost-vsock-pci,guest-cid=123-device vmcoreinfo
```
* *with* extra space the `--qemu-opts` get passed properly 
```
-device vhost-vsock-pci,guest-cid=123 -device vmcoreinfo
```

**Workarounds**
* Without `--debug`, `--qemu-opts` work as expected.
* Add a single space character after `123`
```
vng --debug  --qemu-opts="-device vhost-vsock-pci,guest-cid=123 " # <-- once space between 3 and "
```

matttbe:
Hello,

@manufactory thank you for having reported the issue with enough details to directly understand what was wrong.

If you have the opportunity, can you try with #231 ?

BTW, if you are interested in using VSOCK, maybe you would be interested in the new `--console` option?

manufactory:
> [@manufactory](https://github.com/manufactory) thank you for having reported the issue with enough details to directly understand what was wrong.
Thank you for fixing this in virtually no time @matttbe !

> If you have the opportunity, can you try with [#231](https://github.com/arighi/virtme-ng/pull/231) ?
I can confirm the fix

> BTW, if you are interested in using VSOCK, maybe you would be interested in the new `--console` option?
Thanks for the hint :)

PR #231, opened 2025-02-07
vng: support --debug with --qemu-opts

matttbe:
When running with both `--debug` and `--qemu-opts` flags, one space character was missing in the `--qemu-opts` passed on the command line to `virtme-run`.

This could be fixed by adding a space as a prefix for the QEmu arguments added for the `--debug` case, but it sounds better to add the `--qemu-opts` flags at the end, so the debug args don't take precedence.

Closes: #230
Reported-by: @manufactory


arighi (approved):
LGTM!

PR #232, opened 2025-02-07
vng: support --nvgpu and --qemu-opts

matttbe:
When passing arguments to `virtme-run`, `--qemu-opts` has to be the last one, because it will take all args set after that.

In other words, `--nvgpu` was not supposed to be passed after `--qemu-opts`.

While at it, re-order functions and add a comment to avoid re-doing the same mistake.

Note that an alternative would be to call `virtme-run` with `--qemu-opt` instead of `--qemu-opts` (with 's'). But a comment should be enough.

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -1200,6 +1200,12 @@ class KernelSource:
             cpus = args.cpus
         self.virtme_param["cpus"] = f"--cpus {cpus}"
 
+    def _get_virtme_nvgpu(self, args):
+        if args.nvgpu is not None:
+            self.virtme_param["nvgpu"] = f"--nvgpu 'vfio-pci,host={args.nvgpu}'"
+        else:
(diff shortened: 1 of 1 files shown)


virtme_ng/run.py
```diff
@@ -1305,9 +1305,10 @@ def run(self, args):
             + f'{self.virtme_param["gdb"]} '
             + f'{self.virtme_param["snaps"]} '
             + f'{self.virtme_param["busybox"]} '
+            + f'{self.virtme_param["nvgpu"]} '
             + f'{self.virtme_param["qemu"]} '
             + f'{self.virtme_param["qemu_opts"]} '
-            + f'{self.virtme_param["nvgpu"]} '
+            # Important: qemu_opts has to be the last one
```
arighi, line 1311: Thanks for adding this comment!

arighi (approved):

Issue #233, opened 2025-02-08
Kernel panic on ARM64: "Requested init virtme-init failed (error -8)"


visitorckw:
Kernel panics when running `virtme-ng` to test an ARM64 Linux kernel on an x86-64 host. The issue occurs when `virtme-init` fails to execute with error `-8`, leading to a kernel panic.  

## Steps to reproduce  
1. Build the ARM64 kernel:  
   ```sh
   vng --build --arch arm64
   ```  
2. Run the virtual machine:  
   ```sh
   vng --arch arm64 --cpus 8
   ```  

## Observed behavior  
Boot fails with the following panic message:  
```
[    0.657709] Kernel panic - not syncing: Requested init /home/visitorckw/.local/lib/python3.10/site-packages/virtme/guest/virtme-init failed (error -8).
[    0.658725] CPU: 6 UID: 0 PID: 1 Comm: swapper/0 Not tainted 6.14.0-rc1-virtme #1
[    0.658863] Hardware name: linux,dummy-virt (DT)
[    0.659045] Call trace:
[    0.659204]  show_stack+0x18/0x24 (C)
[    0.659601]  dump_stack_lvl+0x38/0x90
[    0.659700]  dump_stack+0x18/0x24
[    0.659745]  panic+0x388/0x3e0
[    0.659801]  kernel_init+0xd0/0x1dc
[    0.659848]  ret_from_fork+0x10/0x20
[    0.660111] SMP: stopping secondary CPUs
[    0.660588] Kernel Offset: 0x346f67600000 from 0xffff800080000000
[    0.660649] PHYS_OFFSET: 0xfff0f4b080000000
[    0.660688] CPU features: 0x108,0002013c,00800000,0200420b
[    0.660743] Memory Limit: none
[    0.660978] ---[ end Kernel panic - not syncing: Requested init /home/visitorckw/.local/lib/python3.10/site-packages/virtme/guest/virtme-init failed (error -8). ]---
```

## System information  
- **Host architecture:** x86-64  
- **Target architecture:** ARM64  
- **Kernel version:** 6.14.0-rc1  
- **virtme-ng version:** virtme-ng 1.32+137.g2dcde98
- **QEMU version:** QEMU emulator version 6.2.0 (Debian 1:6.2+dfsg-2ubuntu6.24)

matttbe:
Hello, 

Do you have the same issue when using `--no-virtme-ng-init`?
[...]

PR #234, opened 2025-02-09
vng: fail if `--arch` is used without `--root`

matttbe:
Because it doesn't make sense to run on a different architecture, but with the host's root FS.

Hopefully this is not going to block weird use-cases.

Link: #233

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -596,6 +596,7 @@ ARCH_MAPPING = {
         "kernel_target": "Image",
         "kernel_image": "Image",
     },
+    # adding a new arch? Please also update get_host_arch().
 }
 
 MAKE_COMMAND = "make LOCALVERSION=-virtme"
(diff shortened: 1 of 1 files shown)


virtme_ng/run.py
```diff
@@ -912,6 +912,8 @@ def _get_virtme_arch(self, args):
         if args.arch is not None:
             if args.arch not in ARCH_MAPPING:
                 arg_fail(f"unsupported architecture: {args.arch}")
+            if args.root is None:
+                sys.stderr.write("WARNING: --arch used without --root." + "\n")
```
arighi, line 916: I'm wondering if we should just exit with an explicit error when `args.arch != get_host_arch()`, since we're going to fail anyway with obscure errors like #233

matttbe: Thank you for the review!

That makes sense. I didn't do that because there were maybe weird use-cases I didn't think about that. But if any, we can always fix those later on.

arighi (approved):

arighi:
LGTM, thanks!

Issue #237, opened 2025-02-21
Clarification about "path must be defined inside a valid overlay"


bjackman:
I'm trying to work around https://github.com/arighi/virtme-ng/issues/236 by creating subdirectories in my guest's `/mnt` and mounting stuff there via virtiofs.

With `virtme-ng 1.32+129.g1ebb7fb` this works:

```sh
❯❯  unshare -r vng --root ~/src/limmat-kernel/mkosi-rootfs/image/ --user root --rwdir=/mnt=$PWD/tools/testing/selftests/kselftest_install
mount: /etc/fstab: mount point does not exist.
       dmesg(1) may have more information after failed mount system call.
cp: cannot stat '/etc/hosts': No such file or directory
mount: /etc/hosts: mount point does not exist.
       dmesg(1) may have more information after failed mount system call.
mount: /etc/lvm: mount point does not exist.
       dmesg(1) may have more information after failed mount system call.
/usr/lib/tmpfiles.d/journal-nocow.conf:26: Failed to resolve specifier: uninitialized /etc/ detected, skipping.
All rules containing unresolvable specifiers will be skipped.
/run/virtme/guesttools/virtme-init: line 373: deallocvt: command not found
          _      _                                    
   __   _(_)_ __| |_ _ __ ___   ___       _ __   __ _ 
   \ \ / / |  __| __|  _   _ \ / _ \_____|  _ \ / _  |
    \ V /| | |  | |_| | | | | |  __/_____| | | | (_| |
     \_/ |_|_|   \__|_| |_| |_|\___|     |_| |_|\__  |
                                                |___/ 
   kernel version: 6.14.0-rc3+ x86_64
   (CTRL+d to exit)

root@virtme-ng:/# QEMU: Terminated
```

(I think those errors are harmless, just noise because my rootfs is very minimal).

But when I move to a subdirectory I get this failure:

```sh
❯❯  unshare -r vng --verbose --root ~/src/limmat-kernel/mkosi-rootfs/image/ --user root --rwdir=/mnt/kselftest_install=$PWD/tools/testing/selftests/kselftest_install
virtme: waiting for virtiofsd to start
virtme: use 'microvm' QEMU architecture
error: cannot initialize /mnt/kselftest_install inside the guest (path must be defined inside a valid overlay)
```

[...]

Issue #236, opened 2025-02-21
Create directory in the guest to mount --rodir in


bjackman:
Another case where what I want seems simple enough, but I'm not sure what the behaviour/UX should be. 

I don't like using my host rootfs for testing (it changes too much) so I normally boot a minimal guest in `virtme-ng` and then pass stuff I wanna run in there in via `--rodir` or `--rwdir`. I usually just use `/mnt` as the guest directory for that.

The minor annoyance comes when I wanna pass multiple directories in that way: the target directory needs to exist in the guest but if I've already used `/mnt` for one directory I can't use it again.

I think the best ways I can see to solve that would be:

1. Just have `virtme-init` create the target directory unconditionally.
2. Add some optional variant like `--rwdir-create`/`--rodir-create` that does this.

Option 1 seems like it might cause some surprise since that's not how `mount` or `/etc/fstab` works. If it only created 1 level of directory (`mkdir` rather than `mkdir -p`) that might help mitigate the worst sources of confusion. Option 2 seems easiest to reason about but it's kinda annoying to create new "configuration surface" like that. Any other options / other thoughts?

bjackman:
Hm I just realised I was reading the old `virtme` code not the `virtme-ng` Rust code. 

Now I look at the right code I realise that you can use `--overlay-rwdir` to solve this:

```sh
 unshare -r vng --verbose --user root --rwdir=/mnt=$PWD/tools/testing/selftests/kselftest_install  --overlay-rwdir=/mnt/kst -- "cd /mnt/kst; ./run_vmtests.sh -t mmap"
```

I dunno if that's an intentional workflow? If so I can document it.

bjackman:
(Deleted a nonsense comment. I was looking at the old `virtme` init code instead of `virtme_ng_init` when I wrote the above. So I got all confused, but I think after all I was right that this isn't supported).

Example command:

```sh
 unshare -r vng --verbose --root ~/src/limmat-kernel/mkosi-rootfs/image/ --user root --rwdir=/mnt/kst=$PWD/tools/testing/selftests/kselftest_install --overlay-rwdir=/mnt/kst -- "cd /mnt/kst/mm; ./run_vmtests.sh -t mmap
```

In this caser init fails with: `mount: /mnt/kst: mount point does not exist.`

PR #238, opened 2025-02-21
virtme: Fix guest dir presence check when --root is set

bjackman:
I'm trying to work around https://github.com/arighi/virtme-ng/issues/236 by creating subdirectories in my guest's `/mnt` and mounting stuff there via virtiofs. I found that once I replaced my `--rodir=/mnt=/foo/bar` with `--rodir=/mnt/something=/foo/bar` I get this error:

```
error: cannot initialize /mnt/kselftest_install inside the guest (path must be defined inside a valid overlay)
```

This seems to be a leaked assumption that the guest and host rootfs are the same shape. That assumption is wrong when --root is set to something other than `/`.

The fix is just to calculate the path of the directory that needs to be mounted relative to the rootfs on the host, before checking if it exists.

Fixes https://github.com/arighi/virtme-ng/issues/237

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1229,7 +1229,7 @@ def do_it() -> int:
         if not m:
             arg_fail("invalid --%s parameter %r" % (dirtype, dirarg))
         if m.group(2) is not None:
-            guestpath = m.group(1)
+            guestpath = os.path.join(args.root, os.path.relpath("/", m.group(1)))
             hostpath = m.group(2)
         else:
             hostpath = m.group(1)


bjackman:
Er, actually maybe something is wrong here:

```
❯❯  unshare -r ~/src/virtme-ng/vng --root ~/src/limmat-kernel/mkosi-rootfs/image/ --user root --rwdir=/mnt/kselftest_install/foo=$PWD/tools/testing/selftests/kselftest_install -- ls /mnt/kselftest_install

```

Furthermore, this broke something that previously worked:

```
# Without this change:
❯❯  unshare -r ~/src/virtme-ng/vng --root ~/src/limmat-kernel/mkosi-rootfs/image/ --user root --rwdir=/mnt=$PWD/tools/testing/selftests/kselftest_install -- ls /mnt                  
huge
kselftest
kselftest-list.txt
kvm
mkosi-rootfs
mm
run_kselftest.sh

# With this change

```
❯❯  unshare -r ~/src/virtme-ng/vng --root ~/src/limmat-kernel/mkosi-rootfs/image/ --user root --rwdir=/mnt=$PWD/tools/testing/selftests/kselftest_install -- ls /mnt
kselftest_install
logs
```

So maybe I need to look at this more closely.

**Comment by bjackman** [comment]
Oh, I do that because the I build the rootfs as an unprivileged user, which causes init to fail, IIUC because `mount` is `setuid`.

```
❯❯  ls -l ~/src/limmat-kernel/mkosi-rootfs/image/                                                                                            
total 64                                                              
lrwxrwxrwx  1 jackmanb primarygroup    7 Feb 20 15:40 bin -> usr/bin/
drwxr-xr-x  3 jackmanb primarygroup 4096 Feb 20 15:40 boot/
drwxr-xr-x  2 jackmanb primarygroup 4096 Dec 31 10:25 dev/
dr-x------  2 jackmanb primarygroup 4096 Feb 20 15:40 efi/
...
```

```
❯❯  vng --verbose --user root --root ~/src/limmat-kernel/mkosi-rootfs/image/
...
mount: /run: must be superuser to use mount.
       dmesg(1) may have more information after failed mount system call.
[    1.667332] mount (430) used greatest stack depth: 13120 bytes left
mkdir: cannot create directory '/run/virtme': Read-only file system
[    1.682276] mkdir (431) used greatest stack depth: 12928 bytes left
mount: /run/virtme/guesttools: mount point does not exist.
       dmesg(1) may have more information after failed mount system call.
/bin/sh: 1: exec: /run/virtme/guesttools/virtme-init: not found
...
```

Maybe `unshare -r` is a stupid solution to that problem, not sure...

I can try to get a rootfs that's owned by root and see if it changes anything

**Comment by bjackman** [comment]
OK, simple alternative fix: run `find ~/src/limmat-kernel/mkosi-rootfs/image/ -type f -exec chmod -s {} +` to strip the setuid bit from the whole rootfs. Then it boots fine without `unshare -r`. So let's test...

1. Confirm I still get https://github.com/arighi/virtme-ng/issues/237

```
~/src/linux (master|…46) <2> 
❯❯  vng --version
virtme-ng 1.32+142.g687eee6

~/src/linux (master|…46) <2> 
❯❯  vng --verbose --root ~/src/limmat-kernel/mkosi-rootfs/image/ --user root  --rwdir=/mnt/kselftest_install/foo=$PWD/tools/testing/selftests/kselftest_install
error: cannot initialize /mnt/kselftest_install/foo inside the guest (path must be defined inside a valid overlay)
```

2. Apply the commit from this PR, confirm I still repro the issue from my comment above

```
 ❯❯  ~/src/virtme-ng/vng --version
virtme-ng 1.32+143.g381e0f3

❯❯  unshare -r ~/src/virtme-ng/vng --root ~/src/limmat-kernel/mkosi-rootfs/image/ --user root --rwdir=/mnt=$PWD/tools/testing/selftests/kselftest_install -- ls /mnt

```

3. Now try without the `unshare -r`

```
~/src/linux (master|…46) <2>  13.1s 
❯❯  ~/src/virtme-ng/vng --root ~/src/limmat-kernel/mkosi-rootfs/image/ --user root --rwdir=/mnt=$PWD/tools/testing/selftests/kselftest_install -- ls /mnt

```

My brain is in Weekend Mode but I think this rules out `unshare -r` as the issue. Honestly my understanding of namespaces is pretty ropey but I don't think it should be expected to change anything about the layout of the filesystem, only the view you get of the UIDs.

arighi:
How did you generate the rootfs (mkosi I guess)? Which command exactly? I'd like to reproduce the problem using a rootfs similar to yours. Thanks.

bjackman:
The instructions to build it are here: https://github.com/bjackman/limmat-kernel?tab=readme-ov-file#building-rootfs-for-kselftest

That will only work on a distro with `apt`, if you don't have that I guess you should be able to run the build via Podman or you can download a pre-built one here: https://github.com/bjackman/limmat-kernel/releases/tag/v0.2

PR #240, opened 2025-02-26
config: create build dir if defined

matttbe:
Before this patch, if KBUILD_OUTPUT or O env vars were set, but the mentioned directory didn't exist, the script was first running ...

  $ make defconfig

... generating the .config in the build dir, but appending the extra kconfigs in the source dir.

The restriction for the build dir doesn't seem to be needed. Instead, the directory is created if it doesn't already exist.

### DIFF
--- a/virtme/commands/configkernel.py
+++ b/virtme/commands/configkernel.py
@@ -365,7 +365,12 @@ def do_it():
 
     # Check if KBUILD_OUTPUT is defined and if it's a directory
     config_dir = os.environ.get("KBUILD_OUTPUT")
-    if config_dir is not None and os.path.isdir(config_dir):
+    if config_dir is not None:
+        try:
+            os.makedirs(config_dir, exist_ok=True)
+        except Exception as exc:
+            print(f"Error: invalid directory for KBUILD_OUTPUT: {config_dir}")
+            raise SilentError() from exc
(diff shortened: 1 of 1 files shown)


virtme/commands/configkernel.py
```diff
@@ -365,7 +365,8 @@ def do_it():
 
     # Check if KBUILD_OUTPUT is defined and if it's a directory
     config_dir = os.environ.get("KBUILD_OUTPUT")
-    if config_dir is not None and os.path.isdir(config_dir):
+    if config_dir is not None:
+        os.makedirs(config_dir, exist_ok=True)
```
arighi, line 369: Maybe catch if makedirs fails and report an error, something like:
```
try:
    os.makedirs(config_dir, exist_ok=True)
except:
    print("Error: invalid directory for KBUILD_OUTPUT")
    return 1
```

matttbe: Good idea! Done!

arighi (approved):

virtme/commands/configkernel.py
matttbe: And now Pylint is happy.

arighi:
Ah pylint wasn't happy because except is too generic, but I see that you've already fixed it, good! :)

PR #242, opened 2025-03-02
init: clean up previous X11 state

arighi:
When running vng in graphic mode (-g), recent versions of Xorg may fail to start with the following error:
```
 _XSERVTransSocketUNIXCreateListener: ...SocketCreateListener() failed
 _XSERVTransMakeAllCOTSServerListeners: server already running
 (EE)
 Fatal server error:
 (EE) Cannot establish any listening sockets - Make sure an X server isn't already running(EE)
```
This can happen because the host's /tmp is exposed to the guest and Xorg might get confused finding a previous state.

Fix by removing the Xorg state files before starting the X server in the vng guest.

Fixes: da5b586 ("virtme-ng: export real /tmp to guest")

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -440,6 +440,9 @@ if [[ -n "${virtme_graphics}" ]]; then
         pre_exec_cmd=""
     fi
 
+    # Clean up any previous X11 state.
+    rm -f /tmp/.X11*/* /tmp/.X11-lock
+
     # Create a .xinitrc to start the requested graphical application.
     xinit_rc=/run/tmp/.xinitrc
     echo -e "${pre_exec_cmd}\nexec /run/tmp/.virtme-script" > ${xinit_rc}
--- a/virtme_ng_init/src/main.rs
+++ b/virtme_ng_init/src/main.rs
@@ -921,6 +921,9 @@ fn run_user_gui(tty_fd: libc::c_int) {
         // directly here so we may need extra permissions on the tty devices.
         utils::run_cmd("bash", &["-c", &format!("chown {} /dev/char/*", user)]);
 
+        // Clean up any previous X11 state.
+        utils::run_cmd("bash", &["-c", &"rm -f /tmp/.X11*/* /tmp/.X11-lock"]);
+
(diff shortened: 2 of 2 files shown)


virtme/guest/virtme-init
```diff
@@ -440,6 +440,9 @@ if [[ -n "${virtme_graphics}" ]]; then
         pre_exec_cmd=""
     fi
 
+    # Clean up any previous X11 state.
+    rm -f /tmp/.X11*/* /tmp/.X11-lock
```
matttbe, line 444: Just to be sure: bash will not complain if the files don't exist because 'rm' is used with '-f', right?

virtme_ng_init/src/main.rs
```diff
@@ -921,6 +921,9 @@ fn run_user_gui(tty_fd: libc::c_int) {
         // directly here so we may need extra permissions on the tty devices.
         utils::run_cmd("bash", &["-c", &format!("chown {} /dev/char/*", user)]);
 
+        // Clean up any previous X11 state.
+        utils::run_cmd("bash", &["-c", &"rm -f /tmp/.X11*/* /tmp/.X11-lock"]);
```
matttbe, line 925: detail: do you think there is an interest to combine this command with the previous one not to call `bash` twice in a row?

matttbe (approved):
Thank you for fixing this! Just two questions but it is probably not blocking.

virtme/guest/virtme-init
arighi: correct

virtme_ng_init/src/main.rs
arighi: We can probably save a tiny bit of time combining the two commands together, but I think it's negligible and since they're two completely different logical operations I'd prefer keep them separate for better readability.

Maybe if we need to add more commands to initialize xorg we may consider to create a `virtme/guest/virtme-xorg-script`, but at the moment I don't think it's worth it just for these two commands...

matttbe: Thank you for your reply! Fine for me to keep it like this!

PR #244, opened 2025-03-12
setup.py: Build a static virtme-ng-init

### LINKED ISSUE #243: glibc version issues with virtme-ng-init
When using  `--root`, I have been running into boot failues, e.,g:

```
unshare -r vng --verbose --user root --root ~/src/limmat-kernel/mkosi-rootfs/image/ --rwdir=/mnt=$PWD/tools/testing/selftests --debug
```

...

```
/run/virtme/guesttools/bin/virtme-ng-init: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by /run/virtme/guesttools/bin/virtme-ng-init)                                                                                                                        
[    2.578509] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000100   
```

Basically it seems like vng is assuming my host and guest have the same build environment which is not the case (even though they are both Debian. I assume because my host is Debian Testing while my guest is a stable version build - this is exactly the reason I use `--root`, to try and keep my virtme-ng test results reasonably reproducible). 

The general way to dodge this problem is to build static binaries, e.g. when I run kernel selftests I just add `LDFLAGS=-static` to the build command. I dunno how easy that is to do for virtme-ng-init?

**In the meantime, the workaround is to add `--no-virtme-ng-init`.**

arighi:
When using vng on a filesystem different than the host's filesystem (when `--root DIR` is used) we may have a different glibc version or different libraries in general, that are not compatible with the pre-built virtme-ng-init.

To fix this build a static virtme-ng-init, so that it can be used across any rootfs.

This fixes #243.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,12 @@ if build_virtme_ng_init and not os.path.exists("virtme_ng_init/Cargo.toml"):
 # Always include standard site-packages to PYTHONPATH
 os.environ['PYTHONPATH'] = sysconfig.get_paths()['purelib']
 
+# Produce static Rust binaries.
+#
+# This is required to use the same virtme-ng-init across different root
+# filesystems (when `--root DIR` is used).
+os.environ["RUSTFLAGS"] = "-C target-feature=+crt-static " + os.environ.get("RUSTFLAGS", "")
+
 
 class BuildPy(build_py):
     def run(self):


bjackman:
I checked out `983f3af1`, ran `make` from the virtme-ng source code, then this command worked:

```
unshare -r ~/src/virtme-ng/vng --verbose --user root --root ~/src/limmat-kernel/mkosi-rootfs/image/ --rwdir=/mnt=$PWD/tools/testing/selftests --debug
```

Where previously it had failed :)

setup.py
```diff
@@ -37,6 +37,12 @@
 # Always include standard site-packages to PYTHONPATH
 os.environ['PYTHONPATH'] = sysconfig.get_paths()['purelib']
 
+# Produce static Rust binaries.
+#
+# This is required to use the same virtme-ng-init across different root
+# filesystems (when `--root DIR` is used).
+os.environ["RUSTFLAGS"] = "-C target-feature=+crt-static"
```
matttbe, line 44: Just to be sure, is it fine to override the `RUSTFLAGS` variable? Or should it be extended?

arighi: Good point @matttbe, it's probably safer to extend it, I'll update the PR.

setup.py
```diff
@@ -37,6 +37,12 @@
 # Always include standard site-packages to PYTHONPATH
 os.environ['PYTHONPATH'] = sysconfig.get_paths()['purelib']
 
+# Produce static Rust binaries.
+#
+# This is required to use the same virtme-ng-init across different root
+# filesystems (when `--root DIR` is used).
+os.environ["RUSTFLAGS"] = os.environ.get("RUSTFLAGS", "") + " -C target-feature=+crt-static"
```
matttbe, line 44: I guess the user will not want to override this new default flag, so no need to put the user's flag at the end, right? :)

matttbe (approved):
Thank you for the update! The new version looks good to me, just one question, but feel free to merge ;-)

setup.py
matttbe: Good point! (Without the space before `-C` :) )

I don't know what's better. I guess it is not common to export `RUSTFLAGS` in the `.bashrc` file or similar, and use `virtme-ng` with `--root`.

arighi: Yeah, just to avoid potential breakage let's go with the safest way: add our static flag at the beginning and append the user's RUSTFLAGS at the end (pushed another updated).

PR #246, opened 2025-03-12
run: Add '--disable-monitor'

mhartmay:
Note: I'm not sure how to name the command line option.

Sometimes the user is not interested in having a QEMU monitor, therefore let's add this option. This changes for example the 'CTRL-a' behavior on the console.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -294,6 +294,11 @@ def make_parser() -> argparse.ArgumentParser:
         action="store_true",
         help="Fallback to the bash virtme-init (useful for debugging/development)",
     )
+    g.add_argument(
+        "--disable-monitor",
+        action="store_true",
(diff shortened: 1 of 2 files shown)


mhartmay:
Thanks for your feedback!

What do you think is the better command line option?
        "--disable-monitor"
or 
 "--disable-stdio-monitor"

Stdio would be more accurate, as there are still other QEMU monitors available (e.g. the one using TCP).

mhartmay:
Changes: I've worked in your feedback, improve the command line flag documentation a little, rebased and moved my insertions.

arighi:
I think I prefer `--disable-monitor`, because it's shorter and more immediate. It's true that `--disable-stdio-monitor` is more accurate, but it's just too verbose IMHO.

arighi (approved):

PR #249, opened 2025-03-14
build: support newer pip

arighi:
Building vng using pip 25.0.1 triggers a bunch of errors. Update pyproject.toml to make pip happy.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,4 +1,6 @@
 [project]
+name = "virtme-ng"
+dynamic = ["version", "license", "authors", "dependencies", "readme", "description", "classifiers", "scripts"]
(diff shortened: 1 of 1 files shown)


mhartmay:
Hi Andrea, can you please provide a short command how to try this? Thanks!

arighi:
Updated the commit adding the command that I used to trigger the error and also added more dynamic items to pyproject.toml to prevent warnings when building the package via setup.py:
```
$ VNG_PACKAGE=1 BUILD_VIRTME_NG_INIT=1 python setup.py sdist bdist_wheel
```

manufactory:
I can confirm the fix. Cf. #251

PR #253, opened 2025-03-18
virme/commands/run: find_kernel_and_mods: Fix kernel modules version …

mhartmay:
…detection on s390x

This commit fixes the following virtme-ng warning:

````
$ ./vng -v -r v6.6
...
warning: failed to retrieve kernel version from: /root/.cache/virtme-ng/v6.6/s390x/boot/vmlinuz-6.6.0-060600-generic (modules may not work) 
...
```

One functional change, besides the fix:
'kernel.version' is set to None if no kernel version could be detected.


arighi (approved):
LGTM, thanks

PR #257, opened 2025-03-19
Use of 'ruff' as linter and formatter

mhartmay:


mhartmay:
Please double check whether I missed/broke something in commit, as there are known problems in converting from %s to format-strings - see https://docs.astral.sh/ruff/rules/printf-string-formatting/#known-problems


https://github.com/arighi/virtme-ng/pull/257/commits/76434c58ea5031c71e752a57827e8d817cdd7755

PR #258, opened 2025-03-20
RFC: More s390x improvements

mhartmay:


arighi (approved):
Sorry for the late review, I was attending OSPM last week and I'm slowly catching up with emails.

Anyway, everything looks good to me, thanks!

PR #260, opened 2025-03-25
Shellcheck fixes and other bash improvements

mhartmay:


mhartmay:
I'm not sure about commit
https://github.com/arighi/virtme-ng/pull/260/commits/18b2ded2c3efbf96651efb32d4c21c66cedf3e04
as the use of `/usr/bin/env` has also some disadvantages.

And there are still some ShellCheck findings left:

```
In virtme/guest/virtme-udhcpc-script line 9:
    ip link set dev "$interface" up
                     ^--------^ SC2154 (warning): interface is referenced but not assigned.


In virtme/guest/virtme-udhcpc-script line 12:
    ip addr add "$ip/$mask" dev "$interface"
                 ^-^ SC2154 (warning): ip is referenced but not assigned.
                     ^---^ SC2154 (warning): mask is referenced but not assigned.

For more information:
  https://www.shellcheck.net/wiki/SC2154 -- interface is referenced but not a...
```

But maybe those are some "magic variables"?

mhartmay:
Note: I did not test whether `--user` is still working. Can we add (more) unit and system test?

matttbe:
> I'm not sure about commit [18b2ded](https://github.com/arighi/virtme-ng/commit/18b2ded2c3efbf96651efb32d4c21c66cedf3e04) as the use of `/usr/bin/env` has also some disadvantages.

I think all these scripts will run in the VMs where I guess there is no need to use `/usr/bin/env`. Maybe best to change that when someone will need this, and it would make it clearer why this is needed?


> And there are still some ShellCheck findings left:
> 
> ```
> In virtme/guest/virtme-udhcpc-script line 9:
>     ip link set dev "$interface" up
>                      ^--------^ SC2154 (warning): interface is referenced but not assigned.
> 
> 
> In virtme/guest/virtme-udhcpc-script line 12:
>     ip addr add "$ip/$mask" dev "$interface"
>                  ^-^ SC2154 (warning): ip is referenced but not assigned.
>                      ^---^ SC2154 (warning): mask is referenced but not assigned.
> 
> For more information:
>   https://www.shellcheck.net/wiki/SC2154 -- interface is referenced but not a...
> ```
> 
> But maybe those are some "magic variables"?

Indeed, they can be ignored in this file, see: https://udhcp.busybox.net/README.udhcpc

virtme/guest/virtme-init
```diff
@@ -57,19 +57,20 @@ fi
...
+    printf -- '%s\n' "${nr_open}" > /proc/sys/fs/nr_open
 fi
 
 # devtmpfs might be automounted; if not, mount it.
 if ! grep -q devtmpfs /proc/mounts; then
     # Ideally we'll use devtmpfs (but don't rely on /dev/null existing).
     if [[ -c /dev/null ]]; then
-	mount -n -t devtmpfs -o mode=0755,nosuid,noexec devtmpfs /dev \
-	    &>/dev/null
+	    mount -n -t devtmpfs -o mode=0755,nosuid,noexec devtmpfs /dev \
```
matttbe, line 67: The indentation looks strange here and below, no? Maybe better to use spaces here and not a mix of both.

virtme/guest/virtme-init
```diff
@@ -146,7 +147,8 @@ log "basic initialization done"
 ######## The remainder of this script is a very simple init (PID 1) ########
 
 # Does the system use systemd-tmpfiles?
-tmpfiles=`which systemd-tmpfiles 2>/dev/null` && {
+# shellcheck disable=SC2034  # Unused variable left for readability
+tmpfiles=$(command -v systemd-tmpfiles 2>/dev/null) && {
```
matttbe, line 151: is it not clearer without the variable? :)

    command -v systemd-tmpfiles 2>/dev/null && { (...) }

virtme/guest/virtme-init
```diff
@@ -236,42 +238,42 @@ real_sudoers=/etc/sudoers
...
 fi
-tmpfile="`mktemp --tmpdir=/run/tmp`"
-echo "Defaults secure_path=\"/usr/sbin:/usr/bin:/sbin:/bin\"" > $tmpfile
-echo "root ALL = (ALL) NOPASSWD: ALL" >> $tmpfile
+tmpfile="$(mktemp --tmpdir=/run/tmp)"
+echo "Defaults secure_path=\"/usr/sbin:/usr/bin:/sbin:/bin\"" > "$tmpfile"
+echo "root ALL = (ALL) NOPASSWD: ALL" >> "$tmpfile"
 if [[ -n "${virtme_user}" ]]; then
-    echo "${virtme_user} ALL = (ALL) NOPASSWD: ALL" >> $tmpfile
+    printf -- '%s\n' "${virtme_user} ALL = (ALL) NOPASSWD: ALL" >> "$tmpfile"
```
matttbe, line 246: maybe clearer with:

    printf -- '%s ALL = (ALL) NOPASSWD: ALL\n' "${virtme_user}"

while switching to `printf`, better to use it as expected :)

virtme/guest/virtme-init
```diff
@@ -236,42 +238,42 @@ real_sudoers=/etc/sudoers
...
+if grep -q -E '(^| )virtme.dhcp($| )' /proc/cmdline; then
     # Make sure all GIDs are allowed to create raw ICMP sockets (this
     # allows to run ping as regular user).
     echo "0 2147483647" > /proc/sys/net/ipv4/ping_group_range
 
     # udev is liable to rename the interface out from under us.
     for d in /sys/bus/virtio/drivers/virtio_net/virtio*/net/*; do
         virtme_net=$(basename "${d}")
-        busybox udhcpc -i "$virtme_net" -n -q -f -s "$(dirname $0)/virtme-udhcpc-script" &
+        busybox udhcpc -i "$virtme_net" -n -q -f -s "$(dirname "$0")/virtme-udhcpc-script" &
```
matttbe, line 262: Below you use `dirname -- "$0"`, but not here.

Also, I don't think `$0` will start with `-`, so maybe `--` is not needed. But best to be consistent.

virtme/guest/virtme-init
```diff
@@ -334,18 +336,18 @@ if [[ -n "${user_cmd}" ]]; then
...
         log 'starting script'
         if [[ -n "${virtme_user}" ]]; then
             chmod +x /run/tmp/.virtme-script
-            setsid su ${virtme_user} -c /run/tmp/.virtme-script </dev/virtio-ports/virtme.stdin >/dev/virtio-ports/virtme.stdout 2>/dev/virtio-ports/virtme.stderr
+            setsid su - "${virtme_user}" -c /run/tmp/.virtme-script </dev/virtio-ports/virtme.stdin >/dev/virtio-ports/virtme.stdout 2>/dev/virtio-ports/virtme.stderr
         else
             setsid bash /run/tmp/.virtme-script </dev/virtio-ports/virtme.stdin >/dev/virtio-ports/virtme.stdout 2>/dev/virtio-ports/virtme.stderr
         fi
-	ret=$?
+	    ret=$?
```
matttbe, line 351: wrong indentation?

virtme/guest/virtme-init
```diff
@@ -364,14 +366,14 @@ fi
 if [[ -n "${virtme_console}" ]]; then
     consdev=${virtme_console}
 else
-    consdev="`grep ' ... (.C' /proc/consoles  |cut -d' ' -f1`"
+    consdev="$(grep ' ... (.C' /proc/consoles  |cut -d' ' -f1)"
```
matttbe, line 370: (while at it, could you fix the spaces around `|` please?)

virtme/guest/virtme-init
```diff
@@ -391,11 +393,13 @@ if [[ ! -e "/dev/$consdev" ]]; then
 fi
 
 # Redirect current stdout/stderr to consdev
-exec 1>/dev/${consdev}
+exec 1>/dev/"${consdev}"
```
matttbe, line 397: (detail, maybe with the double quotes around the whole string: `"/dev/${consdev}"`?)

virtme/guest/virtme-init
```diff
@@ -391,11 +393,13 @@ if [[ ! -e "/dev/$consdev" ]]; then
...
 exec 2>&1
 
 # Parameters that start with virtme_ shouldn't pollute the environment
-for p in "${!virtme_@}"; do export -n "$p"; done
+for p in "${!virtme_@}"; do
+    export -n -- "${p?}";
```
matttbe, line 402: why using `${p?}`? Just in case there is no `virtme_*` variables?

virtme/guest/virtme-init
```diff
@@ -420,10 +424,15 @@ fi
 # $XDG_RUNTIME_DIR defines the base directory relative to which user-specific
 # non-essential runtime files and other file objects (such as sockets, named
 # pipes, ...) should be stored.
-export XDG_RUNTIME_DIR=/run/user/$(id -u ${virtme_user})
-mkdir -p $XDG_RUNTIME_DIR
 if [[ -n "${virtme_user}" ]]; then
-    chown ${virtme_user} $XDG_RUNTIME_DIR
+    XDG_RUNTIME_DIR=/run/user/"$(id -u -- "${virtme_user}")"
```
matttbe, line 429: (or only one line with `id -u ${virtme_user:+-- "${virtme_user}"}`)

virtme/commands/run.py
```diff
@@ -947,7 +947,7 @@ def console_client(args):
     with open(console_script_path, "w", encoding="utf-8") as file:
         print(
             (
-                "#! /bin/bash\n"
+                "#!/usr/bin/env bash\n"
```
matttbe, line 950: I'm not sure if it is really needed in the VM with only Linux, nor about the consequences. Maybe best to drop this patch for the moment?

matttbe (changes requested):
thank you for looking at this. I did a quick review, and I have a few questions.

matttbe:
> Can we add (more) unit and system test?

Yes please! They are clearly missing. It is difficult to test everything, but any tests would be helpful I'm sure. I guess there are not so many tests because the "base" is missing. If it is easier to add a new test in a few lines to verify a new feature, I'm sure they would be more tests!

virtme/guest/virtme-init
mhartmay: Actually, this was fixed automatically.

I would rather go with:
```suggestion
if command -v systemd-tmpfiles &>/dev/null; then
```

virtme/guest/virtme-init
mhartmay: Yep - I agree.

virtme/guest/virtme-init
mhartmay: Will change it.

virtme/guest/virtme-init
mhartmay: shfmt for the better :)

virtme/guest/virtme-init
mhartmay: Can change it, but doesn't make any difference.

virtme/commands/run.py
mhartmay: Sure, we can postpone the whole patch for now.

mhartmay:
> > Can we add (more) unit and system test?
> 
> Yes please! They are clearly missing. It is difficult to test everything, but any tests would be helpful I'm sure. I guess there are not so many tests because the "base" is missing. If it is easier to add a new test in a few lines to verify a new feature, I'm sure they would be more tests!

Okay, I guess we should open an issue then or something similar to track this.

virtme/guest/virtme-init
matttbe: There were 2 spaces before the pipe, I guess it was an accident.

mhartmay: Ah right, thought you would mean ` |cut` vs. ` | cut`

virtme/guest/virtme-init
mhartmay: Could be, in emacs it looked fine :) But I will probably reformat the whole code base using `shfmt`.

mhartmay:
Update: I've worked in your comments, added pre-commit hooks, an editorconfig to format shell code + a commit formatting the whole code base.

virtme/guest/virtme-init
```diff
@@ -217,9 +217,9 @@ fi
 
 # Install /proc/self/fd symlinks into /dev if not already present
 declare -r -A fdlinks=(["/dev/fd"]="/proc/self/fd"
```
matttbe, line 219: detail: would it look "better" if you move the first item to the next line?

```shell
declare -r -A fdlinks=(
    ["/dev/fd"]="/proc/self/fd"
    ["/dev/stdin"]="/proc/self/fd/0"
    ["/dev/stdout"]="/proc/self/fd/1"
    ["/dev/stderr"]="/proc/self/fd/2"
)
```

matttbe (commented):
thx for the update. Just one question.

mhartmay:
@matttbe I'm not sure about `space_redirects = true` shfmt setting, but my understanding is that most of the code used this style before.

matttbe:
> @matttbe I'm not sure about `space_redirects = true` shfmt setting, but my understanding is that most of the code used this style before.

Indeed, it looks like spaces are usually used in redirections. Exception maybe to `/dev/*` but probably fine with it. It is fine for me to add spaces, but best to check with @arighi ;-)

matttbe (approved):

arighi:
> I'm not sure about commit [18b2ded](https://github.com/arighi/virtme-ng/commit/18b2ded2c3efbf96651efb32d4c21c66cedf3e04) as the use of `/usr/bin/env` has also some disadvantages.
> 
> And there are still some ShellCheck findings left:
> 
> ```
> In virtme/guest/virtme-udhcpc-script line 9:
>     ip link set dev "$interface" up
>                      ^--------^ SC2154 (warning): interface is referenced but not assigned.
> 
> 
> In virtme/guest/virtme-udhcpc-script line 12:
>     ip addr add "$ip/$mask" dev "$interface"
>                  ^-^ SC2154 (warning): ip is referenced but not assigned.
>                      ^---^ SC2154 (warning): mask is referenced but not assigned.
> 
> For more information:
>   https://www.shellcheck.net/wiki/SC2154 -- interface is referenced but not a...
> ```
> 
> But maybe those are some "magic variables"?

Yeah, these are set by udchpc when it runs the script, not sure if there's a way to hint those warnings, but that's correct, they will be defined when the script is executed. :)

arighi:
> > @matttbe I'm not sure about `space_redirects = true` shfmt setting, but my understanding is that most of the code used this style before.
> 
> Indeed, it looks like spaces are usually used in redirections. Exception maybe to `/dev/*` but probably fine with it. It is fine for me to add spaces, but best to check with @arighi ;-)

Correct, there are few exceptions with /dev/null and similar, we never paid too much attention to those, but in general we should add spaces with redirections.

arighi (approved):

mhartmay:
> > I'm not sure about commit [18b2ded](https://github.com/arighi/virtme-ng/commit/18b2ded2c3efbf96651efb32d4c21c66cedf3e04) as the use of `/usr/bin/env` has also some disadvantages.
> > And there are still some ShellCheck findings left:
> > ```
> > In virtme/guest/virtme-udhcpc-script line 9:
> >     ip link set dev "$interface" up
> >                      ^--------^ SC2154 (warning): interface is referenced but not assigned.
> > 
> > 
> > In virtme/guest/virtme-udhcpc-script line 12:
> >     ip addr add "$ip/$mask" dev "$interface"
> >                  ^-^ SC2154 (warning): ip is referenced but not assigned.
> >                      ^---^ SC2154 (warning): mask is referenced but not assigned.
> > 
> > For more information:
> >   https://www.shellcheck.net/wiki/SC2154 -- interface is referenced but not a...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But maybe those are some "magic variables"?
> 
> Yeah, these are set by udchpc when it runs the script, not sure if there's a way to hint those warnings, but that's correct, they will be defined when the script is executed. :)

I've added declarations and that makes shellcheck happy :)

See https://github.com/arighi/virtme-ng/blob/main/virtme/guest/virtme-udhcpc-script#L8-L9

PR #261, opened 2025-03-25
Some SSH improvements: Quoting fix, pre-populated known_hosts, and vsock support for the SSH connection

mhartmay:
Some SSH improvements:
+ quoting fix
+ use of a pre-populated known_hosts file
+ use of SSH config file for the SSH connection
+ vsock support for the SSH connection (it's used by default if available)
+ a new command line flag '--ssh-tcp' is introduced to force the use of
TCP instead of vsock for the SSH connection


matttbe:
Out of curiosity, why not using the `--console` mode? Because you need some specific SSH features?

mhartmay:
SSH allows you port forwarding etc...

virtme/commands/run.py
```diff
@@ -1002,7 +1002,7 @@ def cleanup_console_script():
 
 def ssh_client(args):
     if args.remote_cmd is not None:
-        exec_escaped = args.remote_cmd.replace('"', '\\"')
+        exec_escaped = shlex.quote(args.remote_cmd)
```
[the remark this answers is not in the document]
mhartmay: Okay, will add an example to the commit description.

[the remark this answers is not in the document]
mhartmay: But have to double check...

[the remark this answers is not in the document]
matttbe: OK, I'm surprised by this, but clearer with an example.

BTW, do you still need `shlex.join(cmd)` below? Because it looks like too many quotes have been added. I would expect something like this instead, no?:

    ssh -p 2222 localhost bash -c '"echo yay; uname -n"'

Or eventually something like:

   ssh -p 2222 localhost "bash -c 'echo yay; uname -n'"

No?

mhartmay:
> I still need to review this properly, but I like the idea. This is probably going to break some workflows, so maybe we should have a separate option to enable the ssh/vsock auto-detection. For example, scp is not going to work if we transparently switch to vsock, right?

You can use `scp -o ProxyCommand=...` as we do for SSH. But another idea would be that we provide a SSH client configuration that can be easily (re)used... Similar to what Systemd[1] does.

[1] https://www.freedesktop.org/software/systemd/man/257/systemd-ssh-proxy.html

virtme/commands/run.py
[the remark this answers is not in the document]
mhartmay: Need to test... :)

mhartmay:
@matttbe and @arighi Can you please have a look at the solution [1] I've created (it's still WIP), but IMO it could be the right way.

The idea is that the `ssh_server` method already creates the SSH client configuration [2] and that configuration is then used by `ssh_client` [3]. So the whole SSH client command simplifies to    `cmd = ["ssh", "-F", f"{SSH_CONF_FILE}", VIRTME_SSH_DESTINATION_NAME] + remote_cmd`.

[1] https://github.com/arighi/virtme-ng/pull/261/commits/bd03a9978d21e273d3b1bdc6b29cf4c3dbf7d65f
[2] https://github.com/arighi/virtme-ng/pull/261/commits/bd03a9978d21e273d3b1bdc6b29cf4c3dbf7d65f#diff-5f8f0cd68abbb7e6ca08f129e724fe9f81657058af8053a69f98f45b4d026bb5R1070-R1075
[3] https://github.com/arighi/virtme-ng/pull/261/commits/bd03a9978d21e273d3b1bdc6b29cf4c3dbf7d65f#diff-5f8f0cd68abbb7e6ca08f129e724fe9f81657058af8053a69f98f45b4d026bb5R1022

virtme/commands/run.py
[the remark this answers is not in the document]
mhartmay: Yes, should also work, but that's seems not to be way how Python is doing the escaping.

arighi:
@mhartmay I still need to review all the changes, but one issue I noticed is that `vng --ssh` doesn't show the main shell session now (even if I can ssh via `vng --ssh-client`). Maybe we need to run the whole `systemd-socket-activate` in background, so that the main session can complete properly?

mhartmay:
> @mhartmay I still need to review all the changes, but one issue I noticed is that `vng --ssh` doesn't show the main shell session now (even if I can ssh via `vng --ssh-client`). Maybe we need to run the whole `systemd-socket-activate` in background, so that the main session can complete properly?

Is it now better? Stdout/stderr is still polluted, but besides that everything should be fine :) The patches themselves are now in a much better state than before.

virtme/guest/virtme-sshd-script
```diff
@@ -46,4 +46,10 @@ for key in "${CACHE_DIR}"/etc/ssh/ssh_host_*_key; do
     ARGS+=(-h "${key}")
 done
 
-/usr/sbin/sshd "${ARGS[@]}"
+if [ ! -c /dev/vsock ]; then
+    /usr/sbin/sshd "${ARGS[@]}"
+else
+    # 4294967295 == U32_MAX == -1
+    declare -r VMADDR_CID_ANY=4294967295
+    systemd-socket-activate --accept --listen="vsock:${VMADDR_CID_ANY}:22" --inetd -- /usr/sbin/sshd -i "${ARGS[@]}" &
```
arighi, line 54: This part is still blocking the init for me. Maybe we can do something like this:
```
{
    if [ ! -c /dev/vsock ]; then
        /usr/sbin/sshd "${ARGS[@]}"
    else
        # 4294967295 == U32_MAX == -1
        declare -r VMADDR_CID_ANY=4294967295
        systemd-socket-activate --accept --listen="vsock:${VMADDR_CID_ANY}:22" --inetd -- /usr/sbin/sshd -i "${ARGS[@]}"
    fi
} </dev/null &>/dev/null &

Comment (mhartmay) [reply virtme/guest/virtme-sshd-script:54]: Hmm do you use the Bash version of `init` or the Rust? I see there are differences in regard to handling stdout, but that's a separate issue and should be fixed in the init files.

Does it behave different if you use `--ssh --force-tcp`?

Comment (arighi) [reply virtme/guest/virtme-sshd-script:54]: I'm using virtme-ng-init (the Rust init). If I force the bash init, it works (`--no-virtme-ng-init`). But the stdout is still messed up (extra newline, etc.).

File: virtme/commands/run.py (hunk 1 above)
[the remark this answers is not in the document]
Comment (matttbe) [reply virtme/commands/run.py:1021]: I see, just not "nice" to look at :)

File: virtme/commands/run.py [hunk 3]
```diff
@@ -863,6 +868,10 @@ def can_use_kvm(args):
     return can_access_file("/dev/kvm")
 
 
+def can_use_vsock(args):
+    return not args.force_tcp and can_access_file("/dev/vhost-vsock")
```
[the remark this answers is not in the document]
mhartmay: We could check for that, or even implement a minimal subset of socat in Python - haven't checked it in detail, but shouldn't be tooo hard (at least I hope).

virtme/guest/virtme-sshd-script
mhartmay: The thing with the output is a difference between the Rust and the Bash implementation - I can create another commit for fixing this... (at least partly)

In Rust:
stdin is set to /dev/null and only stderr (not stdout) is logged in case of success. The buffering is also different... e.g. Stdio::piped uses buffering and so on... [1]

In Bash:
Default settings for buffering is used and no redirection of stdout/stderr/stdin [2]

[1] https://github.com/arighi/virtme-ng/blob/main/virtme_ng_init/src/utils.rs#L147-L170
[2] https://github.com/arighi/virtme-ng/blob/main/virtme/guest/virtme-init#L268

mhartmay: I've changed at least the redirection of STDIN (/dev/null) in a new commit:
https://github.com/arighi/virtme-ng/pull/261/commits/4ac06123d3768ea4e0a9552e647e548e8dc42341

For the "daemonization" of the `systemd-socket-activate ...` construct we now use `nohup` + stdout/stderr redirection.

virtme/commands/run.py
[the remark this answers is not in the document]
mhartmay: @arighi @matttbe What do you prefer? 
1. Adding a check for `socat`
2. or adding a new dependency (socat)
3. or adding an own implementation for the `socat`-work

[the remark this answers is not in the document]
matttbe: Checking `socat` is probably the safest option for the moment. Packing an alternative might be good because we only need a very small subset of features, but more work.

I don't know if a message should be printed if vosck is not supported. I guess it might break scripts? Or maybe not if it is printed to stderr?

virtme/guest/virtme-sshd-script
```diff
@@ -46,4 +46,11 @@ for key in "${CACHE_DIR}"/etc/ssh/ssh_host_*_key; do
...
 done
 
-/usr/sbin/sshd "${ARGS[@]}"
+if [ ! -c /dev/vsock ]; then
+    /usr/sbin/sshd "${ARGS[@]}"
+else
+    # 4294967295 == U32_MAX == -1
+    declare -r VMADDR_CID_ANY=4294967295
+    # TODO Use something like syslog or journal for the logging
+    nohup systemd-socket-activate --accept --listen="vsock:${VMADDR_CID_ANY}:22" --inetd -- /usr/sbin/sshd -i "${ARGS[@]}" &> /dev/null &
```
matttbe, line 55: (maybe better to use this notation because `nohup` will do more than needed and be a bit chatty) `(systemd-socket-activate &)`)

virtme/commands/run.py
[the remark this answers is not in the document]
mhartmay: For now, I would not add anything until we have figured out how to do the logging and when to show it. e.g. if `-vvv` is used or `-d` (`--debug`).

virtme/guest/virtme-sshd-script
mhartmay: I've just checked the `sshd` source code what they do to become a daemon [1], because that's what did before when we've called `/usr/sbin/sshd ...` [2]. 

1. fork
2. call `setsid`
3. redirect stdin, stdout and stderr FDs to /dev/null

So if you want, we can do the same and use `setsid` instead of `nohup`.

[1] https://github.com/openssh/openssh-portable/blob/a73890e340fbd6121251854b658a72d738b86c84/openbsd-compat/daemon.c#L52
[2] https://github.com/arighi/virtme-ng/blob/main/virtme/guest/virtme-sshd-script#L49

arighi: Looks good now, thanks @mhartmay and @matttbe !

mhartmay: Hmm need to test something, seems I broke SSH over vsock + Rust init.

mhartmay: I've added a "hack". See https://github.com/arighi/virtme-ng/pull/261/files#diff-ca70a7ba8e943bcf273078a8e5ed5314fa77bf70a93271a2da519e19efe9af5cR51.

virtme/commands/run.py
```diff
@@ -858,6 +868,18 @@ def can_use_kvm(args):
...
 
+def all_tools_available(tools: List[str]) -> bool:
+    return all(map(lambda tool: which(tool) is not None, tools))
+
+
+def can_use_ssh_over_vsock(args) -> bool:
+    return (
+        not args.force_tcp
+        and all_tools_available(["socat", "setsid", "systemd-socket-activate"])
+        and can_access_file("/dev/vhost-vsock")
```
matttbe, line 879: Similar to your other issue inside the VM: is this "file" always present, or does the corresponding kernel module need to be loaded first?

matttbe (commented):
Thank for the update. Just a few questions :)

matttbe:
> Thank for the update. Just a few questions :)

I forgot to say that I only looked at the modification, I didn't validate the modifications (I currently don't have a working environment to test :) )

virtme/guest/virtme-init
```diff
@@ -265,7 +265,7 @@ if grep -q -E '(^| )virtme.dhcp($| )' /proc/cmdline; then
 fi
 
 if grep -q -E '(^| )virtme.ssh($| )' /proc/cmdline; then
-    "$(dirname -- "$0")"/virtme-sshd-script
+    "$(dirname -- "$0")"/virtme-sshd-script < /dev/null
```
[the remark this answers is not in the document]
mhartmay: It's not needed for this change now, but it's makes the difference between the Rust and Bash init implementation smaller. See  [1] and [2] for the Rust implementation.

[1] https://github.com/arighi/virtme-ng/blob/main/virtme_ng_init/src/main.rs#L1008 
[2] https://github.com/arighi/virtme-ng/blob/main/virtme_ng_init/src/utils.rs#L147-L170

virtme/guest/virtme-sshd-script
```diff
@@ -46,4 +46,14 @@ for key in "${CACHE_DIR}"/etc/ssh/ssh_host_*_key; do
     ARGS+=(-h "${key}")
 done
 
-/usr/sbin/sshd "${ARGS[@]}"
+# Make sure vsock module is loaded and active, otherwise the '/dev/vsock' device
+# might not be available.
+modprobe vsock &>/dev/null
```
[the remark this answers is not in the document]
mhartmay: Not 100% sure (need to check the Linux source tree), but it worked this way :) So I guess it's synchronous/blocking.

virtme/guest/virtme-sshd-script
```diff
@@ -46,4 +46,14 @@ for key in "${CACHE_DIR}"/etc/ssh/ssh_host_*_key; do
...
+if [ ! -c /dev/vsock ]; then
```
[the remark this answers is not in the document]
mhartmay: The device is only available if there is a device on the QEMU side.

But actually, I'm thinking about changing this, because if the host kernel and QEMU supports vsock, but the guest kernel does not it would result in SSHD via TCP, but the client config would think it's a SSHD via vsock.

So instead of checking for the device we could simply pass an option via kernel cmdline, e.g `virtme.ssh=vsock`. If the guest kernel would not support it, it would fail and I guess that's better than a mismatch between SSH client expectation and SSHD server that would require debugging.

virtme/guest/virtme-init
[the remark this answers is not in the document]
mhartmay: We should probably check the other code places in the Bash and Rust implementation as well.

virtme/guest/virtme-sshd-script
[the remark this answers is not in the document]
matttbe: yes, certainly clearer with an env var.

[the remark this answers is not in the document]
mhartmay: A new env var or extending the existing virtme.ssh?

virtme/util.py
```diff
@@ -65,3 +68,20 @@ def find_binary_or_raise(
     if ret is None:
         raise RuntimeError(f"Could not find {names!r}")
     return ret
+
+
+def generic_vsock_addr(cid: int, port: int) -> str:
```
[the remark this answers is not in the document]
mhartmay: Sure, will move it to run.py.

virtme/commands/run.py
mhartmay:
```
Vhost-vsock is a software device so there is no probe call that causes
the driver to register its misc char device node.  This creates a
chicken and egg problem: userspace applications must open
/dev/vhost-vsock to use the driver but the file doesn't exist until the
kernel module has been loaded.

Use the devname modalias mechanism so that /dev/vhost-vsock is created
at boot.  The vhost_vsock kernel module is automatically loaded when the
first application opens /dev/host-vsock.
``` [1]

So seems like if the device is opened, the kernel module is getting loaded.

[1] https://patchwork.kernel.org/project/kvm/patch/20170509192858.17623-1-stefanha@redhat.com/

File: virtme/guest/virtme-sshd-script (hunk 4 above)
Comment (mhartmay) [reply virtme/guest/virtme-sshd-script]: I've improved this a little and should now work with kernel module and kernel builtin vsock support. [1]
It also gives an error

https://github.com/arighi/virtme-ng/pull/261/commits/0027dbfb6af208782e89b0c943f1c4c9ec123782#diff-ca70a7ba8e943bcf273078a8e5ed5314fa77bf70a93271a2da519e19efe9af5cR56

if the support in the guest kernel is missing.

**Review by matttbe — COMMENTED** [review]
Thank you for the update. The code looks good to me, but I didn't have a chance to test it, hence not "approved".

Also, a small request for next time: please always add at least the reason why the modification is needed in the commit message, even if it can be obvious to you, it might not be for others or later, e.g. for commits 4 and 5. Why is it interesting to populate the `known_host` file and set a config file.

**Comment by mhartmay** [comment]
> Thank you for the update. The code looks good to me, but I didn't have a chance to test it, hence not "approved".
> 
> Also, a small request for next time: please always add at least the reason why the modification is needed in the commit message, even if it can be obvious to you, it might not be for others or later, e.g. for commits 4 and 5. Why is it interesting to populate the `known_host` file and set a config file.

Changed.

**Comment by matttbe** [comment]
> I've found an issue, but it doesn't seem related to this PR: for some reasons scp doesn't work anymore for me, but the same happens without this PR applied, so I guess it's either something in my environment or other changes. I'll investigate more later today.

Do you use the new config file with the proxy command?

**Comment by arighi** [comment]
I run `vng --ssh` and from another session:
```
$ scp -F $HOME/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng:/etc/hosts /tmp/hosts
subsystem request failed on channel 0
scp: Connection closed
```

**Comment by arighi** [comment]
Hm... rsync works:
```
$ rsync -ve "ssh -F $HOME/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf" virtme-ng:/etc/hosts /tmp/hosts
hosts

sent 43 bytes  received 361 bytes  269,33 bytes/sec
total size is 277  speedup is 0,69
```

And scp also works if I force the legacy mode (without sftp):
```
$ scp -O -F $HOME/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng:/etc/hosts /tmp/hosts
hosts                                  100%  277   654.4KB/s   00:00
```

**Comment by arighi** [comment]
Alright, I found it, we just need to add:
```
Subsystem sftp /usr/lib/ssh/sftp-server
```

To the sshd_config that we generate in `virtme/guest/virtme-sshd-script`, we can probably fix this in a separate PR.

**Comment by arighi** [comment]
Also, the port doesn't seem to work, `vng --ssh-client` always seems to connect to the latest vng instance that was started.

**Comment by arighi** [comment]
@mhartmay I pushed 2 additional commits on top of yours (see https://github.com/arighi/virtme-ng/tree/vsock). With these extra commit everything seems to work for me. Let me know what you think.

**Comment by mhartmay** [comment]
> Also, the port doesn't seem to work, `vng --ssh-client` always seems to connect to the latest vng instance that was started.

Yep, I later realised this myself :/ I'm not sure if our current approach with "duplicating the parameters" used is the right approach. I would rather give the started vng guest an identifier, that can then be used to identify the VM. What do you think about this?

For now, your two additional commits are good enough IMO. What do you think? In the long run, I like using an identifier for the launched guest better.

**Comment by mhartmay** [comment]
@arighi  I've justed updated one commit message.

**Review by matttbe — APPROVED** [review]
thank you both for the modifications and tests!

@arighi do you want to push your modifications on top of @mhartmay 's branch, then we merge? We can also merge, then push the other modifications if that's easier :)

File: virtme/commands/run.py [hunk 10]
```diff
@@ -272,6 +280,11 @@ def make_parser() -> argparse.ArgumentParser:
     g.add_argument(
         "--force-9p", action="store_true", help="Use legacy 9p filesystem as rootfs"
     )
+    g.add_argument(
+        "--force-tcp",
```
matttbe, line 284: detail: should this be part of the `Remote Console` group below instead?

And maybe: `--force-ssh-tcp`?

virtme_ng/run.py
```diff
@@ -272,6 +272,11 @@ def make_parser():
     parser.add_argument(
         "--force-9p", action="store_true", help="Use legacy 9p filesystem as rootfs"
     )
+    parser.add_argument(
+        "--force-tcp",
```
matttbe, line 276: same here: part of the `Remote Console` group and `--force-ssh-tcp`?

matttbe (changes requested):
maybe we should move the new option and rename it? Feel free to ignore if it doesn't make sense.

mhartmay:
I've no strong opinion about this :) @arighi Do you?

virtme_ng/run.py
mhartmay: I've moved `--force-tcp` to the `Remote Console` group, but have not renamed it.

@arighi Opinions?

mhartmay:
@matttbe @arighi I've added a minimal ssh-proxy script, but it does not use `ProxyUseFdpass` [1]. I did not any performance tests... at least it works :)

[1] See https://man7.org/linux/man-pages/man5/ssh_config.5.html

arighi:
> thank you both for the modifications and tests!
> 
> @arighi do you want to push your modifications on top of @mhartmay 's branch, then we merge? We can also merge, then push the other modifications if that's easier :)

Sorry for late response, I'm slowly catching up with emails... @mhartmay are my commits still relevant? Can you apply them to your branch / PR? In the meantime I'm looking at the new changes, so I should be able to figure out the answer to this. :)

arighi:
> I've no strong opinion about this :) @arighi Do you?

I also don't have strong opinions, honestly I think `--force-tcp` is fine, it's shorter than `--force-ssh-tcp` at least.

virtme/commands/run.py
```diff
@@ -997,43 +1021,86 @@ def cleanup_console_script():
...
+    UserKnownHostsFile {VIRTME_SSH_KNOWN_HOSTS}
 
-    # Setup a port forward network interface for the guest.
-    qemuargs.extend(["-device", "{},netdev=ssh".format(arch.virtio_dev_type("net"))])
-    qemuargs.extend(["-netdev", f"user,id=ssh,hostfwd=tcp:127.0.0.1:{args.port}-:22"])
+Host {VIRTME_SSH_DESTINATION_NAME}:*
+    HostName localhost
+
+Host {VIRTME_SSH_DESTINATION_NAME}/*
+    ProxyCommand {pathlib.Path(__file__).parent.parent.resolve().joinpath("virtme-ssh-proxy")} %h
```
arighi, line 1102: We probably want to call it from `guest_tools_path` or it won't work when we install vng from package. And also add to setup.py (see `virtme-sshd-script` for example). But that's probably in the WIP part.

matttbe:
> > I've no strong opinion about this :) @arighi Do you?
> 
> I also don't have strong opinions, honestly I think `--force-tcp` is fine, it's shorter than `--force-ssh-tcp` at least.

I don't have a strong opinion on that as well, as long as it will not conflict with something else later with this generic name :)

We could also use a short version like `--ssh-tcp` if we think there is a risk.

arighi:
@mhartmay I pushed a commit on top of your branch to move `virtme-ssh-proxy` to guest tools, so packaging should be happy. Feel free to edit or squash that into one of your commits. Thanks!

mhartmay:
Btw, I did some quick benchmarking...

$ # Using f"ProxyCommand=socat STDIO SOCKET-CONNECT:{generic_vsock_addr(cid=args.port, port=22)}"
$ hyperfine 'dd if=rand.file | ssh -F /home/mhartmay/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng/2222 dd of=/dev/null'
Benchmark 1: dd if=rand.file | ssh -F /home/mhartmay/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng/2222 dd of=/dev/null
  Time (mean ± σ):      3.968 s ±  0.066 s    [User: 2.716 s, System: 2.335 s]
  Range (min … max):    3.881 s …  4.083 s    10 runs
  
$ # Using our own tool (we could rewrite it in Rust and test it in Python without using asyncio)
$ hyperfine 'dd if=rand.file | ssh -F /home/mhartmay/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng/2222 dd of=/dev/null'
Benchmark 1: dd if=rand.file | ssh -F /home/mhartmay/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng/2222 dd of=/dev/null
  Time (mean ± σ):      4.029 s ±  0.039 s    [User: 2.657 s, System: 1.718 s]
  Range (min … max):    3.979 s …  4.089 s    10 runs

mhartmay:
> > thank you both for the modifications and tests!
> > @arighi do you want to push your modifications on top of @mhartmay 's branch, then we merge? We can also merge, then push the other modifications if that's easier :)
> 
> Sorry for late response, I'm slowly catching up with emails... @mhartmay are my commits still relevant? Can you apply them to your branch / PR? In the meantime I'm looking at the new changes, so I should be able to figure out the answer to this. :)
> 
> Edit: nevermind, I just noticed that you've included my commits, so ignore this.

@arighi  I can merge your commits, but then I'm not sure if I should add your s-o-b then or leave the commits as is. Whatever you prefer :)

arighi:
Feel free to drop my s-o-b line, it's a really small change. 🙂

mhartmay:
> > > I've no strong opinion about this :) @arighi Do you?
> > 
> > 
> > I also don't have strong opinions, honestly I think `--force-tcp` is fine, it's shorter than `--force-ssh-tcp` at least.
> 
> I don't have a strong opinion on that as well, as long as it will not conflict with something else later with this generic name :)
> 
> We could also use a short version like `--ssh-tcp` if we think there is a risk.

I've changed it, but I don't like the syntax :) But I also don't have a strong opinion about this.

`~/git/virtme-ng/vng -r --ssh --ssh-tcp`

mhartmay:
> Feel free to drop my s-o-b line, it's a really small change. 🙂

I've added ProxyFdPass support: if we only support newer SSHD then we can drop the other method (using the asyncio stuff and so on).

I hope the way I merged your commits is okay with you.

arighi (approved):
This looks good to me, feel free to drop WIP from the commits and merge when you think it's ready, even if it's not perfect we can improve it in tree.

About the known_hosts, I was wondering if we should just ignore the whole thing and add something like this to virtme-ng-ssh.conf:
```
    StrictHostKeyChecking no
    LogLevel ERROR
    UserKnownHostsFile /dev/null
```

It's quite brutal and I'm not sure if there are security implications with that, but in this way we don't have to worry about using different ports / protocols.

matttbe:
> Okay, I've checked the SSH changelog and the support was added to the SSH client with version 6.5 [1]. Can we make this the minimum supported version?

@mhartmay 6.5 seems old enough. https://repology.org/project/openssh/versions

> About the known_hosts, I was wondering if we should just ignore the whole thing and add something like this to virtme-ng-ssh.conf:
> 
> ```
>     StrictHostKeyChecking no
>     LogLevel ERROR
>     UserKnownHostsFile /dev/null
> ```
> 
> It's quite brutal and I'm not sure if there are security implications with that, but in this way we don't have to worry about using different ports / protocols.

Is this `known_hosts` issue blocking? Is it not just a question of accepting it once (or ignore it in command line, or retreiving it before, or forcing vng to use a specific one, etc.)?

matttbe:
> > Okay, I've checked the SSH changelog and the support was added to the SSH client with version 6.5 [1]. Can we make this the minimum supported version?
> 
> @mhartmay : 6.5 seems old enough. https://repology.org/project/openssh/versions

I don't know if we need to support Dropbear, nor what it supports.

mhartmay:
I've figured out what the problem was... The SSH client only supports the syntax
`ssh://{VIRTME_SSH_DESTINATION_NAME}:{args.port}` to specify the port in the SSH destination name (the prefix `ssh://` is important!). For vsock that's not needed, as port 22 is the default.

arighi:
> @arighi : Is this `known_hosts` issue blocking? Is it not just a question of accepting it once (or ignore it in command line, or retreiving it before, or forcing vng to use a specific one, etc.)?

The issue was that ssh was complaining that the remote host key was changed when I was switching between tcp / vsock. But now I can't reproduce it anymore, so it was either my known_hosts messed up (I started from a clean one) or @mhartmay fixed it.

Either way, it looks good now, we just need to accept the host key once and it just works, so ignore my comment. Thanks!

arighi:
> > > Okay, I've checked the SSH changelog and the support was added to the SSH client with version 6.5 [1]. Can we make this the minimum supported version?
> > 
> > 
> > @mhartmay : 6.5 seems old enough. https://repology.org/project/openssh/versions
> 
> I don't know if we need to support Dropbear, nor what it supports.

Nah, let's ignnore dropbear. If you're on a lightweight system, using a lightweight ssh server, you get a lightweight vng as well.

arighi:
> IMO, it's now ready to be merged. Only thing I'm not sure about is if we should merge patch `ssh: move virtme-ssh-proxy to guest tools` and `virtme: Add vsock support for the SSH connection to the guest`.

Oh, I just realized that virtme-ssh-proxy is executed on the host, not the guest.... sorry about that. Then we should probably move it in ./bin/ and package it like `virtme-prep-kdir-mods`. I'll update my commit.

Regarding whether to enable vsock by default or no, usually the key questions to consider are, in order of importance:
1) does it cause any breakage?
2) does it improve boot time?
3) does it improve runtime performance?

The breakage aspect is obviously the most important. Then boot time is more relevant than runtime performance. And last, runtime performance.

About the first one, using vsock by default shouldn't cause any obvious breakage if your system is recent enough (and even if it does, we still provide an option to fallback to tcp).

About boot time I did some quick tests with the following commands:
```
$ time vng -r ~/src/linux -- uname -r
$ time vng --ssh -r ~/src/linux -- uname -r
$ time vng --ssh  --ssh-tcp -r ~/src/linux -- uname -r
```

Result:
```
avg time (no ssh): 1.30s
avg time  (vsock): 1.35s
avg time    (ssh): 1.40s 
```

Not sure why, but using vsock seems to improve boot time over tcp ssh.

About runtime performance, vsock is also faster. 

Therefore, it seems reasonable to enable it by default.

arighi:
@mhartmay updated my commit in your branch to move `virtme-ssh-proxy` to the host tools, instead of the guest tools.

If everything looks good I think we can merge this PR.

mhartmay:
> > IMO, it's now ready to be merged. Only thing I'm not sure about is if we should merge patch `ssh: move virtme-ssh-proxy to guest tools` and `virtme: Add vsock support for the SSH connection to the guest`.
> 
> Oh, I just realized that virtme-ssh-proxy is executed on the host, not the guest.... sorry about that. Then we should probably move it in ./bin/ and package it like `virtme-prep-kdir-mods`. I'll update my commit.
> 
> Regarding whether to enable vsock by default or no, usually the key questions to consider are, in order of importance:
> 
>     1. does it cause any breakage?
> 
>     2. does it improve boot time?
> 
>     3. does it improve runtime performance?
> 
> 
> The breakage aspect is obviously the most important. Then boot time is more relevant than runtime performance. And last, runtime performance.
> 
> About the first one, using vsock by default shouldn't cause any obvious breakage if your system is recent enough (and even if it does, we still provide an option to fallback to tcp).

Vsock support is checked in the host kernel (QEMU is not checked), but then we can manually fallback.

> 
> About boot time I did some quick tests with the following commands:
> 
> ```
> $ time vng -r ~/src/linux -- uname -r
> $ time vng --ssh -r ~/src/linux -- uname -r
> $ time vng --ssh  --ssh-tcp -r ~/src/linux -- uname -r
> ```
> 
> Result:
> 
> ```
> avg time (no ssh): 1.30s
> avg time  (vsock): 1.35s
> avg time    (ssh): 1.40s 
> ```
> 
> Not sure why, but using vsock seems to improve boot time over tcp ssh.

I guess the main reason is that no DHCP is used :)

> 
> About runtime performance, vsock is also faster.
> 
> Therefore, it seems reasonable to enable it by default.

+1

mhartmay:
> > @arighi : Is this `known_hosts` issue blocking? Is it not just a question of accepting it once (or ignore it in command line, or retreiving it before, or forcing vng to use a specific one, etc.)?
> 
> The issue was that ssh was complaining that the remote host key was changed when I was switching between tcp / vsock. But now I can't reproduce it anymore, so it was either my known_hosts messed up (I started from a clean one) or @mhartmay fixed it.
> 
> Either way, it looks good now, we just need to accept the host key once and it just works, so ignore my comment. Thanks!

I've fixed the issue, see https://github.com/arighi/virtme-ng/pull/261#issuecomment-2772317077.
Before my fix I used the `ssh ... virtme-ng:2222` notation and this tries to connect to hostname `virtme:2222` using port 22 and that was my localhost SSHD :)

If we use `ssh ... ssh://virtme-ng:2222` for TCP everything works as expected.

mhartmay:
@arighi @matttbe Should we use `vng` or `virtme-ng` as SSH host (alias)?

`ssh -F .... vng/2222`

vs.

`ssh -F .... virtme-ng/2222`

If we change the alias, I guess we would also have to rename the guest hostname (`hostname`). So I tend to rather leave it as is in this PR.

arighi:
> @arighi @matttbe Should we use `vng` or `virtme-ng` as SSH host (alias)?
> 
> `ssh -F .... vng/2222`
> 
> vs.
> 
> `ssh -F .... virtme-ng/2222`
> 
> If we change the alias, I guess we would also have to rename the guest hostname (`hostname`). So I tend to rather leave it as is in this PR.

I prefer virtme-ng, so that we don't have to change hostname and there's no ambiguity between command and hostname.

arighi:
@mhartmay is there a way to use scp over the vsock connection? I haven't figured out a way, the only way that works for me is to use `--ssh-tcp`. BTW we should also update the scp example in the main `README.md`.

Apart than that, is there any other things you want to change? Otherwise I'm going to merge this one. Thanks!

mhartmay:
> @mhartmay is there a way to use scp over the vsock connection? I haven't figured out a way, the only way that works for me is to use `--ssh-tcp`. BTW we should also update the scp example in the main `README.md`.

Oh, I've figured out the reason :D The separator of the hostname and CID causes the problem. We could use something different, e.g. a `#`. 

Quick test showed:
```
scp -F /home/mhartmay/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf DCO-1.1.txt 'virtme-ng#2222':~/

DCO-1.1.txt                                                                                                100% 1276     6.5MB/s   00:00
```

So we could change it to something different than `/`. Unfortunately, Systemd uses a similar syntax... so it would be easier to remember. I've checked `man hostname` [1] and it says 
```
Each element of the hostname must be from 1 to 63 characters long
and the entire hostname, including the dots, can be at most 253
characters long.  Valid characters for hostnames are [ASCII(7)](https://www.man7.org/linux/man-pages/man7/ascii.7.html) letters from a to z, the digits from 0 to 9, and the hyphen (-).
 A hostname may not start with a hyphen.
```

So if we use a `#` as separator it should not cause any problems... We should double check `rsync`, `ssh`, and `scp` (or more general: all related tools for us).

Or what looks for you better? Ideas:

```{name}#{cid}
{name}={cid}  <--- looks "similar" to : used for port (if we use the ssh://... syntax).
{name}%{cid}
{name}&{cid}
{name}_{cid}
```

> 
> Apart than that, is there any other things you want to change? Otherwise I'm going to merge this one. Thanks!

Thanks.

[1] https://www.man7.org/linux/man-pages/man7/hostname.7.html

bin/virtme-ssh-proxy
```diff
@@ -0,0 +1,55 @@
...
+    >>> cid
+    24
+    >>> hostname, cid = get_hostname_cid("virtme/ng/21")
+    >>> hostname
+    'virtme/ng'
+    >>> cid
+    21
+
+    """
+    splitted = ssh_dst.rsplit("/", 1)
```
matttbe, line 22: Should you not use `split()` instead of `rsplit`? Because the hostname should not contain any `/`, no?

(same in the example)


Or, because we should only be called with `virtme-ng/*`, we could check if the beginning is the expected one, then skip the 10 first chars to get the CID.

bin/virtme-ssh-proxy
```diff
@@ -0,0 +1,55 @@
...
+    hostname, cid = splitted[0], int(splitted[1], 10)
```
matttbe, line 23: that could cause a calltrace if the split didn't work (should always work), or if the CID is not an int. Maybe best to check if it is an int and print an readable error (or try to convert and print an error).

bin/virtme-ssh-proxy
```diff
@@ -0,0 +1,55 @@
...
+    return send_fds(socket.socket(fileno=1), [sock.fileno()])
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        prog="virtme-connect",
+    )
+    parser.add_argument("ssh_destination")
+    parser.add_argument("-p", "--port", type=int, default=22)
+    args = parser.parse_args()
```
matttbe, line 49: just to make this script lighter and simply, why not simply using `sys.argv[0]` and `sys.argv[1]` and not using `argparse`?

In the SSH config, you can simply pass the fellowing:

```
ProxyCommand "{ssh_proxy}" %h %p
```

virtme/commands/run.py
```diff
@@ -997,43 +1020,89 @@ def cleanup_console_script():
...
 
-    # Setup a port forward network interface for the guest.
-    qemuargs.extend(["-device", "{},netdev=ssh".format(arch.virtio_dev_type("net"))])
-    qemuargs.extend(["-netdev", f"user,id=ssh,hostfwd=tcp:127.0.0.1:{args.port}-:22"])
+    ssh_proxy = resources.find_script("virtme-ssh-proxy")
+
+    with open(SSH_CONF_FILE, "w", encoding="utf-8") as f:
+        f.write(f"""Host {VIRTME_SSH_DESTINATION_NAME}*
+    CheckHostIP no
+    UserKnownHostsFile {VIRTME_SSH_KNOWN_HOSTS}
```
matttbe, line 1097: In systemd, they use the following:

```
        # Disable all kinds of host identity checks, since these addresses are generally ephemeral.
        StrictHostKeyChecking no
        UserKnownHostsFile /dev/null
```

https://github.com/poettering/systemd/blob/main/src/ssh-generator/20-systemd-ssh-proxy.conf.in

So maybe fine to do the same, similar to what @arighi suggested? In this case, no need to run `ssh-keygen` above. (this can be done in another PR of course)

virtme/commands/run.py
```diff
@@ -997,43 +1020,89 @@ def cleanup_console_script():
...
+
+Host {VIRTME_SSH_DESTINATION_NAME}
+    HostName localhost
+
+Host {VIRTME_SSH_DESTINATION_NAME}/*
+    ProxyCommand "{ssh_proxy}" --port %p %h
```
matttbe, line 1103: see above, maybe easier with:

```suggestion
    ProxyCommand "{ssh_proxy}" %h %p
```

virtme/commands/run.py
```diff
@@ -1638,7 +1707,10 @@ def get_net_mac(index):
         if args.server == "console":
             console_server(args, qemu, arch, qemuargs, kernelargs)
         elif args.server == "ssh":
-            ssh_server(args, arch, qemuargs, kernelargs)
+            # Turn on vsock if available
+            ssh_server(
+                args, arch, qemuargs, kernelargs, vsock=can_use_ssh_over_vsock(args)
+            )
```
matttbe, line 1713: out of curiosity: is this style impose by the new tools?

Maybe "better" with:

```suggestion
            ssh_server(args, arch, qemuargs, kernelargs,
                       vsock=can_use_ssh_over_vsock(args))
```

virtme/commands/run.py
```diff
@@ -1051,7 +1120,7 @@ def do_it() -> int:
         if args.client == "console":
             console_client(args)
         elif args.client == "ssh":
-            ssh_client(args)
+            ssh_client(args, vsock=can_use_ssh_over_vsock(args))
```
matttbe, line 1123: (detail: why not calling `can_use_ssh_over_vsock` from `ssh_client()`? Or not having vsock as optional?)

virtme/commands/run.py
```diff
@@ -1638,7 +1707,10 @@ def get_net_mac(index):
         if args.server == "console":
             console_server(args, qemu, arch, qemuargs, kernelargs)
         elif args.server == "ssh":
-            ssh_server(args, arch, qemuargs, kernelargs)
+            # Turn on vsock if available
+            ssh_server(
+                args, arch, qemuargs, kernelargs, vsock=can_use_ssh_over_vsock(args)
```
matttbe, line 1712: (detail: same here: why not calling `can_use_ssh_over_vsock` from `ssh_server()`? Or not having vsock as optional?)

matttbe:
> I prefer `virtme-ng`, so that we don't have to change hostname and there's no ambiguity between command and hostname.

@arighi: When using it with VSOCK and a port, I don't think it has to correspond to the hostname, it is generic. Then, for me, it is fine to use `vng/*` and not change the hostname.

Note that we could also generate a conf file per hostname, e.g. with the right port in a dedicated file (one per hostname and/or port) and have `Include $PWD/ssh/*.conf` in `SSH_CONF_FILE`. Then we could do: `ssh <hostname>` and not have issues with `scp` and `rsync`. This could be done in another PR.

matttbe:
@mhartmay:

> > @mhartmay is there a way to use scp over the vsock connection? I haven't figured out a way, the only way that works for me is to use `--ssh-tcp`. BTW we should also update the scp example in the main `README.md`.
> 
> Oh, I've figured out the reason :D The separator of the hostname and CID causes the problem. We could use something different, e.g. a `#`.
> 
> Quick test showed:
> 
> ```
> scp -F /home/mhartmay/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf DCO-1.1.txt 'virtme-ng#2222':~/
> 
> DCO-1.1.txt                                                                                                100% 1276     6.5MB/s   00:00
> ```
> 
> So we could change it to something different than `/`.

Good catch!

> Unfortunately, Systemd uses a similar syntax... so it would be easier to remember.

That's strange they picked `/` in systemd then. Maybe they never noticed it was an issue for `scp` and `rsync`? Do you plan to report this to them? If they plan to support another separator, we can maybe pick the same.

> So if we use a `#` as separator it should not cause any problems... We should double check `rsync`, `ssh`, and `scp` (or more general: all related tools for us).
> 
> Or what is your favorite? Ideas:
> 
> ```
> {name}={cid}  <--- looks "similar" to : used for port (if we use the ssh://... syntax).
> {name}%{cid}
> {name}&{cid}
> {name}_{cid}
> ```

I think we should avoid `#` and `&` for the parameters. Maybe `%` or `=`?

I think we should still keep `/` as it is used by systemd, and add a second one, e.g.

```
Host {VIRTME_SSH_DESTINATION_NAME}/* {VIRTME_SSH_DESTINATION_NAME}%*
    ProxyCommand "{ssh_proxy}" %h %p
    ProxyUseFdpass yes
```

matttbe (commented):
sorry for the delay. Here are some comments on the new code I didn't have a chance to look at before.

virtme/commands/run.py
arighi: > In systemd, they use the following:
> 
> ```
>         # Disable all kinds of host identity checks, since these addresses are generally ephemeral.
>         StrictHostKeyChecking no
>         UserKnownHostsFile /dev/null
> ```
> 
> https://github.com/poettering/systemd/blob/main/src/ssh-generator/20-systemd-ssh-proxy.conf.in
> 
> So maybe fine to do the same, similar to what @arighi suggested? In this case, no need to run `ssh-keygen` above. (this can be done in another PR of course)

Yeah, we should probably do that, adding that comment as well to explain why we're doing that in case people are looking at the conf file. I think it can save some potential pain.

bin/virtme-ssh-proxy
arighi: > just to make this script lighter and simply, why not simply using `sys.argv[0]` and `sys.argv[1]` and not using `argparse`?

About this, if we want to add more options in the future argparse gives more flexibility and the overhead compared to sys.argv[] should be negligible. If performance becomes an issue, we can always rewrite this in Rust later.

arighi:
...
> I think we should avoid `#` and `&` for the parameters. Maybe `%` or `=`?
> 
> I think we should still keep `/` as it is used by systemd, and add a second one, e.g.
> 
> ```
> Host {VIRTME_SSH_DESTINATION_NAME}/* {VIRTME_SSH_DESTINATION_NAME}%*
>     ProxyCommand "{ssh_proxy}" %h %p
>     ProxyUseFdpass yes
> ```

Yes, please, let's stick with systemd's sytanx even if it's broken. If they fix/improve it and support another separator, we can follow along, but we should stay consistent with systemd. Thanks!

bin/virtme-ssh-proxy
matttbe: From what I saw, there are not many other parameters you need to pass to the proxy :)
And that's controlled by us so only using the positions should be fine, no?

arighi: Yeah, it can be considered a "backend" tool, not frontend, so positional is fine as well. I guess I can rephrase my comment as "both ways are fine". :)

virtme/commands/run.py
mhartmay: I agree, "micro optimizations" can be done later :)

virtme/commands/run.py
mhartmay: I prefer explicitness, because otherwise you always have to remember what comes first - so I would rather leave it as is. If we later find out that argparse is much slower, we still can change it in another PR (since it's used internal only).

Hope that's fine with you.

virtme/commands/run.py
mhartmay: No good reason, I've moved it to `ssh_client`. Thanks!

virtme/commands/run.py
mhartmay: No reason, I've moved it - thanks!

bin/virtme-ssh-proxy
mhartmay: See comment https://github.com/arighi/virtme-ng/pull/261#discussion_r2030216141

bin/virtme-ssh-proxy
mhartmay: When I've implemented this I haven't had checked the hostname restrictions. 

But I'm not sure if we should add these kind of checks here since it doesn't hurt to leave as is IMO. As it's implemented the script is more generic.

bin/virtme-ssh-proxy
mhartmay: Hmm I've implemented the try-and-catch-error approach.

virtme/commands/run.py
mhartmay: See new commits.

mhartmay:
Latest version supports now:
`ssh -F /home/mhartmay/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng=2222`
and
`ssh -F /home/mhartmay/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng/2222`

virtme/commands/run.py
matttbe: Yes OK, it was just to simplify stuff with `host = sys.argv[0]`, similar to what is done in systemd. But that's a simpler enough script, we can leave it like that

virtme/commands/run.py
```diff
@@ -997,43 +1022,90 @@ def cleanup_console_script():
...
+        ssh_channel_type = "tcp"
+
+    kernelargs.extend(
+        [
+            "virtme.ssh",
+            f"virtme_ssh_channel={ssh_channel_type}",
+            f"virtme_ssh_user={username}",
+        ]
+    )
+    with open(VIRTME_SSH_KNOWN_HOSTS, "w", encoding="utf-8") as f:
```
matttbe, line 1083: When `StrictHostKeyChecking no` will be added, this whole block can be removed.

matttbe (commented):
Thanks for the update, it looks good :)
but I didn't test (yet)

matttbe (approved):

arighi:
Tested both with and without `--ssh-tcp`, different ports, ssh and scp. Everything seems to work fine!

I think it'd be nice to update the scp example in the README.md. I've pushed another commit on top for that. 

@mhartmay let me know if it makes sense to you, then we can merge this. Thanks!

mhartmay:
> Thanks for the update, it looks good :) but I didn't test (yet)

Thanks!

mhartmay:
I've pushed two fixup patches, feel free to apply them and then merge the PR (after testing please :) ). I did some quick testing, but...

README.md
```diff
@@ -424,8 +424,14 @@ Examples
...
-   # In another terminal, copy result.txt from the guest to the host using scp:
-   arighi@gpd3~> scp -P 2222 localhost:~/result.txt .
+   # From another terminal on the host, print the ssh command used to
+   # connect to the vng guest (use this to get the SSH configuration file
+   # and the name of the guest):
+   arighi@host~> vng --ssh-client --dry-run
+   ssh -F /home/arighi/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng=2222
+
+   # Copy result.txt from the guest to the host using scp:
+   arighi@host~> scp -F /home/arighi/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng=2222:~/result.txt .
```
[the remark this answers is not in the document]
matttbe: Also, do you think it might be good to say that with recent systemd version, people can also use: `ssh vsock/2222`?

matttbe:
> What do you think about `,` as separator (thanks to @andreasgrapentin)? Similar to QEMU parameters...

The comma is a good idea, but I still think it would be best to report the issue to systemd maintainers. Maybe they never noticed it was an issue for scp and rsync. If they plan to support another separator, we can pick the same. Do you plan to report that to them, or do you prefer if someone else does?

We can also merge and fix that later if needed, ideally before the next virtme-ng release.

README.md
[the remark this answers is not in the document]
mhartmay: I've forced pushed the squashed commits (with the comma syntax), so we can merge the PR now. @arighi or @matttbe Feel free to improve the documentation on top of it :)

mhartmay:
> > What do you think about `,` as separator (thanks to @andreasgrapentin)? Similar to QEMU parameters...
> 
> The comma is a good idea, but I still think it would be best to report the issue to systemd maintainers. Maybe they never noticed it was an issue for scp and rsync. If they plan to support another separator, we can pick the same. Do you plan to report that to them, or do you prefer if someone else does?

I do not have a strong opinion about that, currently I'm a little blocked by other tasks. If you want, you can report and add me on CC.

> 
> We can also merge and fix that later if needed, ideally before the next virtme-ng release.

+1

README.md
```diff
@@ -424,8 +424,14 @@ Examples
...
-   # In another terminal, copy result.txt from the guest to the host using scp:
-   arighi@gpd3~> scp -P 2222 localhost:~/result.txt .
+   # From another terminal on the host, print the ssh command used to
+   # connect to the vng guest (use this to get the SSH configuration file
+   # and the name of the guest):
+   arighi@host~> vng --ssh-client --dry-run
+   ssh -F /home/arighi/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng,2222
+
+   # Copy result.txt from the guest to the host using scp:
+   arighi@host~> scp -F /home/arighi/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng,2222:~/result.txt .
```
matttbe, line 434: (mmh, my previous comment was marked as outdated and no longer visible from the "files changes" tab, re-adding it so @arighi can see it as there was a question for him)

What about suggesting this command, so users can easily copy it:

    arighi@host~> scp -F ~/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng=2222:~/result.txt .

From what I got, the path is fixed, no?

Then saying that they can get the `ssh` command by looking at the `--dry-run` command, and show the example.

Also, do you think it might be good to say that with recent systemd version, people can also use: `ssh vsock/2222`?

matttbe (commented):
I just tested this feature, it works well, thanks for the last update.

Because I see there is still a `fixup!` commit at the end, I took this opportunity to add a few optional suggestions.

README.md
arighi: Yeah initially I did the same, but then I changed it to show also that using `--dry-run` you can get the full ssh command, so we can cover two things in one command. But if you think it's more confusing than useful, we can simply add the just the scp command.

matttbe: What about writing something like this?

```
   # In another terminal, copy result.txt from the guest to the host using scp:
   arighi@host~> scp -F ~/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng,2222:~/result.txt .

   # The SSH command can be printed using this command, and easily adapted later:
   arighi@host~> vng --ssh-client --dry-run
   ssh -F /home/arighi/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng,2222
```

arighi: oh that's nice, I like it! Feel free to push a commit on top or I can change my commit with that if you prefer

matttbe:
@mhartmay @arighi I just updated the branch with the suggested modifications. Hopefully, everything is now OK :)

matttbe:
> > > What do you think about `,` as separator (thanks to @andreasgrapentin)? Similar to QEMU parameters...
> > 
> > 
> > The comma is a good idea, but I still think it would be best to report the issue to systemd maintainers. Maybe they never noticed it was an issue for scp and rsync. If they plan to support another separator, we can pick the same. Do you plan to report that to them, or do you prefer if someone else does?
> 
> I do not have a strong opinion about that, currently I'm a little blocked by other tasks. If you want, you can report and add me on CC.

I suggested to use `,` in systemd: https://github.com/systemd/systemd/pull/37035

mhartmay:
> > > > What do you think about `,` as separator (thanks to @andreasgrapentin)? Similar to QEMU parameters...
> > > 
> > > 
> > > The comma is a good idea, but I still think it would be best to report the issue to systemd maintainers. Maybe they never noticed it was an issue for scp and rsync. If they plan to support another separator, we can pick the same. Do you plan to report that to them, or do you prefer if someone else does?
> > 
> > 
> > I do not have a strong opinion about that, currently I'm a little blocked by other tasks. If you want, you can report and add me on CC.
> 
> I suggested to use `,` in systemd: [systemd/systemd#37035](https://github.com/systemd/systemd/pull/37035)

Great! Thanks!

arighi:
> I suggested to use `,` in systemd: [systemd/systemd#37035](https://github.com/systemd/systemd/pull/37035)

Thanks a lot for opening this @matttbe, we definitely want to be consistent with systemd.

arighi:
Alright, I've re-tested it and everything seems to work on my side, so let's merge this! :tada: Thanks tons for the great work on this @mhartmay and @matttbe!

matttbe:
> I suggested to use `,` in systemd: [systemd/systemd#37035](https://github.com/systemd/systemd/pull/37035)

FYI, this has been accepted. So we are aligned with systemd. With future systemd versions, we will then be able to use `scp <...> vsock,2222`

PR #266, opened 2025-04-02
configkernel: check for an existing .config before doing new one

jimc:
Currently, configkernel starts preparing a new config, and afterwards checks if one exists.  With --verbose, this is quite noisy, obscuring the "have .config, no changes made" message which follows.

Move the "wheres-config" code (both KBUILD_OUTPUT and make O=..) above the "gen-config" code to avoid the useless work and noise entirely.

Here it is, working for me:

```
jimc@frodo:~/projects/lx/wk-A$ vng -bv O=./buz
cmd: virtme-configkernel --defconfig --verbose --no-update O=./buz make[1]: Entering directory '/home/jimc/projects/lx/wk-A/buz' missing ./buz/Makefile, adding -f $src/Makefile
adding make_args: ['-f', '/home/jimc/projects/lx/wk-A/Makefile']
  GEN     Makefile
  HOSTCC  scripts/basic/fixdep
  HOSTCC  scripts/kconfig/conf.o
  HOSTCC  scripts/kconfig/confdata.o
  HOSTCC  scripts/kconfig/expr.o
  LEX     scripts/kconfig/lexer.lex.c
  YACC    scripts/kconfig/parser.tab.[ch]
  HOSTCC  scripts/kconfig/lexer.lex.o
  HOSTCC  scripts/kconfig/menu.o
  HOSTCC  scripts/kconfig/parser.tab.o
  HOSTCC  scripts/kconfig/preprocess.o
  HOSTCC  scripts/kconfig/symbol.o
  HOSTCC  scripts/kconfig/util.o
  HOSTLD  scripts/kconfig/conf
...
```

v2 pull up O=.. handling and archargs too

### DIFF
--- a/virtme/commands/configkernel.py
+++ b/virtme/commands/configkernel.py
@@ -291,6 +291,34 @@ def do_it():
     arch = architectures.get(args.arch)
     is_native = args.arch == platform.machine()
 
+    # Determine if an initial config is present
+    config = ".config"
+    makef = "Makefile"
+
+    # 1st check if make O= arg or KBUILD_OUTPUT is defined and if
+    # it's a directory; it would have the .config and Makefile
+
+    # make "O=..." takes precedence over KBUILD_OUTPUT.
+    for var in args.envs:
+        if var.startswith("O="):
+            os.environ["KBUILD_OUTPUT"] = var[2:]
+
+    config_dir = os.environ.get("KBUILD_OUTPUT")
+    if config_dir is not None:
+        try:
+            os.makedirs(config_dir, exist_ok=True)
+        except Exception as exc:
+            print(f"Error: invalid directory for KBUILD_OUTPUT: {config_dir}")
+            raise SilentError() from exc
+        config = os.path.join(config_dir, config)
+        makef = os.path.join(config_dir, makef)
+
+    if os.path.exists(config):
+        if args.no_update:
+            print(f"{config} file exists: no modifications have been done")
+            return 0
+
+    # else we make a fresh config
     custom_conf = []
     if args.custom:
         for conf_chunk in args.custom:
@@ -334,6 +362,9 @@ def do_it():
         gccname = shlex.quote(f"{cross_compile_prefix}-gcc")
         archargs.append(f"CROSS_COMPILE={gccname}")
 
+    for var in args.envs:
+        archargs.append(shlex.quote(var))
+
     maketarget: Optional[str]
 
     updatetarget = ""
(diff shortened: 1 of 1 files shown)


virtme/commands/configkernel.py
```diff
@@ -291,6 +291,38 @@ def do_it():
...
+    makef = "Makefile"
+
+    # 1st check if make O= arg or KBUILD_OUTPUT is defined and if
+    # it's a directory; it would have the .config and Makefile
+
+    # make "O=..." takes precedence over KBUILD_OUTPUT.
+    for var in args.envs:
+        if var.startswith("O="):
+            os.environ["KBUILD_OUTPUT"] = var[2:]
+        archargs.append(shlex.quote(var))
```
matttbe, line 308: I think it is better to append user-defined variables at the end, after having filed all the parameters automatically, so a user can always override what was set before.

In other words, either you move this block above:

```python
    if not is_native and shutil.which(f"{cross_compile_prefix}-gcc"):
        gccname = shlex.quote(f"{cross_compile_prefix}-gcc")
        archargs.append(f"CROSS_COMPILE={gccname}")
```

Or, maybe better, you remove this line here, and keep this loop (without the check for `O=`) where it was:

```python
    for var in args.envs:
        archargs.append(shlex.quote(var))
```

By doing that, you can keep these variables where they were

```python
    linuxname = shlex.quote(arch.linuxname)
    archargs = [f"ARCH={linuxname}"]
```

matttbe (changes requested):
thank you for the update.

virtme/commands/configkernel.py
jimc: yes, that makes sense.
its working here.
now to figure out how not to open a new PR

jimc:
I pushed -f a new rev, its reported above on this page, so I guess PR is updated ?

virtme/commands/configkernel.py
```diff
@@ -291,6 +291,34 @@ def do_it():
...
+    config_dir = os.environ.get("KBUILD_OUTPUT")
+    if config_dir is not None:
+        try:
+            os.makedirs(config_dir, exist_ok=True)
+        except Exception as exc:
+            print(f"Error: invalid directory for KBUILD_OUTPUT: {config_dir}")
+            raise SilentError() from exc
+        config = os.path.join(config_dir, config)
+        makef = os.path.join(config_dir, makef)
+
```
arighi, line 315: Can we check here if the config chunks passed via `--config FILE1 --config FILE2 ...` exist (and maybe print a clear error in that case)? They won't be used if the config already exists, but checking here would preserve the old behavior.

matttbe: Good idea! Here in `configkernel.py`, I guess you would like to check the `--custom` parameter?

So in another word, do you want to move this whole new block a few lines below, after this?

```python
    if os.path.exists(config):
        if args.no_update:
            print(f"{config} file exists: no modifications have been done")
            return 0
```

jimc:
I'm away from my computer for the weekend so if you want to do it, go for
it.

matttbe:
> I'm away from my computer for the weekend so if you want to do it, go for it.

I don't think we are in a hurry, take your time.

virtme/commands/configkernel.py
jimc: done in 2nd patch in series.

I looked briefly at vng run.py to add it there too,
but configkernel will do it anyway, 

and theres more opportunitiies
for arg-checking that probably should be added together,
by somebody who knows what checks are proper for each arg/option.

--rwdir, --overlay-rwdir, --pwd --rodir  --rw (maybe extra checks are good here)
--root

arighi:
> @arighi in case of error when used from `vng`, the user will see `virtme-configkernel`'s usage. Is it an issue?

Yeah, we should clean this up, `virtme-configkernel` is part of the old set of virtme commands, we should phase them out over time.

Not necessarily for this PR, maybe we can open another one and fix `arg_fail()` to never print the usage.

arighi (approved):

PR #268, opened 2025-04-07
version: use importlib.metadata instead of pkg_resources

matttbe:
It looks like pkg_resources is no deprecated:

```shell
  $ python3 virtme_ng/version.py
    virtme_ng/version.py:9: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  import pkg_resources
  1.33
```

According to the link, the alternative is to use `importlib.metadata`.


arighi (approved):
nit: minor typo in the commit message (no -> now).

Apart than that LGTM, thanks!

PR #269, opened 2025-04-07
resources: replace pkg_resources by importlib_resources

matttbe:
pkg_resources has been deprecated in Python v3.12.

Switch to importlib_resources instead.

We only need the path, not to read from these resources. So hopefully, this simple conversion can be applied.

Link: https://importlib-resources.readthedocs.io/en/latest/migration.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -154,9 +154,6 @@ setup(
     install_requires=[
         "argcomplete",
         "requests",
-        # `pkg_resources` is removed in python 3.12, moved to setuptools.
-        #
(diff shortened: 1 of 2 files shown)


matttbe:
Mmh, on my side, it is OK, but not with Python 3.10. Will check later if nobody knows what's wrong here :)

arighi:
@matttbe I added a commit on top of your change, it should fix the error with python 3.10. Let me know what you think (and feel free to squash the commit).

matttbe:
@arighi great, thank you for the fix! This looks good to me, and no need to squash, good to show the exception for Python 3.10!

Issue #270, opened 2025-04-07
makefile still contains git submodule update in a target


jimc:
Makefile has: 

all: init

virtme_ng_init/src/*.rs:
	git submodule update --init --recursive

this seems obsolete ?

PR #271, opened 2025-04-08
make: drop submodule update

matttbe:
virtme_ng_init is no longer in a submodule, this target is no longer needed because these files should always be there and no longer need to be generated with a command.

Closes: #270


arighi (approved):
LGTM, thanks!

PR #272, opened 2025-04-08
More fixes

mhartmay:


bin/virtme-prep-kdir-mods
```diff
@@ -47,10 +47,10 @@ fi
 # Delete existing .virtme_modes/lib/modules/0.0.0/modules.dep file at the beginning,
 # and regenerated by depmod at the end. So if we are interrupted during the
 # preparation of .virtme_mods folder, the next run command can correctly trigger the
-# prepararion work again.
+# preparation work again.
 
 if [ -f "$MODDIR/modules.dep" ]; then
```
matttbe, line 52: this `if` is then no longer needed if `rm` is used with `-f`.

> Get rid of interactive question when modules.dep is removed

In which distribution is this still the default behaviour?

virtme_ng/version.py
```diff
@@ -4,10 +4,9 @@
 """virtme-ng version"""
 
 import os
+from importlib.metadata import PackageNotFoundError, version
 from subprocess import DEVNULL, CalledProcessError, check_output
 
-from importlib.metadata import version, PackageNotFoundError
```
matttbe, line 9: We should really have the CI checking that to avoid such commit I think. (or at least to tell us a fix will be needed).

@arighi do you want to try https://pre-commit.ci? It will suggest the autofix we can squash after if needed. From what I got, you simply need to login from this page and accept the integration for this repo.

Or do you prefer the "[lite](https://pre-commit.ci/lite.html)" integration in a new GitHub action workflow? Or nothing?

matttbe (commented):
thank you for these fixes!

I have one small suggestion and question for you, and one for Andrea :)

bin/virtme-prep-kdir-mods
mhartmay: > this `if` is then no longer needed if `rm` is used with `-f`.
> 
> > Get rid of interactive question when modules.dep is removed
> 
> In which distribution is this still the default behaviour?

It was Ubuntu 24.04, but there can be multiple reasons for getting asked (not only if the file exists) :)

arighi:
@matttbe pre-commit.ci enabled, in theory it should be functional now.

matttbe:
> @matttbe pre-commit.ci enabled, in theory it should be functional now.

Great, thank you for that and for the fixes around this tool :)

I guess we can merge this PR after yours.

PR #273, opened 2025-04-08
enable different user logins through virtme ssh client

andreasgrapentin:
(This is a bit of a follow-up to PR #261)

While testing with the improved ssh connection through vsock, I intuitively expected to be able to select the login user through the `--user` parameter when passed to virtme_ng along with `--ssh-client`, but it seems like that parameter is silently ignored.

This PR proposes a slightly modified implementation that passes the `--user` parameter along with the ssh command through ssh's `-l` parameter.

This also requires some modifications to the sshd script that is started from `init` in the guest, because storing the host keys and authorized_keys file somewhere under the current users home directory causes sshd to deny logins with dubious permission errors for other users, such as root. So instead, this PR moves everything to `/opt/virtme/sshd` as a safe system location and fixes up permissions as necessary.


virtme/guest/virtme-sshd-script
```diff
@@ -14,23 +14,37 @@ fi
...
-chmod 600 "${SSH_AUTH_KEYS}" 2> /dev/null
 
-# Generate ssh host keys (if they don't exist already).
-mkdir -p "${CACHE_DIR}/etc/ssh"
-ssh-keygen -A -f "${CACHE_DIR}"
+# fixup permissions
+chown -R root:root "${SSH_DIR}"/etc/ssh
+chmod 600 "${SSH_DIR}"/etc/ssh/ssh_host_*_key
+chmod 644 "${SSH_DIR}"/etc/ssh/ssh_host_*_key.pub
+chmod 644 "${SSH_AUTH_KEYS}"
```
arighi, line 47: Can we merge this into the previous chmod (so can save a chmod exec)?

virtme/guest/virtme-sshd-script
```diff
@@ -14,23 +14,37 @@ fi
...
+SSH_HOME="$(getent passwd "${virtme_ssh_user}" | cut -d: -f6)"
 if [ ! -e "${SSH_HOME}" ]; then
     # Setup an arbitrary ssh location, just to be able to start sshd.
     SSH_HOME=/run/sshd
 fi
+SSH_CACHE="${SSH_HOME}"/.cache/virtme-ng/.ssh
+
+# use an ssh location OUTSIDE of the users home directory to enable sharing
+# of host keys and authorized keys without dubious permission errors
+SSH_DIR=/opt/virtme/sshd
```
arighi, line 28: Can we use `/run/sshd` as `SSH_DIR`, which is already created above?

andreasgrapentin: I considered using a directory below `/run`, but this causes sshd to reject the authorized_keys file:

```
Authentication refused: bad ownership or modes for directory /run
```

in the virtme guest, `/run` has mode 0777, even though on the host it has 0755. I'm not exactly sure why that is, and I didn't want to mess around with rootfs permissions, that was the reason for opting for `/opt` :)

virtme/guest/virtme-sshd-script
andreasgrapentin: sure, no problem. will fix this up with the next commit.

virtme/guest/virtme-sshd-script
andreasgrapentin: but we can move the mkdir into the `if [ ! -e "${SSH_HOME}" ]` block, so that we only run into that exec on the unlikely path.

andreasgrapentin: or get rid of it entirely, because it is covered already in 
```
mkdir -p "${SSH_CACHE}"/etc/ssh
```
later in the script.

arighi: If there's a way to not depend on `/opt` and reuse one of the existent paths it'd be nicer, `/opt` should always be present, but people are using vng in many different environments, also minimal chroots, so we need to make sure we pick a top dir that always exists and where we can write to.

andreasgrapentin: is it a problem if `/opt` does not exist? the root directory should be writable for the guest, right? so we would just create `/opt` if it's missing?

andreasgrapentin: Ah, no it actually isn't writable, that's interesting.

how about `/var/virtme/sshd`, or `/tmp/virtme/sshd`?

andreasgrapentin: I pushed a new commit that changes the mode of /run to 755 instead of the default 777 when mounting in virtme-init (both bash and rust)

that seems to be a sane enough change, and it allows us to place the keys in /run/sshd without running into dubious permission errors.

virtme/guest/virtme-sshd-script
andreasgrapentin: fixed up in 845579b

virtme/guest/virtme-sshd-script
arighi: I like this a lot more, also /run should be 755, that's the standard, so even if we break something then we'll fix that. We should be consistent with the standard. :)

arighi (approved):
LGTM, thanks!

mhartmay:
One small comment, we have to document this :) Because this can be pretty useful.

PR #275, opened 2025-04-08
ci: Fix coding style and make pre-commit hook happy

arighi:


arighi:
@mhartmay FYI, just to make sure it's not breaking your workflow.

.pre-commit-config.yaml
```diff
@@ -11,21 +11,16 @@ repos:
...
       - id: check-shebang-scripts-are-executable
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.0
+    rev: v0.11.4
     hooks:
       - id: ruff
         args: [--fix]
       - id: ruff-format
-  - repo: local
+  - repo: https://github.com/shellcheck-py/shellcheck-py
```
matttbe, line 19: Out of curiosity, was there an issue with the "official" one? Was it not to depend on docker? 

https://github.com/koalaman/shellcheck-precommit

matttbe (approved):
Thank you for having added this to prevent future unexpected modifications when using pre-commit! 

I'm a bit surprised by the modification suggested by shfmt but OK. Not sure whether we want to add the commit in theghit blame ignore file later, a detail :)

.pre-commit-config.yaml
matttbe: I see, thank you!

mhartmay: The docker stuff was also the reason why I did not use the official one... So my fallback was the local installation.

PR #278, opened 2025-04-16
README.md: Improve `virtme-ng.conf` example

mhartmay:
The syntax is incorrect for the given 'virtme-ng.conf' example in the README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -518,8 +518,8 @@ them in `~/.config/virtme-ng/virtme-ng.conf` under `default_opts` and then
 simply run `vng --build`.
 
 Example (always use an external build server called 'kathleen' and run make
-inside a build chroot called `chroot:lunar-amd64`). To do so, modify the
-`default_opts` sections in `~/.config/virtme-ng/virtme-ng.conf` as following:
+inside a build chroot called `chroot:lunar-amd64`). To do so, add the
+`default_opts` section in `~/.config/virtme-ng/virtme-ng.conf` as following:
 ```
 {
     "default_opts": {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -521,10 +521,12 @@ Example (always use an external build server called 'kathleen' and run make
 inside a build chroot called `chroot:lunar-amd64`). To do so, modify the
 `default_opts` sections in `~/.config/virtme-ng/virtme-ng.conf` as following:
```
matttbe, line 522: Here, the instruction suggests to "_modify the `default_opts` sections_". With that, the example was making sense.

But I don't think such file is automatically created. Then, what about:

```suggestion
inside a build chroot called `chroot:lunar-amd64`). To do so, add the
`default_opts` section in `~/.config/virtme-ng/virtme-ng.conf` as following:
```

(Not sure if, while at it, we should mention that this file can be installed in `/etc/`)

matttbe (changes requested):
good catch!

README.md
mhartmay: We could also ship a `virtme-ng.conf.sample` file and mention this file.

matttbe: > We could also ship a `virtme-ng.conf.sample` file and refer to this file.

Is there not already one in `cfg/virtme-ng.conf`?

mhartmay: Hmm, right. I still think it makes sense to list the full config instead of just parts of it... because a lot of people will just copy and modify it right away.

matttbe (approved):
thanks for the update!

matttbe:
> I can squash it and add you as `Co-authored-by: Matthieu Baerts <matttbe@gmail.com>`
> 
> If that's fine with you.

Thanks, but I just did that with the GitHub UI before seeing your message, it looks like it did the job :)

PR #280, opened 2025-04-17
Add support for systemd

rbmarliere:
Hello!

First of all, I'm using vng for testing local builds of [LTP](https://github.com/linux-test-project/ltp) and it's such an amazing tool! Thank you for maintaining it!

One of these days I was trying to run a test which required systemd and realized it was unfortunately not the right tool for the job... Then I saw mentions of adding support for it [here](https://lwn.net/Articles/951313/) and [here](https://github.com/arighi/virtme-ng/discussions/193). I would like to get the ball rolling on this, but I didn't know where to start. I hacked away a very ugly solution and would like to ask for guidance about how to approach this problem and refine it.

Basically, the idea is to add a `--systemd` option argument (that only makes sense currently when in use with `--root`) which would then make sure `/sbin/init` is called instead of `/run/virtme/guesttools/{virtme_init_cmd}`. After the initial systemd setup, we call virtme_init_cmd through a patched serial-getty.service (this way, we are free to keep using `--exec`). Also, we  need to make sure that virtme_init_cmd does not try to do anything that systemd already did.

Sorry for pushing this half-baked ""solution"" but I'd like some comments :)

Thanks in advance,
        rbm

EDIT: For testing, I've been using this: `vng --run . --root /path/to/debian/rootfs/ --rw --systemd`

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,36 @@ Examples
    Linux version 6.7.0-060700rc5-generic (kernel@kathleen) (x86_64-linux-gnu-gcc-13 (Ubuntu 13.2.0-7ubuntu1) 13.2.0, GNU ld (GNU Binutils for Ubuntu) 2.41) #202312102332 SMP PREEMPT_DYNAMIC Sun Dec 10 23:41:31 UTC 2023
 ```
 
+ - Run with systemd as init:
+```
+   $ sudo vng -r --systemd --exec "systemctl status | head"
+   ● virtme-ng
+      State: starting
+      Units: 392 loaded (incl. loaded aliases)
+      Jobs: 4 queued
+      Failed: 3 units
+      Since: Mon 2025-05-26 11:00:47 -03; 4s ago
+   systemd: 257.5+suse.8.gc10a66fb4d
+   Tainted: unmerged-bin
+      CGroup: /
+            ├─init.scope
+```
+
+ - Run with systemd as init in an external rootfs:
+```
+   $ vng -r --systemd --user root --root ./rootfs/sid --exec "systemctl status | head"
+   ● virtme-ng
+      State: degraded
+      Units: 273 loaded (incl. loaded aliases)
+      Jobs: 0 queued
+      Failed: 4 units
+      Since: Mon 2025-05-26 14:01:06 UTC; 2s ago
+   systemd: 257.5-2
+   Tainted: unmerged-bin
+      CGroup: /
+            ├─init.scope
+```
+
  - Run the current kernel creating a 1GB NUMA node with CPUs 0,1,3 assigned
    and a 3GB NUMA node with CPUs 2,4,5,6,7 assigned:
 ```
@@ -584,9 +614,16 @@ Troubleshooting
 ```
 
  - Snap support is still experimental and something may not work as expected
-   (keep in mind that virtme-ng will try to run snapd in a bare minimum system
-   environment without systemd), if some snaps are not running try to disable
-   apparmor, adding `--append="apparmor=0"` to the virtme-ng command line.
+   (keep in mind that, by default, virtme-ng will try to run snapd in a bare
+   minimum system environment without systemd), if some snaps are not running
+   try to disable apparmor, adding `--append="apparmor=0"` to the virtme-ng
+   command line.
+
+ - Systemd support (`--systemd`) is still experimental. If something does not
+   work for you, try masking the unit that is freezing, e.g. `--append
+   "systemd.mask=$PROBLEMATIC_UNIT"`. Be aware that you might also need `--user
+   root`, or if you're using your own `/` as ROOTFS, you may need to run vng
+   itself as root.
 
  - Running virtme-ng instances inside docker: in case of failures/issues,
    especially with stdin/stdout/stderr redirections, make sure that you have
--- a/README.md
+++ b/README.md
@@ -103,6 +103,25 @@ virtme-ng command, such as:
  $ ./vng --help
 ```
 
+Configuration
+=============
+
+* You may customize the default configuration by providing one of the
+  following, by order of preference: `$HOME/.config/virtme-ng/virtme-ng.conf`,
+  `$HOME/.virtme-ng.conf` or `/etc/virtme-ng.conf`. As a fallback for any
+  missing values, the default ones will be used.
+
+* The format of the file is JSON. Default values:
+```
+{
+    "default_opts": {},
+    "systemd": {
+        "masks": ["getty@"]
+    }
+}
+```
+
+
 Requirements
 ============
 
@@ -550,12 +569,12 @@ Default options
 Typically, if you always use virtme-ng with an external build server (e.g.,
 `vng --build --build-host REMOTE_SERVER --build-host-exec-prefix CMD`) you
 don't always want to specify these options, so instead, you can simply define
-them in `~/.config/virtme-ng/virtme-ng.conf` under `default_opts` and then
-simply run `vng --build`.
+them in your configuration file (refer to the [Configuration](#configuration)
+section) under `default_opts` and then simply run `vng --build`.
 
 Example (always use an external build server called 'kathleen' and run make
 inside a build chroot called `chroot:lunar-amd64`). To do so, add the
-`default_opts` section in `~/.config/virtme-ng/virtme-ng.conf` as following:
+`default_opts` section in your configuration file as following:
 ```
 {
     "default_opts": {
@@ -621,9 +640,10 @@ Troubleshooting
 
  - Systemd support (`--systemd`) is still experimental. If something does not
    work for you, try masking the unit that is freezing, e.g. `--append
-   "systemd.mask=$PROBLEMATIC_UNIT"`. Be aware that you might also need `--user
-   root`, or if you're using your own `/` as ROOTFS, you may need to run vng
-   itself as root.
+   "systemd.mask=$PROBLEMATIC_UNIT"` (refer to the
+   [Configuration](#configuration) section for a more permanent setup). Be
+   aware that you might also need `--user root`, or if you're using your own
+   `/` as ROOTFS, you may need to run vng itself as root.
 
  - Running virtme-ng instances inside docker: in case of failures/issues,
    especially with stdin/stdout/stderr redirections, make sure that you have
--- a/cfg/virtme-ng.conf
+++ b/cfg/virtme-ng.conf
@@ -1,4 +1,6 @@
 {
-    "default_opts" : {
+    "default_opts": {},
+    "systemd": {
+        "masks": ["getty@"]
     }
 }
--- a/cfg/virtme-ng.conf
+++ b/cfg/virtme-ng.conf
@@ -1,6 +1,0 @@
-{
-    "default_opts": {},
-    "systemd": {
-        "masks": ["getty@"]
-    }
-}
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -32,6 +32,7 @@ from virtme_ng.utils import (
     SSH_DIR,
     VIRTME_SSH_DESTINATION_NAME,
     VIRTME_SSH_HOSTNAME_CID_SEPARATORS,
+    get_conf,
 )
 
 from .. import architectures, mkinitramfs, modfinder, qemu_helpers, resources, virtmods
@@ -1311,10 +1312,11 @@ def do_it() -> int:
         kernelargs.append("luks=no")
         # disable auditd so there are no errors if the user lacks `--rw`
         kernelargs.append("audit=off")
-        # disable getty@, since we're forcing the use of serial-getty@
-        kernelargs.append("systemd.mask=getty@")
         kernelargs.extend(
             [f"console={console}" for console in arch.serial_console_args() or []],
+        )
+        kernelargs.extend(
+            [f"systemd.mask={unit}" for unit in get_conf("systemd.masks") or []]
         )
 
     if args.root == "/":
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -25,7 +25,9 @@ from time import sleep
 from typing import Any, Dict, List, NoReturn, Optional, Tuple
 
 from virtme_ng.utils import (
+    CACHE_DIR,
     DEFAULT_VIRTME_SSH_HOSTNAME_CID_SEPARATOR,
+    SERIAL_GETTY_FILE,
     SSH_CONF_FILE,
     SSH_DIR,
     VIRTME_SSH_DESTINATION_NAME,
@@ -96,6 +98,11 @@ def make_parser() -> argparse.ArgumentParser:
     g = parser.add_argument_group(title="Common guest options")
     g.add_argument(
         "--root", action="store", default="/", help="Local path to use as guest root"
+    )
+    g.add_argument(
+        "--systemd",
+        action="store_true",
+        help="Execute systemd as init (EXPERIMENTAL)",
     )
     g.add_argument(
         "--rw",
@@ -1297,28 +1304,58 @@ def do_it() -> int:
     else:
         virtme_init_cmd = "virtme-init"
 
+    if args.systemd:
+        # disable systemd-fstab-generator so boot does not freeze while waiting for disks
+        kernelargs.append("fstab=no")
+        # disable systemd-cryptsetup-generator so it doesn't wait for encrypted disks
+        kernelargs.append("luks=no")
+        # disable auditd so there are no errors if the user lacks `--rw`
+        kernelargs.append("audit=off")
+        # disable getty@, since we're forcing the use of serial-getty@
+        kernelargs.append("systemd.mask=getty@")
+        kernelargs.extend(
+            [f"console={console}" for console in arch.serial_console_args() or []],
+        )
+
     if args.root == "/":
-        initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
+        if args.systemd:
+            initcmds = [
+                "init=/bin/sh",
+                "--",
+                "-c",
+                f"SYSTEMD_UNIT_PATH={CACHE_DIR}: exec /sbin/init;",
+            ]
+        else:
+            initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
     else:
         virtfs_config = VirtFSConfig(
             path=guest_tools_path,
             mount_tag="virtme.guesttools",
         )
         export_virtfs(qemu, arch, qemuargs, virtfs_config)
-        initcmds = [
-            "init=/bin/sh",
-            "--",
-            "-c",
-            ";".join(
-                [
-                    "mount -t tmpfs run /run",
-                    "mkdir -p /run/virtme/guesttools",
-                    "/bin/mount -n -t 9p -o ro,version=9p2000.L,trans=virtio,access=any "
-                    + "virtme.guesttools /run/virtme/guesttools",
-                    f"exec /run/virtme/guesttools/{virtme_init_cmd}",
-                ]
-            ),
+        initsh = [
+            "mount -t tmpfs run /run",
+            "mkdir -p /run/virtme/guesttools",
+            "/bin/mount -n -t 9p -o ro,version=9p2000.L,trans=virtio,access=any "
+            + "virtme.guesttools /run/virtme/guesttools",
         ]
+        if args.systemd:
+            virtfs_config = VirtFSConfig(
+                path=str(CACHE_DIR),
+                mount_tag="virtme.cache",
+            )
+            export_virtfs(qemu, arch, qemuargs, virtfs_config)
+            initsh.extend(
+                [
+                    "mkdir -p /run/virtme/cache",
+                    "/bin/mount -n -t 9p -o ro,version=9p2000.L,trans=virtio,access=any "
+                    + "virtme.cache /run/virtme/cache",
+                    "SYSTEMD_UNIT_PATH=/run/virtme/cache: exec /sbin/init",
+                ]
+            )
+        else:
+            initsh.append(f"exec /run/virtme/guesttools/{virtme_init_cmd}")
+        initcmds = ["init=/bin/sh", "--", "-c", "; ".join(initsh)]
 
     # Arrange for modules to end up in the right place
     if kernel.moddir is not None:
@@ -1401,12 +1438,15 @@ def do_it() -> int:
     if args.graphics is None and not args.script_sh and not args.script_exec:
         qemuargs.extend(["-echr", "1"])
 
-        # Redirect kernel errors to stderr, creating a separate console.
-        #
-        # If we don't have access to stderr via procfs (for example when
-        # running inside a container), print a warning and implicitly
-        # suppress the kernel errors redirection.
-        if can_access_file("/proc/self/fd/2"):
+        if args.systemd:
+            # Do nothing if `--systemd` is used, since it relies on the serial console
+            pass
+        elif can_access_file("/proc/self/fd/2"):
+            # Redirect kernel errors to stderr, creating a separate console.
+            #
+            # If we don't have access to stderr via procfs (for example when
+            # running inside a container), print a warning and implicitly
+            # suppress the kernel errors redirection.
             qemuargs.extend(["-chardev", "file,path=/proc/self/fd/2,id=dmesg"])
             qemuargs.extend(["-device", arch.virtio_dev_type("serial")])
             qemuargs.extend(["-device", "virtconsole,chardev=dmesg"])
@@ -1842,6 +1882,28 @@ def do_it() -> int:
     # Load a normal kernel
     qemuargs.extend(["-kernel", kernel.kimg])
     if kernelargs:
+        if args.systemd:
+            init_environment_vars = []
+            for arg in kernelargs:
+                match = re.match(r"(virtme_.*)=(.*)", arg)
+                if not match:
+                    continue
+                init_environment_vars.append(f"{match.group(1)}={match.group(2)}")
+            os.makedirs(CACHE_DIR, exist_ok=True)
+            with open(SERIAL_GETTY_FILE, "w", encoding="utf-8") as f:
+                f.write(
+                    f"""[Service]
+StandardInput=tty
+StandardOutput=tty
+TTYPath=/dev/%I
+TTYReset=yes
+TTYVHangup=yes
+Environment={shlex.join(init_environment_vars)}"""
+                )
+                if args.root == "/":
+                    f.write(f"\nExecStart={guest_tools_path}/{virtme_init_cmd}")
+                else:
+                    f.write(f"\nExecStart=/run/virtme/guesttools/{virtme_init_cmd}")
         qemuargs.extend(["-append", " ".join(quote_karg(a) for a in kernelargs)])
     if initrdpath is not None:
         qemuargs.extend(["-initrd", initrdpath])
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -19,9 +19,12 @@ log() {
 mount -t proc -o nosuid,noexec,nodev proc /proc/
 mount -t sysfs -o nosuid,noexec,nodev sys /sys/
 
-if ! grep -q "run /run" /proc/mounts; then
-    # Mount tmpfs dirs
-    mount -t tmpfs -o mode=0755 run /run/
+if [[ $$ -eq 1 ]]; then
+    # only mount /run if systemd is not the init
+    if ! grep -q "run /run" /proc/mounts; then
+        # Mount tmpfs dirs
+        mount -t tmpfs -o mode=0755 run /run/
+    fi
 fi
 mkdir /run/tmp
 
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -19,8 +19,10 @@ log() {
 mount -t proc -o nosuid,noexec,nodev proc /proc/
 mount -t sysfs -o nosuid,noexec,nodev sys /sys/
 
-# Mount tmpfs dirs
-mount -t tmpfs -o mode=0755 run /run/
+if ! grep -q "run /run" /proc/mounts; then
+    # Mount tmpfs dirs
+    mount -t tmpfs -o mode=0755 run /run/
+fi
 mkdir /run/tmp
 
 # Setup rw filesystem overlays
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -29,7 +29,7 @@ import argcomplete
 
 from virtme.util import SilentError, get_username
 from virtme_ng.mainline import KernelDownloader
-from virtme_ng.utils import CONF_FILE, spinner_decorator
+from virtme_ng.utils import get_conf, spinner_decorator
 from virtme_ng.version import VERSION
 
 
@@ -681,31 +681,8 @@ class KernelSource:
 
     def __init__(self):
         self.virtme_param = {}
-        conf_path = self.get_conf_file_path()
-        self.default_opts = []
-        if conf_path is not None:
-            with open(conf_path, encoding="utf-8") as conf_fd:
-                conf_data = json.loads(conf_fd.read())
-                if "default_opts" in conf_data:
-                    self.default_opts = conf_data["default_opts"]
+        self.default_opts = get_conf("default_opts")
         self.cpus = str(os.cpu_count())
-
-    def get_conf_file_path(self):
-        """Return virtme-ng main configuration file path."""
-
-        # First check if there is a config file in the user's home config
-        # directory, then check for a single config file in ~/.virtme-ng.conf and
-        # finally check for /etc/virtme-ng.conf. If none of them exist, report an
-        # error and exit.
-        configs = (
-            CONF_FILE,
-            Path(Path.home(), ".virtme-ng.conf"),
-            Path("/etc", "virtme-ng.conf"),
-        )
-        for conf in configs:
-            if conf.exists():
-                return conf
-        return None
 
     def _format_cmd(self, cmd):
         return shlex.split(cmd)
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -556,6 +556,12 @@ virtme-ng is based on virtme, written by Andy Lutomirski <luto@kernel.org>.
         + "or to launch this command instead of a prompt (--client).",
     )
 
+    parser.add_argument(
+        "--systemd",
+        action="store_true",
+        help="Execute systemd as init (EXPERIMENTAL)",
+    )
+
     return parser
 
 
@@ -948,6 +954,12 @@ class KernelSource:
         else:
             self.virtme_param["root"] = ""
 
+    def _get_virtme_systemd(self, args):
+        if args.systemd:
+            self.virtme_param["systemd"] = "--systemd"
+        else:
+            self.virtme_param["systemd"] = ""
+
     def _get_virtme_rw(self, args):
         if args.rw:
             self.virtme_param["rw"] = "--rw"
@@ -1301,6 +1313,7 @@ class KernelSource:
         self._get_virtme_user(args)
         self._get_virtme_arch(args)
         self._get_virtme_root(args)
+        self._get_virtme_systemd(args)
         self._get_virtme_rw(args)
         self._get_virtme_rodir(args)
         self._get_virtme_rwdir(args)
@@ -1349,6 +1362,7 @@ class KernelSource:
             + f"{self.virtme_param['user']} "
             + f"{self.virtme_param['arch']} "
             + f"{self.virtme_param['root']} "
+            + f"{self.virtme_param['systemd']} "
             + f"{self.virtme_param['rw']} "
             + f"{self.virtme_param['rodir']} "
             + f"{self.virtme_param['rwdir']} "
--- a/virtme_ng/utils.py
+++ b/virtme_ng/utils.py
@@ -3,6 +3,7 @@
 
 """virtme-ng: configuration path."""
 
+import json
 from pathlib import Path
 
 from virtme_ng.spinner import Spinner
@@ -28,3 +29,31 @@ def spinner_decorator(message):
         return wrapper
 
     return decorator
+
+
+def get_conf_file_path():
+    """Return virtme-ng main configuration file path."""
+
+    # First check if there is a config file in the user's home config
+    # directory, then check for a single config file in ~/.virtme-ng.conf and
+    # finally check for /etc/virtme-ng.conf. If none of them exist, report an
+    # error and exit.
+    configs = (
+        CONF_FILE,
+        Path(Path.home(), ".virtme-ng.conf"),
+        Path("/etc", "virtme-ng.conf"),
+    )
+    for conf in configs:
+        if conf.exists():
+            return conf
+    return None
+
+
+def get_conf(name):
+    conf_path = get_conf_file_path()
+    if conf_path is not None:
+        with open(conf_path, encoding="utf-8") as conf_fd:
+            conf_data = json.loads(conf_fd.read())
+            if name in conf_data:
+                return conf_data[name]
+    return []
--- a/virtme_ng/utils.py
+++ b/virtme_ng/utils.py
@@ -14,9 +14,11 @@ SSH_CONF_FILE = SSH_DIR.joinpath("virtme-ng-ssh.conf")
 VIRTME_SSH_DESTINATION_NAME = "virtme-ng"
 VIRTME_SSH_HOSTNAME_CID_SEPARATORS = ("%", "/")
 DEFAULT_VIRTME_SSH_HOSTNAME_CID_SEPARATOR = VIRTME_SSH_HOSTNAME_CID_SEPARATORS[0]
-SERIAL_GETTY_FILE = Path(CACHE_DIR, "serial-getty@.service")
 CONF_PATH = Path(Path.home(), ".config", "virtme-ng")
 CONF_FILE = Path(CONF_PATH, "virtme-ng.conf")
+SERIAL_GETTY_FILE = Path(CACHE_DIR, "serial-getty@.service")
+
+# NOTE: this must stay in sync with README.md
 CONF_DEFAULT = {
     "default_opts": {},
     "systemd": {
--- a/virtme_ng/utils.py
+++ b/virtme_ng/utils.py
@@ -15,6 +15,7 @@ VIRTME_SSH_HOSTNAME_CID_SEPARATORS = ("%", "/")
 DEFAULT_VIRTME_SSH_HOSTNAME_CID_SEPARATOR = VIRTME_SSH_HOSTNAME_CID_SEPARATORS[0]
 CONF_PATH = Path(Path.home(), ".config", "virtme-ng")
 CONF_FILE = Path(CONF_PATH, "virtme-ng.conf")
+SERIAL_GETTY_FILE = Path(CACHE_DIR, "serial-getty@.service")
 
 
 def spinner_decorator(message):
--- a/virtme_ng/utils.py
+++ b/virtme_ng/utils.py
@@ -49,11 +49,24 @@ def get_conf_file_path():
     return None
 
 
-def get_conf(name):
+def get_conf(key_path):
+    """Return a configured value for a key_path, which might be nested
+
+    >>> get_conf("default_opts")
+    {}
+    """
     conf_path = get_conf_file_path()
-    if conf_path is not None:
-        with open(conf_path, encoding="utf-8") as conf_fd:
-            conf_data = json.loads(conf_fd.read())
-            if name in conf_data:
-                return conf_data[name]
-    return []
+    if conf_path is None:
+        return None
+
+    keys = key_path.split(".")
+
+    with open(conf_path, encoding="utf-8") as conf_fd:
+        conf = json.loads(conf_fd.read())
+        try:
+            for key in keys:
+                conf = conf[key]
+            return conf
+        except (KeyError, TypeError):
+            print(f"WARNING: Key {key_path} not found in {conf_path}.")
+            return []
--- a/virtme_ng/utils.py
+++ b/virtme_ng/utils.py
@@ -14,9 +14,18 @@ SSH_CONF_FILE = SSH_DIR.joinpath("virtme-ng-ssh.conf")
 VIRTME_SSH_DESTINATION_NAME = "virtme-ng"
 VIRTME_SSH_HOSTNAME_CID_SEPARATORS = ("%", "/")
 DEFAULT_VIRTME_SSH_HOSTNAME_CID_SEPARATOR = VIRTME_SSH_HOSTNAME_CID_SEPARATORS[0]
+SERIAL_GETTY_FILE = Path(CACHE_DIR, "serial-getty@.service")
 CONF_PATH = Path(Path.home(), ".config", "virtme-ng")
 CONF_FILE = Path(CONF_PATH, "virtme-ng.conf")
-SERIAL_GETTY_FILE = Path(CACHE_DIR, "serial-getty@.service")
+CONF_DEFAULT = {
+    "default_opts": {},
+    "systemd": {
+        "masks": [
+            # disable getty@, since we're forcing the use of serial-getty@
+            "getty@"
+        ]
+    },
+}
 
 
 def spinner_decorator(message):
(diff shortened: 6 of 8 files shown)


virtme/commands/run.py
```diff
@@ -1301,19 +1306,40 @@ def do_it() -> int:
...
+            "mount -t tmpfs run /run",
+            "mkdir -p /run/virtme/guesttools",
+            "/bin/mount -n -t 9p -o ro,version=9p2000.L,trans=virtio,access=any "
+            + "virtme.guesttools /run/virtme/guesttools",
+        ]
+        if args.systemd:
+            kernelargs.append("console=ttyS0")
+            # TODO: implies --rw
+            initsh.append(
+                "cp /usr/lib/systemd/system/serial-getty@.service /etc/systemd/system"
```
rbmarliere, line 1319: To avoid needing `--rw`, we could keep a patched .service file in `virtme/guest/virtme-getty-serial.service` and use something like:

```
kernelargs.append("systemd.unit=virtme-getty-serial.service")
initsh.append("SYSTEMD_UNIT_PATH=/run/virtme/guesttools: exec /sbin/init")
```

arighi:
> Basically, the idea is to add a `--systemd` option argument (that only makes sense currently when in use with `--root`)  ....

This is really neat, I like this approach.

At some point I started to hack around something, basically I was creating empty dirs and hiding the systemd state using mount --bind, so that I could start systemd in a "fake" clean system. But it was a bit tricky to figure out all the paths that systemd was using, so it never worked. However in this way, in theory, we should also be able to exec systemd's init from virtme-ng-init after the whole rootfs has been prepared.

But I like starting with the --root only approach, which seems less tricky.

Anyway, thanks for starting this, I'll take a look at your changes in details ASAP.

rbmarliere:
> > Basically, the idea is to add a `--systemd` option argument (that only makes sense currently when in use with `--root`)  ....
> 
> This is really neat, I like this approach.
> 
> At some point I started to hack around something, basically I was creating empty dirs and hiding the systemd state using mount --bind, so that I could start systemd in a "fake" clean system. But it was a bit tricky to figure out all the paths that systemd was using, so it never worked. However in this way, in theory, we should also be able to exec systemd's init from virtme-ng-init after the whole rootfs has been prepared.
> 
> But I like starting with the --root only approach, which seems less tricky.

It's a start! @marcosps told me he is a bit more interested in supporting args.root="/", but we'll get there eventually! Thank you for your feedback. I went ahead and polished a few edges already, then, so that your review time is better used.

> Anyway, thanks for starting this, I'll take a look at your changes in details ASAP.

Cheers!

marcosps:
> It's a start! @marcosps told me he is a bit more interested in supporting args.root="/", but we'll get there eventually! Thank you for your feedback. I went ahead and polished a few edges already, then, so that your review time is better used.
> 
That's true, it would be amazing to use systemd with `vng -r`, that would make my daily work much better! But for now, having this fixed to `--root` is a huge improvement!

thant

rbmarliere:
Summary of the last changes:

- Renamed the custom serial-getty service unit file to `virtme/guest/serial-getty@.service`. That's the name systemd expects and since we pass `SYSTEMD_UNIT_PATH=/run/virtme/guesttools:` to the init, it's gonna be the one used. Therefore, no need to modify anything in the root and thus no need for `--rw` anymore.
- Added runtime assembly of the `virtme/guest/virtme-getty-serial.env` file, which is consumed by the serial-getty service through `EnvironmentFile=`. The file is written with the `virtme_*` variables passed to the kernel and which the virtme-init expects to receive.
- Added `virtme_systemd=1` to the environment, used to bypass unneeded operations within virtme-init.

matttbe (commented):
That's nice you are looking at this!

I didn't really review the code, just a very quick glance :)

matttbe:
Hopefully the modifications in `virtme_ng_init/src/main.rs` will not be too difficult!

(Also, don't hesitate to include the modifications suggested by [pre-commit](https://github.com/arighi/virtme-ng/pull/280/commits/d081dea486603079256c222bf6f5c0391dcf9a8e) if you plan to amend your commit ;) )

rbmarliere:
Thanks for looking into it! I'm gonna clean things up and remove the draft status in a bit :)

rbmarliere:
@marcosps I was able to boot my openSUSE root with its kernel (`vng --run --systemd`) with the following patch on top of this PR:

```diff
diff --git a/virtme/commands/run.py b/virtme/commands/run.py
index a40815b961cc..c15d4a8b6483 100644
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1299,7 +1299,25 @@ def do_it() -> int:
         virtme_init_cmd = "virtme-init"
 
     if args.root == "/":
-        initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
+        if args.systemd:
+            kernelargs.append("console=ttyS0")
+            kernelargs.append("virtme_systemd=1")
+            initsh = [
+                "mount -t tmpfs run /run",
+                "mkdir -p /run/virtme/guesttools",
+                f"mount --bind {guest_tools_path} /run/virtme/guesttools",
+                "mount --bind /dev/null /usr/lib/systemd/system-generators/systemd-fstab-generator",
+                "mount --bind /dev/null /usr/lib/systemd/system-generators/systemd-cryptsetup-generator",
+                f"SYSTEMD_UNIT_PATH={guest_tools_path}: exec /sbin/init"
+            ]
+            initcmds = [
+                "init=/bin/sh",
+                "--",
+                "-c",
+                ";".join(initsh)
+            ]
+        else:
+            initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
     else:
         virtfs_config = VirtFSConfig(
             path=guest_tools_path,
```

marcosps:
> @marcosps I was able to boot my openSUSE root with its kernel (`vng --run --systemd`) with the following patch on top of this PR:
> 
> ```diff
> diff --git a/virtme/commands/run.py b/virtme/commands/run.py
> index a40815b961cc..c15d4a8b6483 100644
> --- a/virtme/commands/run.py
> +++ b/virtme/commands/run.py
> @@ -1299,7 +1299,25 @@ def do_it() -> int:
>          virtme_init_cmd = "virtme-init"
>  
>      if args.root == "/":
> -        initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
> +        if args.systemd:
> +            kernelargs.append("console=ttyS0")
> +            kernelargs.append("virtme_systemd=1")
> +            initsh = [
> +                "mount -t tmpfs run /run",
> +                "mkdir -p /run/virtme/guesttools",
> +                f"mount --bind {guest_tools_path} /run/virtme/guesttools",
> +                "mount --bind /dev/null /usr/lib/systemd/system-generators/systemd-fstab-generator",
> +                "mount --bind /dev/null /usr/lib/systemd/system-generators/systemd-cryptsetup-generator",
> +                f"SYSTEMD_UNIT_PATH={guest_tools_path}: exec /sbin/init"
> +            ]
> +            initcmds = [
> +                "init=/bin/sh",
> +                "--",
> +                "-c",
> +                ";".join(initsh)
> +            ]
> +        else:
> +            initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
>      else:
>          virtfs_config = VirtFSConfig(
>              path=guest_tools_path,
> ```

You're fast @rbmarliere ! Indeed, it works! It fails to start some services, like CUPS but I don't care honestly:

```
$ vng -r --systemd --no-virtme-ng-init
[FAILED] Failed to start Load/Save OS Random Seed.
mv: cannot move '/var/log/boot.msg' to '/var/log/boot.omsg': Read-only file system
/bin/sh: line 1: /var/log/boot.msg: Read-only file system
[FAILED] Failed to start Early Kernel Boot Messages.
[FAILED] Failed to start Load Audit Rules.
[DEPEND] Dependency failed for Security Audit Logging Service.
          _      _                                    
                                                |___/ 
   kernel version: 6.14.0-1-default x86_64
   (CTRL+d to exit)

[FAILED] Failed to start CUPS Scheduler.
[FAILED] Failed to start NTP client/server.
[FAILED] Failed to start CUPS Scheduler.
[FAILED] Failed to start CUPS Scheduler.
[FAILED] Failed to start System Logging Service.
[FAILED] Failed to start kgr-test container.
[FAILED] Failed to start libvirt QEMU daemon.
[FAILED] Failed to start CUPS Scheduler.
[FAILED] Failed to start User Manager for UID 1000.
[FAILED] Failed to start System Logging Service.
-bash: cannot set terminal process group (709): Inappropriate ioctl for device
-bash: no job control in this shell
[FAILED] Failed to start System Logging Service.
[FAILED] Failed to start CUPS Scheduler.
[FAILED] Failed to start libvirt QEMU daemon.
[FAILED] Failed to start System Logging Service.
mpdesouza@daedalus:~> [FAILED] Failed to start CUPS Scheduler.
[FAILED] Failed to listen on CUPS Scheduler.
[DEPEND] Dependency failed for CUPS Scheduler.
[FAILED] Failed to start System Logging Service.
[FAILED] Failed to start System Logging Service.
[FAILED] Failed to start libvirt QEMU daemon.
[FAILED] Failed to start libvirt QEMU daemon.
[FAILED] Failed to start libvirt QEMU daemon.
[FAILED] Failed to start libvirt QEMU daemon.
[FAILED] Failed to start Postfix Mail Transport Agent.
```

For `OS seed`, I can see this error:

```
mpdesouza@daedalus:~> /usr/lib/systemd/systemd-random-seed load
Failed to open /var/lib/systemd/random-seed for writing: Permission denied
Failed to open /var/lib/systemd/random-seed for reading: Permission denied
```

But we can square such problems after this PR is merged. I believe that having the --root working is the primary target, so later on we can work on it for supporting `systemd` with the host rootfs as well.

marcosps:
> Thanks for testing! In my case, I saw some failures surrounding NVIDIA... The main problem, I think, is that systemd is trying to write into a RO filesystem.

Hum... maybe yes, but we had issues in the past when trying to write to some directories that are owned by root, so when the overlayfs created the write layer we can't write to it either way...

arighi:
Hi @rbmarliere I did some tests with this, looks great!

There are still some issues to be addressed:
1) we need to support virtme-ng-init as well (I can take a look if I find some time)
2) autologin isn't working for me, I've tried on Ubuntu and Arch (Ubuntu gets stuck doing some snapd seeding crap...) I guess supporting this in a distro-agnostic way is going to be really challenging
3) we should't create `virtme-serial-getty.env` in `guest_tool_dir`, better to move it in `$HOME/.cache/virtme-ng/.systemd/` or something similar

However, if we clarify that this option is experimental in the help I have no problem to merge this and we can keep improving it in tree. Maybe I'll cut a new version first, before merging, just in case. :)

rbmarliere:
Hey hey @arighi ! Thanks for testing.

> * we need to support virtme-ng-init as well (I can take a look if I find some time)

Understood, I'll look into it. It's been a while I don't do any Rust, so it will be good :)

> * autologin isn't working for me, I've tried on Ubuntu and Arch (Ubuntu gets stuck doing some snapd seeding crap...) I guess supporting this in a distro-agnostic way is going to be really challenging

Indeed, I only tried Debian and SUSE... Best we can do is have a list of things to disable and hope the user don't have/need anything else. I will improve the code around that.

> * we should't create `virtme-serial-getty.env` in `guest_tool_dir`, better to move it in `$HOME/.cache/virtme-ng/.systemd/` or something similar

This is a little tricky because we need at least `{guest_tool_dir}/virtme/guest/serial-getty@.service` and that's [where it's needed](https://github.com/arighi/virtme-ng/blob/ce6672d1e33313f0f665188e505fde2ca19d723d/virtme/guest/serial-getty@.service#L2).

> However, if we clarify that this option is experimental in the help I have no problem to merge this and we can keep improving it in tree. Maybe I'll cut a new version first, before merging, just in case. :)

Thanks, that sound good, let me just go through this final refinement then :)

rbmarliere:
FYI I'm exploring a simplification of the whole approach by calling systemd from virtme-init instead. It would be cleaner that way and easier to integrate with virtme-ng-init. Currently it works, but the tricky part is when the user pass `--exec`. It's available here: https://github.com/rbmarliere/virtme-ng/tree/systemd2

I'll remove the draft status once I get some more time to settle on the idea. Suggestions welcome! :)

septatrix:
> First of all, I'm using vng for testing local builds of [LTP](https://github.com/linux-test-project/ltp) and it's such an amazing tool! Thank you for maintaining it!

Depending on your exact use-case you might also want to take a look at [mkosi](https://github.com/systemd/mkosi) and [mkosi-kernel](https://github.com/DaanDeMeyer/mkosi-kernel/tree/main/mkosi.profiles/ltp) (which provides some mkosi "presets" for e.g. LTP)

virtme/commands/run.py
```diff
@@ -1294,26 +1299,44 @@ def do_it() -> int:
...
         )
         export_virtfs(qemu, arch, qemuargs, virtfs_config)
+        initsh = [
+            "mount -t tmpfs run /run",
+            "mkdir -p /run/virtme/guesttools",
+            "/bin/mount -n -t 9p -o ro,version=9p2000.L,trans=virtio,access=any "
+            + "virtme.guesttools /run/virtme/guesttools",
+        ]
+        if args.systemd:
+            kernelargs.append("console=ttyS0")
```
mhartmay, line 1330:
```suggestion
            kernelargs.append(f"console={arch.serial_console_args()}")
```

virtme/commands/run.py
```diff
@@ -1294,26 +1299,44 @@ def do_it() -> int:
         virtme_init_cmd = "virtme-init"
 
     if args.root == "/":
-        initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
+        if args.systemd:
+            kernelargs.append("console=ttyS0")
```
mhartmay, line 1303:
```suggestion
            kernelargs.append(f"console={arch.serial_console_args()}")
```

rbmarliere:
> There are still some issues to be addressed:
> 
>     1. we need to support virtme-ng-init as well (I can take a look if I find some time)
Done!
>     2. autologin isn't working for me, I've tried on Ubuntu and Arch (Ubuntu gets stuck doing some snapd seeding crap...) I guess supporting this in a distro-agnostic way is going to be really challenging
My idea as a follow-up is to have  [these](https://github.com/arighi/virtme-ng/pull/280/files#diff-5f8f0cd68abbb7e6ca08f129e724fe9f81657058af8053a69f98f45b4d026bb5R1309) configurable in `~/.config/virtme-ng/virtme-ng.config`. So that the user can disable more things.
>     3. we should't create `virtme-serial-getty.env` in `guest_tool_dir`, better to move it in `$HOME/.cache/virtme-ng/.systemd/` or something similar
Done! Now the unit file is generated at runtime and written to `~/.cache/virtme-ng/serial-getty@.service`. But if the user passes `--root` then it will also export and mount the cache directory.

virtme/commands/run.py
```diff
@@ -1293,28 +1300,55 @@ def do_it() -> int:
...
+        kernelargs.append(f"console={arch.serial_console_args()[0]}")
+        initsh.extend(
+            [
+                "mount --bind /dev/null /usr/lib/systemd/system-generators/systemd-fstab-generator",
+                "mount --bind /dev/null /usr/lib/systemd/system-generators/systemd-cryptsetup-generator",
+                "mount --bind /dev/null /usr/lib/systemd/system/systemd-journald.service",
+            ]
+        )
+
     if args.root == "/":
```
mhartmay, line 1315: I haven't checked in detail, but to me it seems like the only difference if args.root is set to "/" that the guest_tools_path is either `/run/virtme/guesttools` or `f"{guest_tools_path}"`.

Couldn't we use this fact and make the code simpler? (at multiple places)

marcosps (approved):
For my usecase, running

`sudo vng -r --systemd --user root`

Makes it work with only klog.service failing on openSUSE, which is not a big deal. It currently requires sudo since it will basically grab all systemd services from host in this case.

One minor nit about it is that for now I'm using virtme-ng from master and install it using `pipx`, so it doesn't rewrite the /etc/virtme-ng.conf, so I had to copy the new file over it.

Also, the users config file maybe shuold be placed on ~/.config/virtme-ng to no pollute the user's HOME with random config files.

But this is working! Thanks a lot @rbmarliere for working on it! It will unlock using virtme at SUSE in our internal tooling for testing livepatches!

cfg/virtme-ng.conf
```diff
@@ -1,4 +1,11 @@
 {
-    "default_opts" : {
+    "default_opts": {},
+    "systemd": {
+        "masks": [
+            "/usr/lib/systemd/system-generators/systemd-fstab-generator",
+            "/usr/lib/systemd/system-generators/systemd-cryptsetup-generator",
+            "/usr/lib/systemd/system/auditd.service",
+            "/usr/lib/systemd/system/getty@.service"
```
mhartmay, line 8: Can you please add comments for what the masks are used?

For example, it's not to obvious why `serial-getty@.service` is not masked.

virtme/commands/run.py
mhartmay: Yep, it is out-of-scope.

virtme/commands/run.py
```diff
@@ -1293,28 +1301,51 @@ def do_it() -> int:
     else:
         virtme_init_cmd = "virtme-init"
 
+    initsh = []
+    if args.systemd:
+        kernelargs.append("virtme_systemd=1")
+        kernelargs.append(f"console={arch.serial_console_args()[0]}")
+        initsh.extend(
+            ["mount --bind /dev/null " + unit for unit in get_conf("systemd.masks")]
```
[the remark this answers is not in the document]
rbmarliere: IIRC from my initial testing, using cmdline arguments only work when you use init=/sbin/init directly, but since we need a wrapper /bin/sh in some cases then that's the approach that works every time.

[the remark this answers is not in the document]
mhartmay: Okay, the arguments are interpreted by a systemd generator [1]

[1] https://www.freedesktop.org/software/systemd/man/latest/systemd-debug-generator.html#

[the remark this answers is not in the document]
mhartmay: Out of curiosity, I' just checked the source code of systemd [1] and they create symlinks to /dev/null. What was the motivation to use `mount --bind` instead? :)

[1] https://github.com/systemd/systemd/blob/f9d76a8d1596927b52eded4ce7e8bbf512aff551/src/debug-generator/debug-generator.c#L222-L223

cfg/virtme-ng.conf
rbmarliere: systemd-fstab-generator: it's responsible to mount all the devices from fstab, so if the user users args.root=/ then it will freeze the boot while waiting for the devices.
systemd-cryptsetup-generator: same thing, but this time it disables mounting also the encrypted drives he may have.
auditd: it was bringing a lot of failures if not used with --rw
getty@: the code in this PR effectively creates a customized serial-getty@.service file and force its usage by means of setting SYSTEMD_UNIT_PATH, so masking all the regular getty@ makes sense (so it removes any error that might flood the stream, e.g. if the user has autologin set for getty@tty1)

virtme/commands/run.py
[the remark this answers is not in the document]
septatrix: > IIRC from my initial testing, using cmdline arguments only work when you use init=/sbin/init directly, but since we need a wrapper /bin/sh in some cases then that's the approach that works every time.

I am not sure if systemd reads the cmdline from procfs but PID 1 is also given the cmdline as its own argument, so using something like `exec systemd $(cat /proc/cmdline)` could work (not sure if `cat` works as wanted here though, cannot remember how the cmdline file is formatted)

cfg/virtme-ng.conf
septatrix: Instead of masking the audit service it might be cleaner to just disable auditing in general (there is a kernel cmdline arg for that)

mhartmay: Thanks, can you please add this as comments to the code? Otherwise no one will remember :)

rbmarliere: > Instead of masking the audit service it might be cleaner to just disable auditing in general (there is a kernel cmdline arg for that)

But then it wouldn't be configurable. I.e. the user might enable auditing by removing the mask and deal with the errors.

> Thanks, can you please add this as comments to the code? Otherwise no one will remember :)

Sure, but since its a JSON config. file, where do you suggest it goes? In the README?

mhartmay: Oh no, JSON :D In the commit message at least?

virtme/commands/run.py
```diff
@@ -1293,28 +1301,51 @@ def do_it() -> int:
...
-                    f"exec /run/virtme/guesttools/{virtme_init_cmd}",
+                    + "virtme.cache /run/virtme/cache",
+                    "SYSTEMD_UNIT_PATH=/run/virtme/cache: exec /sbin/init",
                 ]
-            ),
-        ]
+            )
+        else:
+            initsh.append(f"exec /run/virtme/guesttools/{virtme_init_cmd}")
+        initcmds = ["init=/bin/sh", "--", "-c", ";".join(initsh)]
```
mhartmay, line 1348:
```suggestion
        initcmds = ["init=/bin/sh", "--", "-c", "; ".join(initsh)]
```
Makes it easier to read... But a little out-of-scope... :)

virtme/commands/run.py
```diff
@@ -1397,12 +1428,14 @@ def do_it() -> int:
...
         qemuargs.extend(["-echr", "1"])
 
-        # Redirect kernel errors to stderr, creating a separate console.
-        #
-        # If we don't have access to stderr via procfs (for example when
-        # running inside a container), print a warning and implicitly
-        # suppress the kernel errors redirection.
-        if can_access_file("/proc/self/fd/2"):
+        if args.systemd:
+            pass
```
mhartmay, line 1432: Can you please add a comment why nothing needs to be done here?

cfg/virtme-ng.conf
rbmarliere: Added in https://github.com/arighi/virtme-ng/pull/280/commits/4ffb83c45a2782f631b7ec48bfa07edca902ec4c

virtme/commands/run.py
rbmarliere: Added in https://github.com/arighi/virtme-ng/pull/280/commits/4ffb83c45a2782f631b7ec48bfa07edca902ec4c

virtme/commands/run.py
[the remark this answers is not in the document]
rbmarliere: > Out of curiosity, I' just checked the source code of systemd [1] and they create symlinks to /dev/null. What was the motivation to use `mount --bind` instead? :)
> 
> [1] https://github.com/systemd/systemd/blob/f9d76a8d1596927b52eded4ce7e8bbf512aff551/src/debug-generator/debug-generator.c#L222-L223

I was unable to make it work with the cmdline argument. E.g.:

```
diff --git a/virtme/commands/run.py b/virtme/commands/run.py
index 0c9e628238cc..4391934abe3d 100644
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1305,14 +1305,11 @@ def do_it() -> int:
     if args.systemd:
         kernelargs.append("virtme_systemd=1")
         kernelargs.append(f"console={arch.serial_console_args()[0]}")
-        initsh.extend(
-            ["mount --bind /dev/null " + unit for unit in get_conf("systemd.masks")]
-        )
+        kernelargs.append("systemd.mask=systemd-fstab-generator")
 
     if args.root == "/":
         if args.systemd:
-            initsh.append(f"SYSTEMD_UNIT_PATH={CACHE_DIR}: exec /sbin/init")
-            initcmds = ["init=/bin/sh", "--", "-c", "; ".join(initsh)]
+            initcmds = ["init=/sbin/init"]
         else:
             initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
     else:
```

```
[    0.906040][  T382] (sd-exec-[382]: About to execute /usr/lib/systemd/system-generators/systemd-fstab-gen
erator (null) 
(...)
[    0.952893][  T397] systemd-fstab-generator[397]: Parsing /etc/fstab...
```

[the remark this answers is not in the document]
rbmarliere: [It is a valid method](https://www.freedesktop.org/software/systemd/man/latest/systemd.generator.html) to mask it:

```
A symlink to /dev/null or an empty file can be used to mask a generator, thereby preventing it from running. 
```

A bind mount would also be a valid mask in a read-only filesystem (which is the case of args.root="/", right?).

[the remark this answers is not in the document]
rbmarliere: If I target another root:

```
[    0.000000][    T0] Command line: virtme_hostname=virtme-ng nr_open=1048576 virtme_systemd=1 console=ttyS0 systemd.mask=systemd-fstab-generator virtme_root_mods=1 virtme_rw_overlay0=/etc virtme_rw_overlay1=/lib virtme_rw_overlay2=/home virtme_rw_overlay3=/opt virtme_rw_overlay4=/srv virtme_rw_overlay5=/usr virtme_rw_overlay6=/var virtme_rw_overlay7=/tmp earlyprintk=serial,ttyS0,115200 virtme_console=ttyS0 psmouse.proto=exps " virtme_stty_con=rows 76 cols 108 iutf8" TERM=screen-256color virtme_root_user=1 debug init=/bin/sh -- -c "mount -t tmpfs run /run; mkdir -p /run/virtme/guesttools; /bin/mount -n -t 9p -o ro,version=9p2000.L,trans=virtio,access=any virtme.guesttools /run/virtme/guesttools; mkdir -p /run/virtme/cache; /bin/mount -n -t 9p -o ro,version=9p2000.L,trans=virtio,access=any virtme.cache /run/virtme/cache; SYSTEMD_UNIT_PATH=/run/virtme/cache: exec /sbin/init"
(...)
[    0.803586][    T1] systemd[1]: Linked unit file: /run/systemd/generator.early/systemd-fstab-generator.se
rvice → /dev/null
```

I guess in the case of args.root="/" it is unable to create the symlink and fail silently, thus proceeding with the generator execution.

virtme_ng/utils.py
```diff
@@ -27,3 +29,38 @@ def wrapper(*args, **kwargs):
...
+        Path(Path.home(), ".virtme-ng.conf"),
+        Path("/etc", "virtme-ng.conf"),
+    )
+    for conf in configs:
+        if conf.exists():
+            return conf
+    return None
+
+
+def get_conf(key_path):
```
mhartmay, line 52: Can you please a docstring with doctest for it?

virtme_ng/utils.py
```diff
@@ -27,3 +29,38 @@ def wrapper(*args, **kwargs):
...
+    keys = key_path.split(".")
+
+    with open(conf_path, encoding="utf-8") as conf_fd:
+        conf = json.loads(conf_fd.read())
+        try:
+            for key in keys:
+                conf = conf[key]
+            return conf
+        except (KeyError, TypeError):
+            return []
```
mhartmay, line 66: Please print  a warning.

virtme/commands/run.py
```diff
@@ -1840,6 +1874,24 @@ def get_net_mac(index):
...
+StandardInput=tty
+StandardOutput=tty
+TTYPath=/dev/%I
+TTYReset=yes
+TTYVHangup=yes
+Environment=""")
+                for arg in kernelargs:
+                    match = re.match(r"(virtme_.*)=(.*)", arg)
+                    if match:
+                        f.write(f"{match.group(1)}='{match.group(2)}' ")
```
mhartmay, line 1890:
```suggestion
        if args.systemd:
              init_environment_vars = []
              for arg in kernelargs:
                    match = re.match(r"(virtme_.*)=(.*)", arg)
                    if match:
                        init_environment_vars.append(f"{match.group(1)}={shlex.quote(match.group(2))})
             os.makedirs(CACHE_DIR, exist_ok=True)
             with open(SERIAL_GETTY_FILE, "w", encoding="utf-8") as f:
                 f.write(f"""[Service]
StandardInput=tty
StandardOutput=tty
TTYPath=/dev/%I
TTYReset=yes
TTYVHangup=yes
Environment={shlex.quote(shlex.join(init_environment_vars))}""")
```

(untested, and probably is the indentation off).

virtme/commands/run.py
```diff
@@ -1293,28 +1301,51 @@ def do_it() -> int:
...
+            initsh.append(f"SYSTEMD_UNIT_PATH={CACHE_DIR}: exec /sbin/init")
+            initcmds = ["init=/bin/sh", "--", "-c", "; ".join(initsh)]
+        else:
+            initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
     else:
         virtfs_config = VirtFSConfig(
             path=guest_tools_path,
             mount_tag="virtme.guesttools",
         )
         export_virtfs(qemu, arch, qemuargs, virtfs_config)
```
mhartmay, line 1323:
```suggestion
```
Can be dropped now, no?

virtme/commands/run.py
```diff
@@ -1293,28 +1301,51 @@ def do_it() -> int:
...
-        initcmds = [
-            "init=/bin/sh",
-            "--",
-            "-c",
-            ";".join(
+        virtfs_config = VirtFSConfig(
+            path=str(CACHE_DIR),
+            mount_tag="virtme.cache",
+        )
+        export_virtfs(qemu, arch, qemuargs, virtfs_config)
```
mhartmay, line 1328: Is this always needed or only in case of systemd?

rbmarliere: Good catch, it's only for systemd (the serial-getty@.service file is written to the CACHE_DIR).

virtme/commands/run.py
```diff
@@ -1293,28 +1301,51 @@ def do_it() -> int:
     else:
         virtme_init_cmd = "virtme-init"
 
+    initsh = []
+    if args.systemd:
+        kernelargs.append("virtme_systemd=1")
+        kernelargs.append(f"console={arch.serial_console_args()[0]}")
```
mhartmay, line 1307:
```suggestion
        kernelargs.append(" ".join([f"console={console}" for console in arch.serial_console_args()}")
```

Should have suggested it already in the beginning.

rbmarliere: Fixed in https://github.com/arighi/virtme-ng/pull/280/commits/3c7a573327e33912d2f87dd43acff2f30fe2931d

virtme/commands/run.py
rbmarliere: Fixed in https://github.com/arighi/virtme-ng/pull/280/commits/918d2069374938b3936c6fbe60d8b5bda72a7aa6

virtme/commands/run.py
rbmarliere: Thanks, fixed in https://github.com/arighi/virtme-ng/pull/280/commits/918d2069374938b3936c6fbe60d8b5bda72a7aa6. However, the second quote() in Environment= should not be needed, right? Otherwise it yields some weird string that breaks the unit file.

virtme_ng/utils.py
rbmarliere: https://github.com/arighi/virtme-ng/pull/280/commits/b24ab70280e65814e27cf9e877b2f7ac87c0cd7d

rbmarliere: Added in https://github.com/arighi/virtme-ng/pull/280/commits/b24ab70280e65814e27cf9e877b2f7ac87c0cd7d

virtme/commands/run.py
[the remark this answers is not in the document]
rbmarliere: Thank you! Added in https://github.com/arighi/virtme-ng/pull/280/commits/bc013d3a271b4b3a9c6cc185c68c3f624f899344

virtme/commands/run.py
rbmarliere: For reference,

```
Environment='virtme_hostname=virtme-ng virtme_systemd=1 virtme_root_mods=1 virtme_rw_overlay0=/etc virtme_rw_overlay1=/lib virtme_rw_overlay2=/home virtme_rw_overlay3=/opt virtme_rw_overlay4=/srv virtme_rw_overlay5=/usr virtme_rw_overlay6=/var virtme_rw_overlay7=/tmp virtme_console=ttyS0 '"'"'virtme_stty_con='"'"'"'"'"'"'"'"'rows 76 cols 108 iutf8'"'"'"'"'"'"'"'"''"'"' virtme_chdir=mnt/ext/src/linux/virtme-ng/sys    temd virtme_root_user=1' 
```
vs
```
Environment=virtme_hostname=virtme-ng virtme_systemd=1 virtme_root_mods=1 virtme_rw_overlay0=/etc virtme_rw_overlay1=/lib virtme_rw_overlay2=/home virtme_rw_overlay3=/opt virtme_rw_overlay4=/srv virtme_rw_overlay5=/usr virtme_rw_overlay6=/var virtme_rw_overlay7=/tmp virtme_console=ttyS0 virtme_stty_con='rows 76 cols 108 iutf8' virtme_chdir=mnt/ext/src/linux/virtme-ng/systemd virtme_root_user=1
```

virtme_ng/utils.py
rbmarliere: Added in https://github.com/arighi/virtme-ng/pull/280/commits/b24ab70280e65814e27cf9e877b2f7ac87c0cd7d

virtme/commands/run.py
mhartmay: I've updated my original answer [1].

[1] https://github.com/arighi/virtme-ng/pull/280#discussion_r2086860872

rbmarliere: Perfect, it works. Thanks for looking into this. Added in https://github.com/arighi/virtme-ng/pull/280/commits/3988d3ce5731627b2bf8663aeca55e653fdd710d

virtme/commands/run.py
```diff
@@ -1840,6 +1876,30 @@ def get_net_mac(index):
...
     qemuargs.extend(["-kernel", kernel.kimg])
     if kernelargs:
+        if args.systemd:
+            init_environment_vars = []
+            for arg in kernelargs:
+                match = re.match(r"(virtme_.*)=(.*)", arg)
+                if not match:
+                    continue
+                init_environment_vars.append(
+                    shlex.quote(f"{match.group(1)}={match.group(2)}")
```
mhartmay, line 1886:
```suggestion
                    f"{match.group(1)}={match.group(2)}"
```
should be enough. The `shlex.join` handles the quoting for us.

rbmarliere: Indeed, done in https://github.com/arighi/virtme-ng/pull/280/commits/6207c487e808843d57fcfba1508fb581a1024f55

marcosps:
I just run some tests on s390 and x86_64, and it works as expected. Some caveats when using `--systemd`:

* if you use `--root /some/dir`, everything works as before. Good work @rbmarliere !!!

Now, if you don't user `--root`:
* You need to run as `root` and specify `--user root` to have everything working `perfectly` for now, since some files under /etc are owned by root on the rootfs, so there is no help: virtiofsd can't access those files as an ordinary user.
* There are a couple errors and warnings when using `virtme-init` (the bash one), related to `cgroups` and `/sys` being already mounted, but these are harmless. On `virtme-ng-init` (in rust) it works without these errors. But again, it works in the end.
* There is a service failing on openSUSE Tumbleweed, which moves `/var/log/boot.msg` to `/var/log/boot.omsg`, and it fails. I would expect this to be mounted used overlayfs, and thus it would be able to move the file on rw layer, but again, it's harmless.

As I said, it works: you can start services (tested are sshd and libvirt, but systemd is consistent), and I believe that it can be merged as it is. The details above can be squared later, since it's said that systemd support is still experimental. Thanks a lot @rbmarliere for doing this amazing work!

marcosps:
@matttbe @mhartmay @arighi what do you think about the latest version? :)

virtme/commands/run.py
```diff
@@ -1293,28 +1301,53 @@ def do_it() -> int:
...
         virtme_init_cmd = "virtme-init"
 
+    initsh = []
+    if args.systemd:
+        kernelargs.append("virtme_systemd=1")
+        kernelargs.append(
+            " ".join([f"console={console}" for console in arch.serial_console_args()])
+        )
+        initsh.extend(
+            [f"mount --bind /dev/null {unit}" for unit in get_conf("systemd.masks")]
```
arighi, line 1311: You may want to do:
```
 [f"mount --bind /dev/null {unit}" for unit in get_conf("systemd.masks") or []]
```

Otherwise if `systemd.masks` isn't defined we get a `'NoneType' object is not iterable` error.

arighi (changes requested):
I still need to review this in details, I left a small comment, but in general it looks good and it doesn't seem to break anything.

It still doesn't work out of the box on Arch or Ubuntu, but, again, I still need to play with it a bit.

With that little suggestion I made I think we can merge this and keep improving in tree. It seems pretty self-contained and it shouldn't break anything else. Thanks!

virtme/commands/run.py
rbmarliere: Fixed in https://github.com/arighi/virtme-ng/pull/280/commits/ca98cdc43f8bc3442dc1e34d71c09982ea585834!

Thanks @arighi I'll try it with Ubuntu|Arch next :)

virtme/commands/run.py
```diff
@@ -1293,28 +1301,56 @@ def do_it() -> int:
     else:
         virtme_init_cmd = "virtme-init"
 
+    initsh = []
+    if args.systemd:
+        kernelargs.append("virtme_systemd=1")
```
matttbe, line 1306: detail: if you only need this in the init script, maybe you can simply add "virtme.systemd", no need to have an env var, no?

```suggestion
        kernelargs.append("virtme.systemd")
```

See how other var are used, e.g. `virtme.snapd`.

Also, when the init script is started with systemd, does systemd not set specific env vars or create specific files? (or maybe `$$` / `getpid()` would not be 1)

It is more a detail, already a lot of things are being exported, but if that's not needed and simple to get the info elsewhere, maybe that would be easier to deal with? (not sure, up to you :) )

virtme/guest/virtme-init
```diff
@@ -19,8 +19,12 @@ log() {
 mount -t proc -o nosuid,noexec,nodev proc /proc/
 mount -t sysfs -o nosuid,noexec,nodev sys /sys/
 
-# Mount tmpfs dirs
-mount -t tmpfs -o mode=0755 run /run/
+if [[ -z $virtme_systemd || $virtme_systemd == '0' ]]; then
```
matttbe, line 22: Detail: maybe clearer with:

```suggestion
if [[ "$virtme_systemd" != '1' ]]; then
```

virtme_ng/run.py
```diff
@@ -256,6 +255,12 @@ def make_parser():
         + "(useful with --arch)",
     )
 
+    parser.add_argument(
```
matttbe, line 258: (detail: maybe best to move this options elsewhere, not in the middle of the "filesystems" options, at least not between `--root` and `--root-release` ; but not sure if there is really a good spot :) maybe at the end, maybe we should re-organise the options)

virtme_ng/utils.py
```diff
@@ -27,3 +29,50 @@ def wrapper(*args, **kwargs):
...
+
+    with open(conf_path, encoding="utf-8") as conf_fd:
+        conf = json.loads(conf_fd.read())
+        try:
+            for key in keys:
+                conf = conf[key]
+            return conf
+        except (KeyError, TypeError):
+            print(f"WARNING: Key {key_path} not found in {conf_path}.")
+            return []
```
matttbe, line 78: why returning an empty list and not `None` like above or another type? We don't know the expected type, it could be a string, a list, a dictionary, etc.

(I guess you should return the default value instead)

virtme_ng/utils.py
```diff
@@ -27,3 +29,50 @@ def wrapper(*args, **kwargs):
...
+    keys = key_path.split(".")
+
+    with open(conf_path, encoding="utf-8") as conf_fd:
+        conf = json.loads(conf_fd.read())
+        try:
+            for key in keys:
+                conf = conf[key]
+            return conf
+        except (KeyError, TypeError):
+            print(f"WARNING: Key {key_path} not found in {conf_path}.")
```
matttbe, line 77: if you have a warning here, maybe you need one above as well? (no config file)

(or maybe it is not needed if you return a good default value instead)

cfg/virtme-ng.conf
```diff
@@ -1,4 +1,11 @@
 {
-    "default_opts" : {
+    "default_opts": {},
+    "systemd": {
+        "masks": [
```
matttbe, line 4: If I'm not mistaken, the corresponding local config file will not be updated during the upgrade. I even think that this default config file might not even be installed, e.g. when `virtme-ng` is installed with `pip` or other Python installers. So if this is correct, I guess you will need to handle that case, e.g. the new `get_conf()` should return a default value given in argument.

virtme_ng/utils.py
```diff
@@ -27,3 +29,50 @@ def wrapper(*args, **kwargs):
...
+        Path(Path.home(), ".virtme-ng.conf"),
+        Path("/etc", "virtme-ng.conf"),
+    )
+    for conf in configs:
+        if conf.exists():
+            return conf
+    return None
+
+
+def get_conf(key_path):
```
matttbe, line 52: see above, I think this helper should handle default values if the config is not set. The caller could also pass `None` or `[]`, etc. if needed.

virtme_ng_init/src/main.rs
```diff
@@ -193,7 +193,8 @@ const SYSTEM_MOUNTS: &[MountInfo] = &[
 const USER_SCRIPT: &str = "/run/tmp/.virtme-script";
 
 fn check_init_pid() {
```
matttbe, line 195: (maybe this check no longer makes sense, maybe we can remove it? Instead, we could check if `id()` != 1 to know if systemd is being used or not)

matttbe (changes requested):
thank you for looking at this!

I only reviewed the code, I didn't test (yet). It looks good, just a few small comments! The main one is about the way the config file is handled today: this file might not contain the new default, the code needs to handle that by using default values when needed.

virtme/commands/run.py
```diff
@@ -1840,6 +1879,28 @@ def get_net_mac(index):
     # Load a normal kernel
     qemuargs.extend(["-kernel", kernel.kimg])
     if kernelargs:
+        if args.systemd:
+            init_environment_vars = []
+            for arg in kernelargs:
+                match = re.match(r"(virtme_.*)=(.*)", arg)
+                if not match:
+                    continue
+                init_environment_vars.append(f"{match.group(1)}={match.group(2)}")
```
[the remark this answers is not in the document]
rbmarliere: I think it would be best to keep, e.g. if the user is debugging the cmdline and what virtme is passing.

cfg/virtme-ng.conf
rbmarliere: Added a commit addressing the config: https://github.com/arighi/virtme-ng/pull/280/commits/95158951d1bbe11971755083837e58ba177e2a55

virtme/commands/run.py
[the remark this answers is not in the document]
matttbe: fine for me, safer to keep them.

README.md
```diff
@@ -584,9 +600,17 @@ Troubleshooting
...
+   minimum system environment without systemd), if some snaps are not running
+   try to disable apparmor, adding `--append="apparmor=0"` to the virtme-ng
+   command line.
+
+ - Systemd support (`--systemd`) is still experimental, be aware that you might
+   also need `--user root`, or if you're using your own `/` as ROOTFS, you need
+   to run vng itself as root. If something does not work for you, you might
+   need to mask some services. You can see the ones that are masked by default
+   in the CONF_DEFAULT definition, which can be customized via configuration file
+   (refer to the "Default options" section).
```
matttbe, line 613: I think it would help if you add an example here, a bit similar to what was done for `default_opts`, because reading the code to know how to write the config file is not nice for the users I think :)

I would prefer to have the default values only in a config file that is shipped with vng, but it doesn't seem to work. We might then have to maintain a copy in the `README` to help users (or a link to the code, but then we will depend on GitHub).

If you do that, please add a comment above the declaration of `CONF_DEFAULT` as a reminder to update the README file if we change it.

virtme_ng/utils.py
```diff
@@ -27,3 +40,46 @@ def wrapper(*args, **kwargs):
...
+        CONF_FILE,
+        Path(Path.home(), ".virtme-ng.conf"),
+        Path("/etc", "virtme-ng.conf"),
+    )
+    for conf_path in conf_paths:
+        if conf_path.exists():
+            with open(conf_path, encoding="utf-8") as conf_fd:
+                conf = json.loads(conf_fd.read())
+                return conf
+    return CONF_DEFAULT
```
matttbe, line 62: mmh, I think that's not enough: some people might already have this file, but it might not contain new entries added in a newer version. In this case, your new helper here will return the content of the file, but `get_conf()` will return `None` + a warning. (see below)

I think we can avoid the warning by either asking `get_conf()` caller to provide the default value, or by looking at the default value in `CONF_DEFAULT`. WDYT?

matttbe (changes requested):
thank you for the update! I think the new version helps, but it seems not enough to properly handle updates.

virtme_ng/utils.py
rbmarliere: Ah, good catch. I think it would be best to simply return the default. I'll give it a spin.

README.md
rbmarliere: As a user, I'd prefer to have some proper documentation and leave configuring to the user, so it's best to have the default in the README. Do you agree with this approach?

matttbe: Sorry, end of the week, I don't think I understood your approach. Do you mind rephrase it please? :)

I think the users need some guidance on how to change the configuration. If it cannot be done by simply modifying a local file, it would be good to show how it should look like, no?

virtme_ng/utils.py
matttbe: On one hand, `CONF_DEFAULT` regroups all the default values for the config file in the same place. If the caller provides the default value, this value would be closer to where it is used. Both versions have pros and cons :)

README.md
rbmarliere: Added https://github.com/arighi/virtme-ng/pull/280/commits/878eac47344c0cfc78e0476feadc2317ef56e91c, not sure about the wording but that seems to be the way to go IMHO.

virtme/commands/run.py
```diff
@@ -1297,28 +1305,55 @@ def do_it() -> int:
...
 
+    initsh = []
+    if args.systemd:
+        kernelargs.extend(
+            [f"console={console}" for console in arch.serial_console_args() or []]
+        )
+        initsh.extend(
+            [
+                f"mount --bind /dev/null {unit}"
+                for unit in get_conf("systemd.masks") or []
```
septatrix, line 1316: Why not just add `systemd.mask=foobar.service` to the kernel cmdline?

README.md
```diff
@@ -294,6 +319,22 @@ Examples
    Linux version 6.7.0-060700rc5-generic (kernel@kathleen) (x86_64-linux-gnu-gcc-13 (Ubuntu 13.2.0-7ubuntu1) 13.2.0, GNU ld (GNU Binutils for Ubuntu) 2.41) #202312102332 SMP PREEMPT_DYNAMIC Sun Dec 10 23:41:31 UTC 2023
 ```
 
+ - Run a kernel within a separate chroot, using its systemd as init:
+```
+   $ vng -r ./fookernel --user root --root ./rootfs/sid --user root --systemd -- systemctl status | head
```
septatrix, line 324: `--user root` is listed twice?

septatrix (changes requested):
Sadly I have not been able to get it properly running so far but I think my whole vng repo is borked right now so it's likely the issue is on my end

virtme/commands/run.py
```diff
@@ -1842,6 +1880,28 @@ def get_net_mac(index):
...
+StandardOutput=tty
+TTYPath=/dev/%I
+TTYReset=yes
+TTYVHangup=yes
+Environment={shlex.join(init_environment_vars)}"""
+                )
+                if args.root == "/":
+                    f.write(f"\nExecStart={guest_tools_path}/{virtme_init_cmd}")
+                else:
+                    f.write(f"\nExecStart=/run/virtme/guesttools/{virtme_init_cmd}")
```
septatrix, line 1904: What exactly is the purpose of still spawning the vng inits?
Also it might be cleaner to use `systemd-run-generator` though if tty connection is desired I think we would still need at least a drop-in

virtme/commands/run.py
matttbe: > The kernel cmdline arg is set by qemu so before starting systemd? I am saying that you can forego all those devnull bind mounts because normal units can be masked by `systemd.mask=*` and the generators can be disabled by setting `fstab=no` and `luks=no` which is a lot cleaner

Something I'm probably missing: do you want to set `systemd.masks` and use that in the virtme-ng init scripts? Or are all these variables going to be interpreted by systemd?

rbmarliere: Sorry, had to delete my wrong answer. Indeed, it works:

```
# cat /etc/virtme-ng.conf 
{
        "default_opts": {},
        "systemd": {
                "masks": []
        }
}
# vng -r --systemd -a "fstab=no" -a "luks=no" --exec "systemctl status"                                     
● virtme-ng                                                                                                 
    State: starting                                                                                         
    Units: 392 loaded (incl. loaded aliases)                                                                
     Jobs: 4 queued                                                                                         
   Failed: 3 units                                                                                          
    Since: Mon 2025-05-26 06:42:03 -03; 4s ago                                                              
  systemd: 257.5+suse.8.gc10a66fb4d                                                                         
  Tainted: unmerged-bin                                                                                     
   CGroup: /
(...)
```

Thanks for the approach @septatrix, I'll see what to do now.

matttbe: Ah great, let's do that then! Thanks for the suggestion and for working on that!

virtme/commands/run.py
rbmarliere: Oh I just saw this comment after pushing the new changes... Thanks for the pointers. It's a good approach but then we would only support roots with systemd > 255, right?

virtme/commands/run.py
```diff
@@ -1297,28 +1304,54 @@ def do_it() -> int:
     else:
         virtme_init_cmd = "virtme-init"
 
+    if args.systemd:
+        kernelargs.append("fstab=no")
+        kernelargs.append("luks=no")
+        kernelargs.append("audit=off")
```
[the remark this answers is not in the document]
rbmarliere: Ah, good call. I'll do that.

PR #281, opened 2025-04-20
virtme-ng: Do not truncate stdout/stderr on process exit

arighi:
When an external command is executed by vng, its stdout or stderr might get truncated if the child process exits before all the output is consumed.

This can happen because data may still be buffered in the stdout or stderr pipes after the child terminates.

Fix this by making sure both stdout and stderr are fully drained, even after the process has exited, ensuring that all output is captured and correctly reported.

Example:

 - before this change:

       $ vng -vb --config /no-file
       cmd: virtme-configkernel --defconfig --verbose --no-update --custom /no-file
       usage: virtme-configkernel [-h] [--arch ARCHITECTURE] [--cross-compile CROSS_COMPILE_PREFIX]

 - after this change:

       $ vng -vb --config /no-file
       cmd: virtme-configkernel --defconfig --verbose --no-update --custom /no-file
       usage: virtme-configkernel [-h] [--arch ARCHITECTURE] [--cross-compile CROSS_COMPILE_PREFIX] [--custom CUSTOM] [--configitem CONFITEM] [--no-update] [--verbose] [--allnoconfig | --defconfig | --update] [envs ...]
       virtme-configkernel: error: argument --custom: '/no-file' is not a valid file.

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -37,6 +37,7 @@ def check_call_cmd(command, quiet=False, dry_run=False):
     if dry_run:
         print(" ".join(command))
         return
+
     with Popen(
         command,
         stdout=PIPE,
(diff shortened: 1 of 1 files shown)


virtme_ng/run.py
```diff
@@ -49,20 +50,38 @@ def check_call_cmd(command, quiet=False, dry_run=False):
...
 
+        stdout_open = True
+        stderr_open = True
+
         # Use select to poll for new data in the file descriptors
-        while process.poll() is None:
-            ready_to_read, _, _ = select([stdout_fd, stderr_fd], [], [], 1)
-            for file in ready_to_read:
-                if file == stdout_fd:
+        while stdout_open or stderr_open:
```
matttbe, line 57: I guess we could use a `while True` here if you validate these conditions just after, but that's a detail, no need to change :)

virtme_ng/run.py
```diff
@@ -49,20 +50,38 @@ def check_call_cmd(command, quiet=False, dry_run=False):
         stdout_fd = process.stdout.fileno()
         stderr_fd = process.stderr.fileno()
 
+        stdout_open = True
+        stderr_open = True
+
         # Use select to poll for new data in the file descriptors
-        while process.poll() is None:
```
matttbe, line 53: just to be sure I understand the issue: was the problem before that this could return `None` while there were still data to read?

Do you need the rest of the modifications to avoid calling `select()` with a `fd` where we had a `EOF` even if it has not been closed?

(on the other hand, the new code is not very complex, I'm happy to have this merged :) )

matttbe (commented):
thank you for the modifications! I just have one question, but the new version looks good, feel free to merge!

virtme_ng/run.py
arighi: Oh that's right, let me change that to a `while True`, thanks for noticing!

virtme_ng/run.py
arighi: Right, the problem is that `process.poll()` can return `None` if either stdout or stderr are closed, but we need to consume all data from both.

And yes, we can probably compact the code a bit more and always pass both fds to select(), even if one is EOF'd the behavior should be the same. Let me try to make the code a bit more compact.

matttbe: thank you for the explanation and the update!

matttbe (approved):

PR #282, opened 2025-04-20
ssh/vsock: Use "/" as the default CID/port separator

arighi:
Some versions of ssh reject "," as CID separator, resulting in errors like:

 $ vng --ssh-client
 hostname contains invalid characters

Switch to "/" as the default separator, as it appears to be more portable and compatible across different ssh versions.

Fixes: 24e056b ("virtme: Add vsock support for the SSH connection to the guest")


arighi:
cc: @mhartmay what do you think about this?

arighi:
> > Some versions of ssh reject "," as CID separator
> 
> Do you know which versions don't support that?
> 
> We could also use `=` or `%` as separator if it is supported by older SSH versions, but also scp and rsync. It is still time to change `,` on systemd as well if there is a "better" one.

I'm seeing this with:
```
arighi@gpd3~> ssh -V
OpenSSH_10.0p2, OpenSSL 3.5.0 8 Apr 2025
arighi@gpd3~> ssh -F /home/arighi/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf -l arighi virtme-ng,2222
hostname contains invalid characters
```

It seems happy both with `=` or `%`, then I guess `,` is the only one problematic...

matttbe:
> > > Some versions of ssh reject "," as CID separator
> > 
> > 
> > Do you know which versions don't support that?
> > We could also use `=` or `%` as separator if it is supported by older SSH versions, but also scp and rsync. It is still time to change `,` on systemd as well if there is a "better" one.
> 
> I'm seeing this with:
> 
> ```
> arighi@gpd3~> ssh -V
> OpenSSH_10.0p2, OpenSSL 3.5.0 8 Apr 2025
> arighi@gpd3~> ssh -F /home/arighi/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf -l arighi virtme-ng,2222
> hostname contains invalid characters
> ```

That's the latest version released this month!

In the [release notes](https://www.openssh.com/releasenotes.html#10.0p1): 

>  * ssh(1): prohibit the comma character in hostnames accepted, but allow an underscore as the first character in a hostname

Here is the linked commit: https://github.com/openssh/openssh-portable/commit/487cf4c18c123b66c1f3f733398cd37e6b2ab6ab

It is not clearly explained why this has been introduced.

But is scp still OK with the comma?

> It seems happy both with `=` or `%`, then I guess `,` is the only one problematic...

Maybe better to change then? Same in systemd?

arighi:
> > > > Some versions of ssh reject "," as CID separator
> > > 
> > > 
> > > Do you know which versions don't support that?
> > > We could also use `=` or `%` as separator if it is supported by older SSH versions, but also scp and rsync. It is still time to change `,` on systemd as well if there is a "better" one.
> > 
> > 
> > I'm seeing this with:
> > ```
> > arighi@gpd3~> ssh -V
> > OpenSSH_10.0p2, OpenSSL 3.5.0 8 Apr 2025
> > arighi@gpd3~> ssh -F /home/arighi/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf -l arighi virtme-ng,2222
> > hostname contains invalid characters
> > ```
> 
> Thank you for having shared this. That's the latest version released this month!

Yeah Arch is moving pretty fast...

> 
> In the [release notes](https://www.openssh.com/releasenotes.html#10.0p1):
> 
> > * ssh(1): prohibit the comma character in hostnames accepted, but allow an underscore as the first character in a hostname
> 
> Here is the linked commit: [openssh/openssh-portable@487cf4c](https://github.com/openssh/openssh-portable/commit/487cf4c18c123b66c1f3f733398cd37e6b2ab6ab)
> 
> It is not clearly explained why this has been introduced.

Ohh thanks for digging into this.

> 
> But is scp still OK with the comma?

Nope, scp is also complaining:
```
$ scp -F ~/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng,2222:~/results.txt .
hostname contains invalid characters
scp: Connection closed
```

> 
> > It seems happy both with `=` or `%`, then I guess `,` is the only one problematic...
> 
> Maybe better to change then? Same in systemd?

At this point I think it'd be better to drop `,` and use something else, `%` seems to work fine both with ssh and scp.

matttbe:
> > > > > Some versions of ssh reject "," as CID separator
> > > > 
> > > > 
> > > > Do you know which versions don't support that?
> > > > We could also use `=` or `%` as separator if it is supported by older SSH versions, but also scp and rsync. It is still time to change `,` on systemd as well if there is a "better" one.
> > > 
> > > 
> > > I'm seeing this with:
> > > ```
> > > arighi@gpd3~> ssh -V
> > > OpenSSH_10.0p2, OpenSSL 3.5.0 8 Apr 2025
> > > arighi@gpd3~> ssh -F /home/arighi/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf -l arighi virtme-ng,2222
> > > hostname contains invalid characters
> > > ```
> > 
> > 
> > Thank you for having shared this. That's the latest version released this month!
> 
> Yeah Arch is moving pretty fast...

I originally thought only old versions didn't support the comma. I'm surprised that's the new ones which have an issue.

> > In the [release notes](https://www.openssh.com/releasenotes.html#10.0p1):
> > > * ssh(1): prohibit the comma character in hostnames accepted, but allow an underscore as the first character in a hostname
> > 
> > 
> > Here is the linked commit: [openssh/openssh-portable@487cf4c](https://github.com/openssh/openssh-portable/commit/487cf4c18c123b66c1f3f733398cd37e6b2ab6ab)
> > It is not clearly explained why this has been introduced.
> 
> Ohh thanks for digging into this.
> 
> > But is scp still OK with the comma?
> 
> Nope, scp is also complaining:
> 
> ```
> $ scp -F ~/.cache/virtme-ng/.ssh/virtme-ng-ssh.conf virtme-ng,2222:~/results.txt .
> hostname contains invalid characters
> scp: Connection closed
> ```

Arf... Thank you for having checked!


> > > It seems happy both with `=` or `%`, then I guess `,` is the only one problematic...
> > 
> > 
> > Maybe better to change then? Same in systemd?
> 
> At this point I think it'd be better to drop `,` and use something else, `%` seems to work fine both with ssh and scp.

Good idea! I can suggest a modification on systemd side as well.

matttbe (changes requested):
> Good idea! I can suggest a modification on systemd side as well.

In fact, they already switched to `%` a few days ago for the same reason :)

Then, can we switch to `("%", "/")` as well, and adapt the README file? :)

matttbe:
> Then, can we switch to `("%", "/")` as well, and adapt the README file? :)

I hope that's OK, I just did this [modification](https://github.com/arighi/virtme-ng/compare/a2bea07fe28137770b3300acecb34686596a0669..e0ca24f84cbe87adb9a84b98bd3abe52457dfee8) and updated the commit message :)

arighi:
> > Then, can we switch to `("%", "/")` as well, and adapt the README file? :)
> 
> I hope that's OK, I just did this [modification](https://github.com/arighi/virtme-ng/compare/a2bea07fe28137770b3300acecb34686596a0669..e0ca24f84cbe87adb9a84b98bd3abe52457dfee8) and updated the commit message :)

That's perfect! Thanks for driving this @matttbe!

arighi:
We also need to update `virtme-ssh-proxy`, updated the commit with this change as well.

matttbe:
> We also need to update `virtme-ssh-proxy`, updated the commit with this change as well.

Good catch, sorry, I forgot to check :)

matttbe (approved):

mhartmay:
> cc: @mhartmay what do you think about this?

Sry for the late response & thanks for fixing this!

PR #285, opened 2025-04-24
console: abort if socat is not available

matttbe:
'socat' is used to connect to the guest via VSOCK.

Instead of throwing a FileNotFoundError exception because 'socat' is not available, abort before with a more comprehensive message.


arighi (approved):
Tested and LGTM, thanks!

PR #287, opened 2025-04-28
Small improvements

mhartmay:


virtme/commands/run.py
```diff
@@ -1856,7 +1856,7 @@ def get_net_mac(index):
         qemuargs.extend(args.qemu_opts)
 
     if args.show_command:
-        print(" ".join(shlex.quote(a) for a in qemuargs))
```
matttbe, line 1861: just to be sure, this was already correct, but clearer with `shlex.join()`, right?

matttbe (approved):
Thanks for the small improvements!

Just one question, but not blocking.

virtme/commands/run.py
mhartmay: Yep, see

https://github.com/python/cpython/blob/b739ec5ab78ed55367516de7a11e732cb3f1081d/Lib/shlex.py#L315-L317

matttbe: ah nice, thank you for having checked!

Issue #286, opened 2025-04-25
[question] how to specify CC variable


helen-fornazier:
Hey,

I was wondering if this was supopse to work:

`vng O=.virtme/build CC="ccache gcc" --build`

I'm getting this error:

```
> vng O=.virtme/build CC='ccache gcc' --build                                                                                                                                                                     (base) 
ccache: unknown C compiler            
scripts/Kconfig.include:45: Sorry, this C compiler is not supported.
make[4]: *** [/home/ccu/ig/kernel/scripts/kconfig/Makefile:85: syncconfig] Error 1
make[3]: *** [/home/ccu/ig/kernel/Makefile:680: syncconfig] Error 2
make[2]: *** [/home/ccu/ig/kernel/Makefile:781: include/config/auto.conf.cmd] Error 2
make[1]: *** [/home/ccu/ig/kernel/Makefile:224: __sub-make] Error 2
```

I wonder if it is not splitting the `CC` variable, or if I'm doing something wrong.

Thanks!

matttbe:
> I wonder if it is not splitting the CC variable

Very likely indeed. I guess it should not be difficult to handle properly, but I cannot check for the moment. In the meantime, you can follow the recommendation from the README file:

    $ PATH="/usr/lib/ccache:${PATH}" \
        KBUILD_BUILD_TIMESTAMP=0 \
        vng --build
    # or export the two variables before, see 'man ccache' for more details

mhartmay:
Hi,

we have probably to rework large portions of the code since quoting is not handled properly.

See https://github.com/arighi/virtme-ng/blob/main/virtme_ng/run.py#L757 for example.

Let's see if I can come up with a quick fix - ideally, we can get rid of `_format_cmd`.

mhartmay:
> Hi [@mhartmay](https://github.com/mhartmay)
> 
> > Let's see if I can come up with a quick fix - ideally, we can get rid of `_format_cmd`.
> 
> Oh, I hope you didn't start, I started looking at this (and other args that were not quotes) after I merged your PR, see [#288](https://github.com/arighi/virtme-ng/pull/288). Please tell me if that's OK.

I did :) But no worries, it's good that you're fixing this stuff!

PR #288, opened 2025-04-28
Handle quotes in more commands

### LINKED ISSUE #265: SSH command line interface
Connecting to a vng instance via ssh may require additional options (see, for example, #261), which can be non trivial for users to figure out.

Currently we only provide ssh server and client helper options:
```
$ vng --ssh [PORT]
$ vng --ssh-client [PORT] [--remote-cmd CMD]
```

It'd be nice to have a more complete set of options to operate with ssh without having to figure out all the required command line options. Something like the following would break the current API unfortunately, but I think it'd be helpful:
```
# Start sshd in the vng instance
$ vng --sshd [--port PORT] [--force-tcp]

# Connect to a vng instance via ssh
$ vng --ssh [--port PORT] [--force-tcp]

# Run a command in a vng instance
$ vng --ssh [--port PORT] [--force-tcp] -- COMMAND

# Copy a file using scp (SRC or DST can prepend "virtme-ng:" to determine if the instance is the source or destination)
$ vng --scp SRC DST [--port PORT] [--force-tcp]

# Copy a file using rsync (SRC or DST can prepend "virtme-ng:" to determine if the instance is the source or destination)
$ vng --rsync SRC DST [--port PORT] [--force-tcp]
```

This is just a proposal, let me know what you think. Opinions, criticisms? @mhartmay @matttbe

matttbe:
A follow-up of #287 from @mhartmay: different fixes and one small feature.

This PR would close #265 and #286

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,9 +222,15 @@ Examples
 
  - Build the kernel using a separate build directory, and run it, in verbose:
 ```
-   $ mkdir -p .virtme/build
-   $ vng O=.virtme/build --build --verbose
-   $ vng O=.virtme/build --verbose
+   $ export KBUILD_OUTPUT=.virtme/build
+   $ vng --build --verbose
+   $ vng --verbose
+```
+
+ - Same example, but using `O=`:
+```
+   $ vng --build --verbose -- O=.virtme/build
+   $ vng --verbose -- O=.virtme/build
 ```
 
  - Accelerate the kernel rebuilds using CCache (if installed):
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -1189,10 +1189,10 @@ class KernelSource:
             for item in args.append:
                 split_items = shlex.split(item)
                 for split_item in split_items:
-                    append.append("-a " + split_item)
+                    append += ["-a", split_item]
         if args.debug:
-            append.append("-a nokaslr")
-        self.virtme_param["append"] = " ".join(append)
+            append += ["-a", "nokaslr"]
+        self.virtme_param["append"] = shlex.join(append)
 
     def _get_virtme_memory(self, args):
         if args.memory is None:
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -902,7 +902,7 @@ class KernelSource:
             else:
                 envs.append(var)
         if envs:
-            args.exec = " ".join(envs)
+            args.exec = shlex.join(envs)
         if args.exec is not None:
             self.virtme_param["exec"] = f"--script-sh {shlex.quote(args.exec)}"
         else:
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -543,17 +543,17 @@ virtme-ng is based on virtme, written by Andy Lutomirski <luto@kernel.org>.
     )
 
     g_remote.add_argument(
+        "--ssh-tcp",
+        action="store_true",
+        help="Use TCP for the SSH connection to the guest",
+    )
+
+    g_remote.add_argument(
         "--remote-cmd",
         action="store",
         metavar="COMMAND",
         help="To start in the VM a different command than the default one (--server), "
         + "or to launch this command instead of a prompt (--client).",
-    )
-
-    parser.add_argument(
-        "--ssh-tcp",
-        action="store_true",
-        help="Use TCP for the SSH connection to the guest",
     )
 
     return parser
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -869,11 +869,12 @@ class KernelSource:
             target = "bzImage"
             cross_compile = None
             cross_arch = None
-        make_command = ["make", "LOCALVERSION=-virtme"]
-        if args.compiler:
-            make_command += [f"HOSTCC={args.compiler}", f"CC={args.compiler}"]
+        make_command = ["make"]
         if args.skip_modules:
             make_command.append(target)
+        make_command.append("LOCALVERSION=-virtme")
+        if args.compiler:
+            make_command += [f"HOSTCC={args.compiler}", f"CC={args.compiler}"]
         if cross_compile and cross_arch:
             make_command += [f"CROSS_COMPILE={cross_compile}", f"ARCH={cross_arch}"]
         # Propagate additional Makefile variables
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -1106,6 +1106,14 @@ class KernelSource:
             self.virtme_param["remote_cmd"] = (
                 f"--remote-cmd {shlex.quote(args.remote_cmd)}"
             )
+        elif args.envs and (
+            args.console_client is not None
+            or args.ssh_client is not None
+            or args.console is not None
+        ):
+            self.virtme_param["remote_cmd"] = (
+                f"--remote-cmd {shlex.quote(shlex.join(args.envs))}"
+            )
         else:
             self.virtme_param["remote_cmd"] = ""
 
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -617,8 +617,6 @@ ARCH_MAPPING = {
     # adding a new arch? Please also update get_host_arch().
 }
 
-MAKE_COMMAND = "make LOCALVERSION=-virtme"
-
 REMOTE_BUILD_SCRIPT = """#!/bin/bash
 cd ~/.virtme
 git reset --hard __virtme__
@@ -733,33 +731,30 @@ class KernelSource:
     def config(self, args):
         """Perform a make config operation on a kernel source directory."""
         arch = args.arch
-        cmd = "virtme-configkernel --defconfig"
+        cmd = ["virtme-configkernel", "--defconfig"]
         if args.verbose:
-            cmd += " --verbose"
+            cmd.append("--verbose")
         if not args.force and not args.kconfig:
-            cmd += " --no-update"
+            cmd.append("--no-update")
         if arch is not None:
             if arch not in ARCH_MAPPING:
                 arg_fail(f"unsupported architecture: {arch}")
             arch = ARCH_MAPPING[arch]["qemu_name"]
-            cmd += f" --arch {arch}"
+            cmd += ["--arch", arch]
         user_config = str(Path.home()) + "/.config/virtme-ng/kernel.config"
         if os.path.exists(user_config):
-            cmd += f" --custom {user_config}"
+            cmd += ["--custom", user_config]
         if args.config:
             for conf in args.config:
-                cmd += f" --custom {conf}"
+                cmd += ["--custom", conf]
         if args.configitem:
             for citem in args.configitem:
-                cmd += f" --configitem {citem}"
+                cmd += ["--configitem", citem]
         # Propagate additional Makefile variables
-        for var in args.envs:
-            cmd += f" {var} "
+        cmd += args.envs
         if args.verbose:
-            print(f"cmd: {cmd}")
-        check_call_cmd(
-            self._format_cmd(cmd), quiet=not args.verbose, dry_run=args.dry_run
-        )
+            print(f"cmd: {shlex.join(cmd)}")
+        check_call_cmd(cmd, quiet=not args.verbose, dry_run=args.dry_run)
 
     def _make_remote(self, args, make_command):
         check_call_cmd(
@@ -874,23 +869,19 @@ class KernelSource:
             target = "bzImage"
             cross_compile = None
             cross_arch = None
-        make_command = MAKE_COMMAND
+        make_command = ["make", "LOCALVERSION=-virtme"]
         if args.compiler:
-            make_command += f" HOSTCC={args.compiler} CC={args.compiler}"
+            make_command += [f"HOSTCC={args.compiler}", f"CC={args.compiler}"]
         if args.skip_modules:
-            make_command += f" {target}"
+            make_command.append(target)
         if cross_compile and cross_arch:
-            make_command += f" CROSS_COMPILE={cross_compile} ARCH={cross_arch}"
+            make_command += [f"CROSS_COMPILE={cross_compile}", f"ARCH={cross_arch}"]
         # Propagate additional Makefile variables
-        for var in args.envs:
-            make_command += f" {var} "
+        make_command += args.envs
+        make_command += ["-j", self.cpus]
         if args.build_host is None:
             # Build the kernel locally
-            check_call_cmd(
-                self._format_cmd(make_command + " -j" + self.cpus),
-                quiet=not args.verbose,
-                dry_run=args.dry_run,
-            )
+            check_call_cmd(make_command, quiet=not args.verbose, dry_run=args.dry_run)
         else:
             # Build the kernel on a remote build host
             self._make_remote(args, make_command)
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -879,6 +879,8 @@ class KernelSource:
         # Propagate additional Makefile variables
         make_command += args.envs
         make_command += ["-j", self.cpus]
+        if args.verbose:
+            print(f"cmd: {shlex.join(make_command)}")
         if args.build_host is None:
             # Build the kernel locally
             check_call_cmd(make_command, quiet=not args.verbose, dry_run=args.dry_run)
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -702,7 +702,7 @@ class KernelSource:
         return None
 
     def _format_cmd(self, cmd):
-        return list(filter(None, cmd.split(" ")))
+        return shlex.split(cmd)
 
     def _is_dirty_repo(self):
         cmd = "git --no-optional-locks status -uno --porcelain"
@@ -1187,7 +1187,7 @@ class KernelSource:
         append = []
         if args.append is not None:
             for item in args.append:
-                split_items = item.split()
+                split_items = shlex.split(item)
                 for split_item in split_items:
                     append.append("-a " + split_item)
         if args.debug:


virtme_ng/run.py
```diff
@@ -874,23 +872,20 @@ def make(self, args):
             target = "bzImage"
             cross_compile = None
             cross_arch = None
-        make_command = MAKE_COMMAND
+        make_command = MAKE_COMMAND.split()
```
mhartmay, line 875: Please use `shlex.split` or even better, change MAKE_COMMAND to an array, but please do not forget to copy the value before changing.

```
MAKE_COMMAND = [...]

...
make_command = MAKE_COMMAND.copy()
```

mhartmay (commented):
Hmmm `args.envs` is no good name then, is it?

And I'm not sure if this can be used in a different way than you intended. e.g. in the README.md there is an example how you can pass makefile variables to the makefile call `vng O=.virtme/build --verbose`.

So what will happen if you use (I did not test):

`vng echo --ssh-client test`

Update: Ideally, we can get rid of `_format_cmd`.

matttbe:
> Hmmm `args.envs` is no good name then, is it?

Indeed, but it looks like it is the same name in other places, and it was already used as a non "env var" in different places, like to execute a command:

    vng -- uname -r

Anyway, we can change, not sure what would be the better name.

> And I'm not sure if this can be used in a different way than you intended. e.g. in the README.md there is an example how you can pass makefile variables to the makefile call `vng O=.virtme/build --verbose`.

Maybe I should update this example.

> So what will happen if you use (I did not test):
> 
> `vng echo --ssh-client test`

No, you cannot set these envs var in different places (even with `vng echo --ssh-client -- test`)

> Update: Ideally, we can get rid of `_format_cmd` or at least it should make use of `shlex.split(...)` instead of `cmd.split()`.

I didn't know about `shlex.split()`, good idea. If I'm not mistaken, `_format_cmd()` should be safe now, but better be safe than sorry.

virtme_ng/run.py
mhartmay: Yep.

matttbe: thanks, that's what I did: https://github.com/arighi/virtme-ng/compare/14a20c7d06120d24b6d8692e0ece19daeaabc5f8..dc05350469d80bd056dbbcc2134f49b37d4fd9bb

virtme_ng/run.py
```diff
@@ -874,23 +870,20 @@ def make(self, args):
...
             cross_compile = None
             cross_arch = None
-        make_command = MAKE_COMMAND
+        make_command = ["make", "LOCALVERSION=-virtme"]
         if args.compiler:
-            make_command += f" HOSTCC={args.compiler} CC={args.compiler}"
+            make_command += [f"HOSTCC={args.compiler}", f"CC={args.compiler}"]
         if args.skip_modules:
-            make_command += f" {target}"
+            make_command.append(target)
```
mhartmay, line 877: Small nit and not related to your PR, but why is the order as it is? Normally, I would expect, first the target and then the makefile variables.

README.md
```diff
@@ -222,9 +222,15 @@ Examples
 
  - Build the kernel using a separate build directory, and run it, in verbose:
 ```
-   $ mkdir -p .virtme/build
```
mhartmay, line 225: Is the directory creation no longer necessary?

virtme_ng/run.py
```diff
@@ -733,33 +731,31 @@ def checkout(self, args):
...
-                cmd += f" --custom {conf}"
+                cmd += ["--custom", conf]
         if args.configitem:
             for citem in args.configitem:
-                cmd += f" --configitem {citem}"
+                cmd += ["--configitem", citem]
         # Propagate additional Makefile variables
-        for var in args.envs:
-            cmd += f" {var} "
+        if args.envs is not None:
```
mhartmay, line 754: Is None ever possible?

The original code would also fail in case of `args.envs == None`.
```
for var in None:
```
raises an exception:
`TypeError: 'NoneType' object is not iterable`

matttbe: Arf, I changed this in my last version, and forgot to commit. Indeed, the condition is useless, `args.envs` is always set, but can be empty. I will remove this line and another one.

virtme_ng/run.py
```diff
@@ -1186,7 +1189,7 @@ def _get_virtme_append(self, args):
         append = []
         if args.append is not None:
             for item in args.append:
-                split_items = item.split()
+                split_items = shlex.split(item)
                 for split_item in split_items:
                     append.append("-a " + split_item)
```
mhartmay, line 1194: Can you please change the whole function.... it looks very suspicious to me.

```
        append = []
        if args.append is not None:
            for item in args.append:
                split_items = shlex.split(item)
                for split_item in split_items:
                   append += ["-a", split_item]
        if args.debug:
            append + = ["-a", "nokaslr"]
        self.virtme_param["append"] = shlex.join(append)
```

Still not beautiful and untested.

README.md
matttbe: good point, I forgot to mention that in the commit message. I think after a recent change on the `virtme-configkernel` side, it is no longer needed to create this dir. I will update the commit message

virtme_ng/run.py
matttbe: Note sure, but I think it doesn't matter with the Makefile, you can run `make FOO=1 target -j48 BAR=2` I think

virtme_ng/run.py
mhartmay: And might be a good idea to use the long option for better readability (although `--kopt` reads not that good :D).

virtme_ng/run.py
mhartmay: It's easier to read IMO :) I would expect the arguments to be sorted a little bit at least... But yes, both should work.

virtme_ng/run.py
matttbe: Good idea! Will do!

virtme_ng/run.py
matttbe: Sure, I can change that.

mhartmay:
LGTM, although untested.

arighi (approved):
Thanks @matttbe for working at this and sorry for the late review (I've been swamped with other proejcts).

Reviewed & tested, all looks good to me!

matttbe:
> Reviewed & tested, all looks good to me!

There was no hurry, thank you for the review and the testing. There are many different cases around these options, more testing is always good to have 🙂

arighi:
BTW @matttbe @mhartmay are you ok to cut a new release of vng? Do you have any other pending changes? Everything seems to work fine on my side with this applied, I've verified all my usual test cases. :)

mhartmay:
> BTW @matttbe @mhartmay are you ok to cut a new release of vng? Do you have any other pending changes? Everything seems to work fine on my side with this applied, I've verified all my usual test cases. :)

Fine with me :) 

And a quick test has shown, that

```
vng O=.virtme/build CC='ccache gcc'  --build
```
works now.

matttbe:
> BTW @matttbe @mhartmay are you ok to cut a new release of vng? Do you have any other pending changes? Everything seems to work fine on my side with this applied, I've verified all my usual test cases. :)

Fine by me, no other pending changes. Thank you for having checked. Good idea to have a new release now!

matttbe:
@arighi thank you for the new release. I noticed you reverted commit eb30683799c782d3b0362c0a66b4970bcf77f704 ("vng: exec: quote envs args"). By chance, do you have the command that was executed by the CI?

Note that as explained in the commit message of the reverted commit, it was changing the behaviour in some cases because the previous one was requiring workarounds, e.g. adding extra backslashes where it shouldn't. Maybe we can also handle these workarounds.

virtme_ng/run.py
```diff
@@ -909,7 +903,7 @@ def _get_virtme_exec(self, args):
             else:
                 envs.append(var)
         if envs:
-            args.exec = " ".join(envs)
+            args.exec = shlex.join(envs)
         if args.exec is not None:
             self.virtme_param["exec"] = f"--script-sh {shlex.quote(args.exec)}"
```
mhartmay, line 908: @matttbe So with your change in https://github.com/arighi/virtme-ng/pull/288/files#diff-729892ba898a62ab180cb4ca44413dd5a538d12fb842bee5d34520e05bf042bdR908 I guess we can remove the `shlex.quote` here and hopefully this fixes the problem from @arighi .

virtme_ng/run.py
```diff
@@ -1113,6 +1107,14 @@ def _get_virtme_remote_cmd(self, args):
             self.virtme_param["remote_cmd"] = (
                 f"--remote-cmd {shlex.quote(args.remote_cmd)}"
             )
+        elif args.envs and (
+            args.console_client is not None
+            or args.ssh_client is not None
+            or args.console is not None
+        ):
+            self.virtme_param["remote_cmd"] = (
+                f"--remote-cmd {shlex.quote(shlex.join(args.envs))}"
```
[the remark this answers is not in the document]
mhartmay: A quick test has shown:

```
$ python3
>>> from shlex import join
>>> print(join(['echo', '-n', 'Multiple words', ';', 'echo', 'bla']))`
echo -n 'Multiple words' ';' echo bla
$ echo -n 'Multiple words' ';' echo bla
Multiple words ; echo bla% 
```
So IMO it seems like there is no need (and we must not) use `shlex.quote` on the result of `shlex.join(...)`.

virtme_ng/run.py
matttbe: @mhartmay thank you for looking at that.

I think we still need the `shlex.quote()` or simple/double quotes to go with `--script-sh`, no?

    --script-sh "echo bla"

Without `shlex.quote()`, we would execute:

    --script-sh echo bla

no?

My fear is that some people tried to use `\` to escape some quotes, and it no longer works. Or indeed, maybe they are using ";" to split commands.

mhartmay: > @mhartmay thank you for looking at that.
> 
> I think we still need the `shlex.quote()` or simple/double quotes to go with `--script-sh`, no?
> 
> ```
> --script-sh "echo bla"
> ```
> 
> Without `shlex.quote()`, we would execute:
> 
> ```
> --script-sh echo bla
> ```
> 
> no?

If this line ` args.exec = shlex.join(envs)` is used for the assignment, then `shlex.quote` is not needed IMO. But if there is another way how `args.exec` can be set then we have to use `shlex.quote`.

So something like
```
if envs:
    args.exec = shlex.join(envs)
else:
   args.exec = shlex.quote(args.exec)
```

Note: I've not checked how args.exec can be set - so please fix the code above accordingly :)

> 
> My fear is that some people tried to use `\` to escape some quotes, and it no longer works. Or indeed, maybe they are using `;` to split commands, but then not sure how.

If they use `;` to split commands, this should never have been worked... because for this you have to set `shell=True` in `Popen` and co. See [1].

If someone needs to run multiple commands and using pipeing and so on, they have to use (untested)
```
--script-sh "sh -c  'echo first; echo second'"
```

[1] https://docs.python.org/3/library/subprocess.html#subprocess.run

matttbe: > > @mhartmay thank you for looking at that.
> > I think we still need the `shlex.quote()` or simple/double quotes to go with `--script-sh`, no?
> > ```
> > --script-sh "echo bla"
> > ```
> > 
> > Without `shlex.quote()`, we would execute:
> > ```
> > --script-sh echo bla
> > ```
> > 
> > no?
> 
> If this line ` args.exec = shlex.join(envs)` is used for the assignment, then `shlex.quote` is not needed IMO. But if there is another way how `args.exec` can be set then we have to use `shlex.quote`.
> 
> So something like
> 
> ```
> if envs:
>     args.exec = shlex.join(envs)
> else:
>    args.exec = shlex.quote(args.exec)
> ```
> 
> Note: I've not checked how args.exec can be set - so please fix the code above accordingly :)

`args.envs` is an array containing all non consumed parameters, while `args.exec` is a string: e.g.

```
$ vng -r -- echo blah
args.envs == ['echo', 'blah']
$ vng -r --exec "echo blah"
args.exec == echo blah
```

If `args.envs` is set, we convert it as a string, and replace `args.exec`:

```
$ vng -r -- echo blah
args.envs == ['echo', 'blah']
args.exec == echo blah
```

We then need to pass this string as argument to `--script-sh`, and this value needs to be quoted because of the possible spaces in `args.exec`. So we need `shlex.quote()` or simple/double quotes around `args.exec`, no?

So one of these lines, but I guess we prefer the first one:

```
self.virtme_param["exec"] = f"--script-sh {shlex.quote(args.exec)}"
self.virtme_param["exec"] = f"--script-sh '{args.exec}'"
```

> > My fear is that some people tried to use `\` to escape some quotes, and it no longer works. Or indeed, maybe they are using `;` to split commands, but then not sure how.
> 
> If they use `;` to split commands, this should never have been worked... because for this you have to set `shell=True` in `Popen` and co. See [1].

Good point!

> If someone needs to run multiple commands and using pipeing and so on, they have to use (untested)
> 
> ```
> --script-sh "sh -c  'echo first; echo second'"
> ```

Indeed. It might be good to know what command stopped working then :)

arighi:
> @arighi thank you for the new release. I noticed you reverted commit [eb30683](https://github.com/arighi/virtme-ng/commit/eb30683799c782d3b0362c0a66b4970bcf77f704) ("vng: exec: quote envs args"). By chance, do you have the command that was executed by the CI?
> 
> Note that as explained in the commit message of the reverted commit, it was changing the behaviour in some cases because the previous one was requiring workarounds, e.g. adding extra backslashes where it shouldn't. Maybe we can also handle these workarounds.

Hi @matttbe , sorry for reverting without giving too much details, the new release completely broke the sched_ext CI and I had to come up with a quick fix...

The command that broke is this one: 
```
 vng -v -- 'stress-ng -t 31 --aggressive -M -c `nproc` -f `nproc` --affinity 1 --affinity-delay 1 --affinity-pin'
```

I think the problem is with the backticks (which is a terrible syntax and we should use the one supported by stress-ng, passing 0, I'll fix this separately in the sched_ext CI), but in any case we shouldn't break a command that was previously working, so ideally we should re-apply your fix and support also this syntax somehow. I'll do some tests later today.

virtme_ng/run.py
mhartmay: > note that if before (or now, after the revert) we were passing:

You mean version 1.35?

> 
> ```
> $ vng -r -- sh -c 'echo first; echo second'
> ```
> 
> it would do:
> 
> ```
> $ vng -r -- sh -c 'echo first; echo second'
> args.envs == ['sh', '-c', 'echo first; echo second']
> args.exec == sh -c echo first; echo second
> --script-sh 'sh -c echo first; echo second'
> ```
> 
> Instead of:
> 
> ```
> $ vng -r -- sh -c 'echo first; echo second'
> args.envs == ['sh', '-c', 'echo first; echo second']
> args.exec == sh -c 'echo first; echo second'
> --script-sh 'sh -c '"'"'echo first; echo second'"'"''
> ```
> 
> Which fixes stuff.

What? :) I'm a little confused...

> I guess when people wanted to use this command, they added extra quotes:
> 
> ```
> $ vng -r -- sh -c '"echo first; echo second"'
> ```

That looks really confusing to me.

> 
> Which would be a good workaround with `" ".join()`:
> 
> ```
> print(f"--script-sh {shlex.quote(" ".join(['sh', '-c', '"echo first; echo second"']))}")
> --script-sh 'sh -c "echo first; echo second"'
> ```
> 
> But not with `shlex.join()`:
> 
> ```
> print(f"--script-sh {shlex.quote(shlex.join(['sh', '-c', '"echo first; echo second"']))}")
> --script-sh 'sh -c '"'"'"echo first; echo second"'"'"''
> ```
> 
> But `sh -c '"echo first; echo second"'` is not a correct command. So as mentioned in the [commit](https://github.com/arighi/virtme-ng/commit/eb30683799c782d3b0362c0a66b4970bcf77f704) message, there is a behaviour change, but for a good reason. No?

Some quick tests:

```
$ git checkout v1.34
./vng -r -- sh -c 'echo first; uname -r; echo separator; hostname | grep --only-matching me'
first
6.8.0-58-generic
separator
me
$ git checkout v1.35
./vng -r -- sh -c 'echo first; uname -r; echo separator; hostname'

6.8.0-58-generic
separator
me
```
So this is broken in v1.35...

```
$ git checkout v1.34
./vng -r -- 'echo first; uname -r; echo separator; hostname | grep --only-matching me'
first
6.8.0-58-generic
separator
me
```
...but not in v1.34.

```
$ git checkout v1.34
./vng -r -- 'echo first; uname -r; echo separator; hostname | grep --only-matching me'
/run/tmp/.virtme-script: line 1: echo first; uname -r; echo separator; hostname | grep --only-matching me: command not found
```
So this was broken in v1.34...

And something that was and is currently broken....
```
$ ./vng -r --exec 'echo hello' -- uname -r
6.8.0-58-generic
```

So we should either disallow the combination of `--exec` and positional arguments or fix it. I tend to disallow it.

Next interesting scenario:
```
./vng -r --exec 'echo hello' --exec 'echo bli'
bli
````
But I guess, that's as expected.

**Comment by mhartmay** [comment]
@matttbe @arighi 
Can you please test https://github.com/mhartmay/virtme-ng/tree/quote-fixes

**Comment by arighi** [comment]
> @matttbe @arighi Can you please test https://github.com/mhartmay/virtme-ng/tree/quote-fixes
> 
> I need to check some other places, but looks better to me :)

Nope, this one doesn't work:
```
$ vng -vr -- echo a b
usage: virtme-run [-h] [--installed-kernel [VERSION] | --kimg [KIMG] | --kdir KDIR] --mods {none,use,auto} [-a KOPT] [--root ROOT] [--rw] [--gdb]
                  [--graphics [BINARY]] [--verbose] [--net [NET]] [--net-mac-address NET_MAC_ADDRESS] [--balloon] [--sound] [--vmcoreinfo]
                  [--snaps] [--disk NAME=PATH] [--blk-disk NAME=PATH] [--memory MEMORY] [--numa NUMA] [--numa-distance NUMA_DISTANCE]
                  [--cpus CPUS] [--name NAME] [--user USER] [--script-sh SHELL_COMMAND] [--script-exec BINARY] [--arch ARCHITECTURE]
                  [--cross-compile CROSS_COMPILE_PREFIX] [--busybox PATH_TO_BUSYBOX] [--qemu-bin QEMU_BIN] [-q QEMU_OPT] [--qemu-opts ...]
                  [--disable-microvm] [--disable-kvm] [--force-initramfs] [--force-9p] [--dry-run] [--show-command]
                  [--save-initramfs SAVE_INITRAMFS] [--show-boot-console] [--no-virtme-ng-init] [--disable-monitor] [--pwd | --cwd CWD]
                  [--rwdir RWDIR] [--rodir RODIR] [--overlay-rwdir OVERLAY_RWDIR] [--nvgpu NVGPU] [--server [{console,ssh}]]
                  [--client [{console,ssh}]] [--port PORT] [--remote-cmd COMMAND] [--ssh-tcp]
virtme-run: error: unrecognized arguments: a b
```

**Comment by matttbe** [comment]
> > @matttbe @arighi Can you please test https://github.com/mhartmay/virtme-ng/tree/quote-fixes
> > I need to check some other places, but looks better to me :)
> 
> Nope, this one doesn't work:
> 
> ```
> $ vng -vr -- echo a b
> usage: virtme-run [-h] [--installed-kernel [VERSION] | --kimg [KIMG] | --kdir KDIR] --mods {none,use,auto} [-a KOPT] [--root ROOT] [--rw] [--gdb]
>                   [--graphics [BINARY]] [--verbose] [--net [NET]] [--net-mac-address NET_MAC_ADDRESS] [--balloon] [--sound] [--vmcoreinfo]
>                   [--snaps] [--disk NAME=PATH] [--blk-disk NAME=PATH] [--memory MEMORY] [--numa NUMA] [--numa-distance NUMA_DISTANCE]
>                   [--cpus CPUS] [--name NAME] [--user USER] [--script-sh SHELL_COMMAND] [--script-exec BINARY] [--arch ARCHITECTURE]
>                   [--cross-compile CROSS_COMPILE_PREFIX] [--busybox PATH_TO_BUSYBOX] [--qemu-bin QEMU_BIN] [-q QEMU_OPT] [--qemu-opts ...]
>                   [--disable-microvm] [--disable-kvm] [--force-initramfs] [--force-9p] [--dry-run] [--show-command]
>                   [--save-initramfs SAVE_INITRAMFS] [--show-boot-console] [--no-virtme-ng-init] [--disable-monitor] [--pwd | --cwd CWD]
>                   [--rwdir RWDIR] [--rodir RODIR] [--overlay-rwdir OVERLAY_RWDIR] [--nvgpu NVGPU] [--server [{console,ssh}]]
>                   [--client [{console,ssh}]] [--port PORT] [--remote-cmd COMMAND] [--ssh-tcp]
> virtme-run: error: unrecognized arguments: a b
> ```

Indeed, we still need the quotes. Here it is calling:

```
virtme-run (...) --script-sh echo a b
```

(instead of `--script-sh "echo a b"`). That's what I was trying to explain above :)

**Comment by mhartmay** [comment]
Hmm, I've updated my branch:
```
./vng -r --exec 'stress-ng -t 31 --aggressive -M -c `nproc` -f `nproc` --affinity 1 --affinity-delay 1 --affinity-pin'
```
This works, but this does not and IMO, that's the correct behavior.

For the positional arguments syntax you would have to use:
```
./vng -r -- sh -c 'stress-ng -t 31 --aggressive -M -c `nproc` -f `nproc` --affinity 1 --affinity-delay 1 --affinity-pin'
```

**Comment by matttbe** [comment]
> > @arighi thank you for the new release. I noticed you reverted commit [eb30683](https://github.com/arighi/virtme-ng/commit/eb30683799c782d3b0362c0a66b4970bcf77f704) ("vng: exec: quote envs args"). By chance, do you have the command that was executed by the CI?
> > Note that as explained in the commit message of the reverted commit, it was changing the behaviour in some cases because the previous one was requiring workarounds, e.g. adding extra backslashes where it shouldn't. Maybe we can also handle these workarounds.
> 
> Hi @matttbe , sorry for reverting without giving too much details, the new release completely broke the sched_ext CI and I had to come up with a quick fix...

I understand this rushed action due to stress (... `-ng` :) )

> The command that broke is this one:
> 
> ```
>  vng -v -- 'stress-ng -t 31 --aggressive -M -c `nproc` -f `nproc` --affinity 1 --affinity-delay 1 --affinity-pin'
> ```
> 
> I think the problem is with the backticks (which is a terrible syntax and we should use the one supported by stress-ng, passing 0, I'll fix this separately in the sched_ext CI), but in any case we shouldn't break a command that was previously working, so ideally we should re-apply your fix and support also this syntax somehow. I'll do some tests later today.

I guess the issue is the simple quotes, right? `shlex.join()` would keep the quotes around the whole command.

> Hm... since shlex.join() produces a quoted string, maybe we just need to strip the quote. This one seems to work for me:
> 
> ```
> diff --git a/virtme_ng/run.py b/virtme_ng/run.py
> index 8d21e05..bbd466d 100644
> --- a/virtme_ng/run.py
> +++ b/virtme_ng/run.py
> @@ -903,7 +903,7 @@ class KernelSource:
>              else:
>                  envs.append(var)
>          if envs:
> -            args.exec = shlex.join(envs)
> +            args.exec = shlex.join(envs).strip("'")
>          if args.exec is not None:
>              self.virtme_param["exec"] = f"--script-sh {shlex.quote(args.exec)}"
>          else:
> ```

I think that will break other things. What about: if `args.exec` has one item, plus starts and finishes with a simple/double quotes, then we strip them?

I understand we don't want to break stuff, but the behaviour change seems needed, and also fix stuff :)
What I mean is that it seems difficult to fix all workarounds, e.g. `vng -r -- sh -c '"echo first; echo second"'` → 2 times quotes as a workaround to keep quotes... What about communicating that the behaviour has been changed/fixed in the release?

**Comment by matttbe** [comment]
> Hmm, I've updated my branch:

Back to 1.34 with an extra commit to prevent `--exec` + `--`?
> 
> ```
> ./vng -r --exec 'stress-ng -t 31 --aggressive -M -c `nproc` -f `nproc` --affinity 1 --affinity-delay 1 --affinity-pin'
> ```
> 
> This works, but this does not:
> 
> ```
> ./vng -r -- 'stress-ng -t 31 --aggressive -M -c `nproc` -f `nproc` --affinity 1 --affinity-delay 1 --affinity-pin'
> ```
> 
> and IMO, that's the correct behavior.

Indeed. But we can do something to prevent "regression".

> For the positional arguments syntax you would have to use:
> 
> ```
> ./vng -r -- sh -c 'stress-ng -t 31 --aggressive -M -c `nproc` -f `nproc` --affinity 1 --affinity-delay 1 --affinity-pin'
> ```

Or I guess this, no?

```
./vng -r -- stress-ng -t 31 --aggressive -M -c '`nproc`' -f '`nproc`' --affinity 1 --affinity-delay 1 --affinity-pin
```

But another workaround :)

**Comment by mhartmay** [comment]
@matttbe @arighi 
Can you please test my latest branch
https://github.com/mhartmay/virtme-ng/tree/quote-fixes

I've added this workaround...
https://github.com/mhartmay/virtme-ng/commit/87173e664f5520fd74ad2ace0f314ce68e71fe2d

So no, hopefully everything works :D

**Comment by matttbe** [comment]
This should be enough, no?

```
index 46e3511..8550b1e 100644
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -909,7 +909,10 @@ class KernelSource:
             else:
                 envs.append(var)
         if envs:
-            args.exec = " ".join(envs)
+            if len(envs) == 1:
+                args.exec = envs[0]
+            else:
+                args.exec = shlex.join(envs)
         if args.exec is not None:
             self.virtme_param["exec"] = f"--script-sh {shlex.quote(args.exec)}"
         else:
```

**Comment by matttbe** [comment]
> @matttbe @arighi Can you please test my latest branch https://github.com/mhartmay/virtme-ng/tree/quote-fixes
> 
> I've added this workaround... [mhartmay@87173e6](https://github.com/mhartmay/virtme-ng/commit/87173e664f5520fd74ad2ace0f314ce68e71fe2d)
> 
> So now, hopefully everything works :D

hehe, same as mine. Indeed. That seems OK.

When I posted my diff, I was thinking about doing that in one line:

    args.exec = envs[0] if len(envs) == 1 else shlex.join(envs)

**Comment by matttbe** [comment]
> > > @matttbe @arighi Can you please test my latest branch https://github.com/mhartmay/virtme-ng/tree/quote-fixes
> > > I've added this workaround... [mhartmay@87173e6](https://github.com/mhartmay/virtme-ng/commit/87173e664f5520fd74ad2ace0f314ce68e71fe2d)
> > > So now, hopefully everything works :D
> > 
> > 
> > Can you add this (or something similar)?
> > ```
> > diff --git a/virtme_ng/run.py b/virtme_ng/run.py
> > index 6a77467..0e142fb 100644
> > --- a/virtme_ng/run.py
> > +++ b/virtme_ng/run.py
> > @@ -910,6 +910,8 @@ class KernelSource:
> >                  args.exec = envs[0]
> >              else:
> >                  args.exec = shlex.join(envs)
> > +            if not args.exec:
> > +                args.exec = None
> > ```
> 
> I would rather set the default of --exec to `None`, but yep.
> 
> [...snip]

it is set to `None` by default, no? But `args.exec` could be a list with an empty string (`['']`) I think, no?

**Comment by matttbe** [comment]
> > > > > @matttbe @arighi Can you please test my latest branch https://github.com/mhartmay/virtme-ng/tree/quote-fixes
> > > > > I've added this workaround... [mhartmay@87173e6](https://github.com/mhartmay/virtme-ng/commit/87173e664f5520fd74ad2ace0f314ce68e71fe2d)
> > > > > So now, hopefully everything works :D
> > > > 
> > > > 
> > > > Can you add this (or something similar)?
> > > > ```
> > > > diff --git a/virtme_ng/run.py b/virtme_ng/run.py
> > > > index 6a77467..0e142fb 100644
> > > > --- a/virtme_ng/run.py
> > > > +++ b/virtme_ng/run.py
> > > > @@ -910,6 +910,8 @@ class KernelSource:
> > > >                  args.exec = envs[0]
> > > >              else:
> > > >                  args.exec = shlex.join(envs)
> > > > +            if not args.exec:
> > > > +                args.exec = None
> > > > ```
> > > 
> > > 
> > > I would rather set the default of --exec to `None`, but yep.
> > > [...snip]
> > 
> > 
> > it is set to `None` by default, no? But `args.exec` could be a list with an empty string (`['']`) I think, no?
> 
> Yep, you're right with `None`, and no `--exec` is not a list, see
> 
> ```
> ./vng -r --exec 'echo hello' --exec 'echo bli'
> bli
> ```

sorry, I wanted to mention `envs`, not `exec`. The names are too similar :)

**Comment by mhartmay** [comment]
> > > > > > @matttbe @arighi Can you please test my latest branch https://github.com/mhartmay/virtme-ng/tree/quote-fixes
> > > > > > I've added this workaround... [mhartmay@87173e6](https://github.com/mhartmay/virtme-ng/commit/87173e664f5520fd74ad2ace0f314ce68e71fe2d)
> > > > > > So now, hopefully everything works :D
> > > > > 
> > > > > 
> > > > > Can you add this (or something similar)?
> > > > > ```
> > > > > diff --git a/virtme_ng/run.py b/virtme_ng/run.py
> > > > > index 6a77467..0e142fb 100644
> > > > > --- a/virtme_ng/run.py
> > > > > +++ b/virtme_ng/run.py
> > > > > @@ -910,6 +910,8 @@ class KernelSource:
> > > > >                  args.exec = envs[0]
> > > > >              else:
> > > > >                  args.exec = shlex.join(envs)
> > > > > +            if not args.exec:
> > > > > +                args.exec = None
> > > > > ```
> > > > 
> > > > 
> > > > I would rather set the default of --exec to `None`, but yep.
> > > > [...snip]
> > > 
> > > 
> > > it is set to `None` by default, no? But `args.exec` could be a list with an empty string (`['']`) I think, no?
> > 
> > 
> > Yep, you're right with `None`, and no `--exec` is not a list, see
> > ```
> > ./vng -r --exec 'echo hello' --exec 'echo bli'
> > bli
> > ```
> 
> sorry, I wanted to mention `envs`, not `exec`. The names are too similar :)

You mean this:
```
$./vng -r -- ''
qemu-system-s390x: -chardev file,id=console,path=/dev/null: Duplicate ID 'console' for chardev
```?

**Comment by matttbe** [comment]
> > > > > > > @matttbe @arighi Can you please test my latest branch https://github.com/mhartmay/virtme-ng/tree/quote-fixes
> > > > > > > I've added this workaround... [mhartmay@87173e6](https://github.com/mhartmay/virtme-ng/commit/87173e664f5520fd74ad2ace0f314ce68e71fe2d)
> > > > > > > So now, hopefully everything works :D
> > > > > > 
> > > > > > 
> > > > > > Can you add this (or something similar)?
> > > > > > ```
> > > > > > diff --git a/virtme_ng/run.py b/virtme_ng/run.py
> > > > > > index 6a77467..0e142fb 100644
> > > > > > --- a/virtme_ng/run.py
> > > > > > +++ b/virtme_ng/run.py
> > > > > > @@ -910,6 +910,8 @@ class KernelSource:
> > > > > >                  args.exec = envs[0]
> > > > > >              else:
> > > > > >                  args.exec = shlex.join(envs)
> > > > > > +            if not args.exec:
> > > > > > +                args.exec = None
> > > > > > ```
> > > > > 
> > > > > 
> > > > > I would rather set the default of --exec to `None`, but yep.
> > > > > [...snip]
> > > > 
> > > > 
> > > > it is set to `None` by default, no? But `args.exec` could be a list with an empty string (`['']`) I think, no?
> > > 
> > > 
> > > Yep, you're right with `None`, and no `--exec` is not a list, see
> > > ```
> > > ./vng -r --exec 'echo hello' --exec 'echo bli'
> > > bli
> > > ```
> > 
> > 
> > sorry, I wanted to mention `envs`, not `exec`. The names are too similar :)
> 
> You mean this:
> 
> ```
> $./vng -r -- ''
> qemu-system-s390x: -chardev file,id=console,path=/dev/null: Duplicate ID 'console' for chardev
> ```
> 
> ?

Yes, I guess that's what @arighi was showing in https://github.com/arighi/virtme-ng/pull/288#issuecomment-2851005682

The fix could be simply this:

```
index 46e3511..c079e89 100644
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -910,8 +910,8 @@ class KernelSource:
                 envs.append(var)
         if envs:
             args.exec = " ".join(envs)
-        if args.exec is not None:
+        if args.exec:
             self.virtme_param["exec"] = f"--script-sh {shlex.quote(args.exec)}"
         else:
             self.virtme_param["exec"] = ""
```

mhartmay:
@matttbe @arighi Thanks for the feedback, I've created a PR https://github.com/arighi/virtme-ng/pull/291

PR #291, opened 2025-05-05
Quote fixes

mhartmay:


mhartmay:
Do we have the same problems with remote cmd?

matttbe (approved):
Thank you, the code looks good to me, and the quick tests I did before with only the second patch were OK.

Note: for the 2nd commit, you mentioned twice the same command. Maybe you wanted to mention a version without `sh -c "..."`, but only with quotes?

Also, it means that if we try to execute a single command containing a space, we will need to add extra quotes, e.g. 

    vng -r -- "/run/my command"    ## not OK
    vng -r -- "'/run/my command'"  ## OK

but that's the same behaviour as before (except 1.34), and unlikely. So it is fine

matttbe:
> Do we have the same problems with remote cmd?

Do you mean if we use:

    vng --ssh-client -- 'echo a b c `nproc`'

This is a new command, I guess we don't need this workaround, people should use:

    vng --ssh-client -- echo 'a b c `nproc`'

mhartmay:
> Thank you, the code looks good to me, and the quick tests I did before with only the second patch were OK.
> 
> Note: for the 2nd commit, you mentioned twice the same command. Maybe you wanted to mention a version without `sh -c "..."`, but only with quotes?

Yep.

> 
> Also, it means that if we try to execute a single command containing a space, we will need to add extra quotes, e.g.
> 
> ```
> vng -r -- "/run/my command"    ## not OK
> vng -r -- "'/run/my command'"  ## OK
> ```

Good finding.

> 
> but that's the same behaviour as before (except 1.34), and unlikely. So it is fine

Yep.

PR #294, opened 2025-05-06
vng: Fix remote build

arighi:
Remote build command is currently broken:

```
$ vng -vb --build-host remote-host
...
Traceback (most recent call last):
...
    make_command + " -j$(nproc --all)",
    ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
TypeError: can only concatenate list (not "str") to list
```

Fix by treating make_command as a list consistently.

Fixes: 276b378 ("vng: build/quote: quotes envs and args")

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -788,7 +788,7 @@ class KernelSource:
             tmp.write(
                 REMOTE_BUILD_SCRIPT.format(
                     args.build_host_exec_prefix or "",
-                    make_command + " -j$(nproc --all)",
+                    shlex.join(make_command) + " -j$(nproc --all)",
                 )
             )
             tmp.flush()
@@ -835,9 +835,7 @@ class KernelSource:
                     ["fakeroot", "debian/rules", "clean"], quiet=not args.verbose
                 )
             check_call_cmd(
-                self._format_cmd(
-                    make_command + f" -j {self.cpus}" + " modules_prepare"
-                ),
+                make_command + ["-j", self.cpus, "modules_prepare"],
                 quiet=not args.verbose,
                 dry_run=args.dry_run,
             )
@@ -871,19 +869,19 @@ class KernelSource:
             cross_arch = None
         make_command = ["make"]
         if args.skip_modules:
-            make_command.append(target)
-        make_command.append("LOCALVERSION=-virtme")
+            make_command += [target]
+        make_command += ["LOCALVERSION=-virtme"]
         if args.compiler:
             make_command += [f"HOSTCC={args.compiler}", f"CC={args.compiler}"]
         if cross_compile and cross_arch:
             make_command += [f"CROSS_COMPILE={cross_compile}", f"ARCH={cross_arch}"]
         # Propagate additional Makefile variables
         make_command += args.envs
-        make_command += ["-j", self.cpus]
-        if args.verbose:
-            print(f"cmd: {shlex.join(make_command)}")
         if args.build_host is None:
             # Build the kernel locally
+            make_command += ["-j", self.cpus]
+            if args.verbose:
+                print(f"cmd: {shlex.join(make_command)}")
             check_call_cmd(make_command, quiet=not args.verbose, dry_run=args.dry_run)
         else:
             # Build the kernel on a remote build host


virtme_ng/run.py
```diff
@@ -872,18 +870,18 @@ def make(self, args):
...
         if cross_compile and cross_arch:
             make_command += [f"CROSS_COMPILE={cross_compile}", f"ARCH={cross_arch}"]
         # Propagate additional Makefile variables
         make_command += args.envs
-        make_command += ["-j", self.cpus]
-        if args.verbose:
-            print(f"cmd: {shlex.join(make_command)}")
         if args.build_host is None:
             # Build the kernel locally
+            make_command += ["-j", self.cpus]
```
matttbe, line 882: arf, good catch, I missed that

virtme_ng/run.py
```diff
@@ -835,9 +835,7 @@ def _make_remote(self, args, make_command):
                     ["fakeroot", "debian/rules", "clean"], quiet=not args.verbose
                 )
             check_call_cmd(
-                self._format_cmd(
-                    make_command + f" -j {self.cpus}" + " modules_prepare"
-                ),
+                make_command + [f"-j {self.cpus}", "modules_prepare"],
```
matttbe, line 838: maybe better to split the first arg:

    ["-j", self.cpus, "modules_prepare"],

BTW, not related to this commit, but `self.cpus` feels wrong, I guess it should be `-j$(nproc --all)"` like above, to get the CPU cores from the remote host, not the local ones.

(note that we could add `make_command.append("-j$(nproc --all)")` at the beginning of this function not to repeat it here and above.)

virtme_ng/run.py
```diff
@@ -788,7 +788,7 @@ def _make_remote(self, args, make_command):
             tmp.write(
                 REMOTE_BUILD_SCRIPT.format(
                     args.build_host_exec_prefix or "",
-                    make_command + " -j$(nproc --all)",
+                    " ".join(make_command + ["-j$(nproc --all)"]),
```
matttbe, line 791: It feels wrong to use `" ".join()`, which will remove quoted words. Instead, we could use `shlex.join()`, but it will add a single quote around `-j$(nproc --all)` on purpose.

What about:

    shlex.join(make_command) + " -j$(nproc --all)"

matttbe (changes requested):
Thank you for fixing this! Sorry about these issues, it looks like I forgot to validate that part :(

I have some small suggestions if you don't mind

arighi:
> Thank you for fixing this! Sorry about these issues, it looks like I forgot to validate that part :(

No problem, we're also improving the code overall, so it's totally worth it!

virtme_ng/run.py
```diff
@@ -872,18 +870,18 @@ def make(self, args):
         make_command = ["make"]
         if args.skip_modules:
             make_command.append(target)
-        make_command.append("LOCALVERSION=-virtme")
+        make_command += ["LOCALVERSION=-virtme"]
```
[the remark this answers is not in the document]
arighi: The idea was to make the code more consistent and always use += even when adding a single item. And yes, I need to modify also that line above as well. Will fix this.

virtme_ng/run.py
arighi: Actually `modules_prepare` is always executed locally, even when the kernel is built remotely, so we could use `nproc --all`, but `self.cpus` is also correct.

virtme_ng/run.py
matttbe: (Here as well, do you prefer not to change? I'm confused by your :+1: :) )

```suggestion
                    shlex.join(make_command) + " -j$(nproc --all)",
```

arighi: I think I switched to that in the new push. Did I miss something?

matttbe: Strange, it looks like you did (forced) push, but not modifying that.

arighi: I'm dumb, I forgot to squash the changes into the commit... should be good now. :sweat_smile:

arighi:
> Thank you for the update!
> 
> (I just squashed pre-commit fix-up commit into yours before merging).

Perfect, thanks!

PR #300, opened 2025-05-15
virtme/commands/run: Do not fail if nr_open can't be accessed

arighi:
When running in a container with limited privileges, some procfs files may be inaccessible (e.g., /proc/sys/fs/nr_open), triggering permission denied errors.

In such cases, ignore the error and continue, so that we can still get a functional vng session.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1189,9 +1189,10 @@ def do_it() -> int:
     try:
         with open("/proc/sys/fs/nr_open", encoding="utf-8") as file:
             nr_open = file.readline().strip()
-            kernelargs.append(f"nr_open={nr_open}")
-    except FileNotFoundError:
+    except (FileNotFoundError, PermissionError):
         pass
+    else:
+        kernelargs.append(f"nr_open={nr_open}")
 
     # Parse NUMA settings.
     if args.numa:


virtme/commands/run.py
```diff
@@ -1190,7 +1190,7 @@ def do_it() -> int:
         with open("/proc/sys/fs/nr_open", encoding="utf-8") as file:
             nr_open = file.readline().strip()
             kernelargs.append(f"nr_open={nr_open}")
-    except FileNotFoundError:
+    except (FileNotFoundError, PermissionError):
         pass
```
mhartmay, line 1194:
```suggestion
        pass
    else:
        kernelargs.append(f"nr_open={nr_open}")
```

Unrelated, but this would be a little bit cleaner, but doesn't really matter here :)

virtme/commands/run.py
```diff
@@ -1190,7 +1190,7 @@ def do_it() -> int:
         with open("/proc/sys/fs/nr_open", encoding="utf-8") as file:
             nr_open = file.readline().strip()
             kernelargs.append(f"nr_open={nr_open}")
```
mhartmay, line 1192:
```suggestion
```

mhartmay (approved):
LGTM.

arighi:
@mhartmay squashed your suggestion, thanks!

matttbe (approved):
thanks for the fix!

PR #307, opened 2025-06-04
virtme_ng/utils: resolve symlinks for binary paths

kuba-moo:
Fedora 42 ships with busybox as a symlink:

 $ which busybox
 /usr/sbin/busybox
 $ ll /usr/sbin/busybox
 [...] /usr/sbin/busybox -> ../bin/busybox

This leads to false positive:

  virtme-run: a statically linked busybox could not be found, please install busybox-static

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -937,7 +937,9 @@ def is_statically_linked(binary_path):
     try:
         # Run the 'file' command on the binary and check for the string
         # "statically linked"
-        result = subprocess.check_output(["file", binary_path], universal_newlines=True)
+        result = subprocess.check_output(
+            ["file", "-L", binary_path], universal_newlines=True
+        )
         return "statically linked" in result
     except subprocess.CalledProcessError:
         return False


matttbe:
@kuba-moo thank you for reporting this issue and for the fix!

`find_binary` helper is used in a few different places, it is not clear to me to see the impact. Would it not be "safer" to modify `is_statically_linked` helper to make it compatible with symlinks?

```diff
diff --git a/virtme/commands/run.py b/virtme/commands/run.py
index 86d786b..933fd1b 100644
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -937,7 +937,9 @@ def is_statically_linked(binary_path):
     try:
         # Run the 'file' command on the binary and check for the string
         # "statically linked"
-        result = subprocess.check_output(["file", binary_path], universal_newlines=True)
+        result = subprocess.check_output(
+            ["file", "-L", binary_path], universal_newlines=True
+        )
         return "statically linked" in result
     except subprocess.CalledProcessError:
         return False
```

Would this work for you or is using symlinks for busybox causes other issues?

I don't know what other people are thinking about that :)

matttbe:
> Yup, this works too, I started with something along these lines. Not sure which one is cleaner :S

Me neither :)

The CI is [complaining](https://github.com/arighi/virtme-ng/actions/runs/15448840282/job/43559104981) (I already restarted twice), but no error messages... I don't know if it is due to the modification. Then maybe safer to use `file` with `-L` I would say. Do you mind updating the PR or do you prefer if I look at that?

matttbe (approved):
Thank you for the update! The CI looks happier :)

PR #323, opened 2025-08-18
Some smaller changes for virtme-init-ng

mhartmay:


arighi:
> Right, we should add `rust-version = "..."` to `Cargo.toml` then [1].
> 
> [1] https://doc.rust-lang.org/cargo/reference/rust-version.html

I've tried to build virtme-ng-init all the way back to Ubuntu 24.04 (that ships rustc-1.75) and everything seems fine. Do you know which version should we enforce exactly? Otherwise I think it's good as it is, the code should build across most distro and releases.

arighi:
In any case I'm not too worried about not supporting really old rustc versions, as @matttbe mentioned only a few people should build virtme-ng-init and all these people are probably using a recent distro.

mhartmay:
I've created a follow-up PR based on the discussion https://github.com/arighi/virtme-ng/pull/325.

PR #325, opened 2025-08-22
Set MSRV, configure linter and fix cargo clippy findings

mhartmay:


matttbe (approved):
Thank you for the improvements!

It looks good to me, but I would prefer if someone else who is more used to work with Rust than me checks at least the modifications in `Cargo.toml`.

matttbe:
> I was also wondering if we should run the `cargo clippy` command as a separate job in the github CI. WDYT?

Maybe the `pre-commit` framework can help as well?

arighi:
> > I was also wondering if we should run the `cargo clippy` command as a separate job in the github CI. WDYT?
> 
> Maybe the `pre-commit` framework can help as well?

That would be even better. I'll just merge this for now and then we can integrate this in pre-commit as a separate PR.

PR #326, opened 2025-08-25
pre-commit: Add cargo hooks

mhartmay:
Add cargo hooks for `cargo clippy|check|fmt`, but let's not run them on https://pre-commit.ci/ as they are not available there.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,8 +20,19 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Coding style
-      run: cargo fmt -- --check
+    - uses: actions/setup-python@v3
+    - uses: pre-commit/action@v3.0.1
+      with:
+        extra_args: cargo-clippy-fix --all-files
+    - uses: pre-commit/action@v3.0.1
+      with:
+        extra_args: cargo-clippy --all-files
+    - uses: pre-commit/action@v3.0.1
+      with:
+        extra_args: cargo-check --all-files
+    - uses: pre-commit/action@v3.0.1
+      with:
+        extra_args: cargo-fmt --all-files
     - name: Build
       run: cargo build --verbose
     - name: Run tests


.pre-commit-config.yaml
```diff
@@ -1,4 +1,7 @@
 ---
+ci:
+    skip: [cargo-clippy, cargo-check, cargo-fmt]
```
matttbe, line 3: I understand why it cannot work there, but then, is the idea to the [current](https://github.com/arighi/virtme-ng/blob/main/.github/workflows/rust.yml) Rust GH Action to run `pre-commit` (only the Rust related ones?) to be sure the same commands are executed?

BTW: thank you for adding that! :)

.github/workflows/rust.yml
```diff
@@ -20,8 +20,19 @@ jobs:
...
+        extra_args: cargo-clippy-fix --all-files
+    - uses: pre-commit/action@v3.0.1
+      with:
+        extra_args: cargo-clippy --all-files
+    - uses: pre-commit/action@v3.0.1
+      with:
+        extra_args: cargo-check --all-files
+    - uses: pre-commit/action@v3.0.1
+      with:
+        extra_args: cargo-fmt --all-files
```
matttbe, line 35: Can you not run `pre-commit` once with `--files virtme_ng_init/src/*.rs` as extra arguments?

    pre-commit run --files virtme_ng_init/src/*.rs

matttbe: Edit: I forgot we were already in the `virtme_ng_init` dir here, so I guess the arguments can be `--files */*`:

    pre-commit run --files */*

arighi (approved):

.github/workflows/rust.yml
matttbe: Addressed in https://github.com/arighi/virtme-ng/pull/329

Issue #187, opened 2024-10-19
Stale file handle in the guest after changing file on the host


Romeus:
### Prerequisites

Vng is built from source and installed in a virtual environment:

(virtme) hedin@laptop:~/prj$ vng --version
virtme-ng 1.31+5.g1da4504
                        
Host OS version:
hedin@laptop:~$ cat /etc/os-release 
PRETTY_NAME="Ubuntu 24.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.1 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo

### Reproducer

1) Build the latest linux-next and run it:
(virtme) hedin@laptop:~/prj$ vng -r kernel_builds/linux-next-virtme/                                                
          _      _                                                                                                                       
   __   _(_)_ __| |_ _ __ ___   ___       _ __   __ _                                                                                    
   \ \ / / |  __| __|  _   _ \ / _ \_____|  _ \ / _  |                                                                                   
    \ V /| | |  | |_| | | | | |  __/_____| | | | (_| |                                                                                   
     \_/ |_|_|   \__|_| |_| |_|\___|     |_| |_|\__  |                                                                                   
                                                |___/                                                                                    
   kernel version: 6.12.0-rc3-next-20241018-virtme x86_64                                                                                
   (CTRL+d to exit) 

2) Create a file on the host:
hedin@laptop:~$ echo "123" >  test.txt

3) Check it on the guest:
hedin@virtme-ng:~$ cat test.txt                                                                                                          
123

4) Modify it on the host by adding something:
hedin@laptop:~$ vim test.txt

5) Check the changes in the guest:
hedin@virtme-ng:~$ cat test.txt                                                                                                          
cat: test.txt: Stale file handle                                                                                                         

6) Close the editor in the host and check file on the guest one more time:
hedin@virtme-ng:~$ cat test.txt   
cat: test.txt: Stale file handle 

7) List files in the guest:
hedin@virtme-ng:~$ ls
[...]

PR #340, opened 2025-09-17
virtme-ng: Run virtiofsd with -o cache=always to avoid ESTALE errors

arighi:
When files are modified or replaced on the host, the guest may hit "Stale file handle" (ESTALE) errors when accessing them through virtiofs. This happens because virtiofsd revalidates inode and attribute information with the host on every access, and the guest ends up holding references to inodes that no longer exist.

Run virtiofsd with `-o cache=always` to address this. In this mode, inode and attribute metadata are always cached in the guest and never revalidated with the host. This prevents the guest from observing stale handles, since it no longer sees the underlying inode changes.

This effectively masks the ESTALE errors at the cost of weaker coherency: host-side file replacements may not be immediately visible to the guest, but for most use cases this trade-off is acceptable.

Fixes: https://github.com/arighi/virtme-ng/issues/187

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -746,7 +746,7 @@ class VirtioFS:
             stderr = ""
         os.system(
             f"{virtiofsd_path} --syslog --no-announce-submounts "
-            + f"--socket-path {self.sock} --shared-dir {path} --sandbox none {stderr} &"
+            + f"--socket-path {self.sock} --shared-dir {path} --sandbox none {stderr} -o cache=always &"
         )
         max_attempts = 5
         check_duration = 0.1


matttbe (approved):
Thank you! It should indeed be fine like that.

Do you think we need to document the workaround in the README?

Just in case someone has some issues related to that, this person (or robot, trying to be inclusive here!) can look at these links:
- https://gitlab.com/virtio-fs/virtiofsd/-/issues/188
- https://gitlab.com/virtio-fs/virtiofsd/-/issues/194
- (https://gitlab.com/virtio-fs/virtiofsd/-/merge_requests/173)

arighi:
> Thank you! It should indeed be fine like that.
> 
> Do you think we need to document the workaround in the README?

Maybe we can add a note in the troubleshooting section, mentioning the potential out-of-sync data between host and guest. I'll add that in a separate PR.

> 
> Just in case someone has some issues related to that, this person (or robot, trying to be inclusive here!) can look at these links:
> 
> * https://gitlab.com/virtio-fs/virtiofsd/-/issues/188
> * https://gitlab.com/virtio-fs/virtiofsd/-/issues/194
> * (https://gitlab.com/virtio-fs/virtiofsd/-/merge_requests/173)

Awesome! Thanks for adding these links!

PR #347, opened 2025-09-26
virtme/architectures: Enable topology extensions for AMD CPUs

arighi:
Certain AMD CPUs require topology extensions (topoext) to be enabled to expose extra CPU topology information to the virtme-ng guest, such as threads per core and accurate core/socket layout.

Enable it by default both for x86 and microvm architectures. This allows to properly enable SMT inside virtme-ng guests on AMD CPUs.

On non-AMD CPUs, topoext is ignored, therefore it's safe to enable it unconditionally by default.


marcosps (approved):
I don't have an AMD machine, but it works as expected with and without microvm being used.

Issue #346, opened 2025-09-25
Provide a way to define a default shell


jclsn:
I am running into some problems with zsh + oh-my-zsh + plugins etc. So it would be nice if I could tell virtme-ng to use bash by default. This way I could also visually distinguish the sessions at a glance. I had a look at the `virtme-ng.conf`, but couldn't find anything like this.

arighi:
@jclsn how about a `--shell` option to override the default user shell? See https://github.com/arighi/virtme-ng/pull/348.

jclsn:
Looks good! Would this also be configurable via the `virtme-ng.conf`

PR #348, opened 2025-09-26
virtme-ng: Introduce --shell

arighi:
Introduce a `--shell` option to override the default shell from `/etc/passwd` when launching a user session inside the vng guest.

This allows starting the guest session with an alternative shell or program (i.e., tmux).

Example:
```
  $ vng --shell /bin/bash
  ...
  $ echo $SHELL
  /bin/bash

  $ vng --shell /bin/fish
  ...
  > echo $SHELL
  /bin/fish
```

This addresses issue #346.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -486,8 +486,13 @@ if [[ -n ${virtme_graphics} ]]; then
     fi
     # Drop to console if the graphical app failed.
 fi
+
 if [[ -n ${virtme_user} ]]; then
-    setsid bash -c "su -- ${virtme_user}" 0<> "/dev/$consdev" 1>&0 2>&0
+    shell=""
+    if [[ -n ${virtme_shell} ]]; then
+        shell="-s ${virtme_shell}"
+    fi
+    setsid bash -c "su ${shell} -- ${virtme_user}" 0<> "/dev/$consdev" 1>&0 2>&0
 else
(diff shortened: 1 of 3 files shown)


jclsn:
Works like a charm and fixes the issues I had with zsh + oh-my-zsh. The shell now loads much faster and no issues with the cursor anymore.

marcosps (approved):
Works like a charm! Even nicer to start tmux right away from a new vng session. Thanks a lot!

arighi:
> > Works like a charm! Even nicer to start tmux right away from a new vng session. Thanks a lot!
> 
> I did have issues with tmux too. How do you do it?

It works for me just with:
```
$ vng --shell /usr/bin/tmux
```

jclsn:
> > > Works like a charm! Even nicer to start tmux right away from a new vng session. Thanks a lot!
> > 
> > 
> > I did have issues with tmux too. How do you do it?
> 
> It works for me just with:
> 
> ```
> $ vng --shell /usr/bin/tmux
> ```

Yeah, I still have a lot of rendering issues there with zsh. That is why I usually prefer to connect to a vng session via ssh. Works fine with bash though. Anyone else using zsh here?

matttbe:
@arighi thank you for having added this new option!

Do you think this option should also be respected when using `--console/--console-client`?
And even with SSH? In this case, it might be interesting to use `chsh`, no?

PR #350, opened 2025-09-28
virtme: systemd: Disable zram service

arighi:
Kernels built with the default virtme-ng configuration lack zram support, so starting the zram service can hang the systemd boot.

Prevent the hang by implicitly disabling the zram service (system.zram=0).

This fixes a boot hang issue in CachyOS when using --systemd.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1323,6 +1323,8 @@ def do_it() -> int:
         kernelargs.append("luks=no")
         # disable auditd so there are no errors if the user lacks `--rw`
         kernelargs.append("audit=off")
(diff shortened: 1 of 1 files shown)


mhartmay:
> Kernels built with the default virtme-ng configuration lack zram support, so starting the zram service can hang the systemd boot.

Hmm in theory, we could check/probe for that support, no? Do we want to have such a feature in the future?

> 
> Prevent the hang by implicitly disabling the zram service (system.zram=0).
                                                                                            ^^^^^^              ^
                                                                                            generator          d

Small nit: Please add a reference to the `zram-generator.conf` manpage.

Besides that it LGTM. Thanks for fixing this.

PR #371, opened 2025-10-24
virtme/run: Set virtme_root_user=1 also when using `--root` option

rbmarliere:
Currently, the only way to bypass the mount of /root as a tmpfs in the init is by means of $virtme_root_user. Therefore it's only possible to fully utilize the root user within an external filesystem passed through `--root` if the user executes virtme-ng as root.

By making use of the existing variable, it allows users to run virtme-ng targeting directories they have read and write access to, e.g.:

`vng --run <kernel> --root <rootfs> --user root`

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -657,8 +657,6 @@ git reset --hard __virtme__
 
 def create_root(destdir, arch, release):
     """Initialize a rootfs directory, populating files/directory if it doesn't exist."""
-    if os.path.exists(destdir):
-        return
     # Use Ubuntu's cloud images to create a rootfs, these images are fairly
     # small and they provide a nice environment to test kernels.
     if release is None:
@@ -950,7 +948,12 @@ class KernelSource:
 
     def _get_virtme_root(self, args):
         if args.root is not None:
-            create_root(args.root, args.arch or get_host_arch(), args.root_release)
+            if os.path.exists(args.root):
+                if not os.access(args.root, os.R_OK | os.X_OK):
+                    sys.stderr.write(f"Must have read permission on {args.root}\n")
+                    sys.exit(1)
+            else:
+                create_root(args.root, args.arch or get_host_arch(), args.root_release)
             self.virtme_param["root"] = f"--root {args.root}"
         else:
             self.virtme_param["root"] = ""


virtme/commands/run.py
```diff
@@ -1810,10 +1810,10 @@ def get_net_mac(index):
...
         qemuargs.extend(["-device", args.nvgpu])
 
-    # If we are running as root on the host pass this information to the guest
-    # (this can be useful to properly support running virtme-ng instances
-    # inside docker)
-    if os.geteuid() == 0:
+    # If we are running as root on the host, or using root user within an
+    # external root filesystem, pass this information to the guest (this can be
+    # useful to properly support running virtme-ng instances inside docker)
+    if os.geteuid() == 0 or (args.user == "root" and args.root != "/"):
```
arighi: I think we should also verify that the user has read/writ/exec access to `/root` within the external root dir, something like the following (untested):
```
    if (
        os.geteuid() == 0
        or (
            args.user == "root"
            and args.root != "/"
            and os.access(os.path.join(args.root, "root"), os.R_OK | os.W_OK | os.X_OK)
        )
    ):
         kernelargs.append("virtme_root_user=1")
```

WDYT?

rbmarliere: Well, that wouldn't hurt but I don't think it's necessary IMHO. The thing is, if we were to check if the user has permissions for `$ROOT/root`, we might aswell check `$ROOT`, right? I haven't tested also but I assume the errors would be pretty obvious.

arighi: The thing is that if we don't do the check, we'd end up with an environment where root can't to access `$HOME`, which sounds a bit broken... and yes, as you pointed out, I think we should also check `args.root` as well, like:
```
if (
    os.geteuid() == 0
    or (
        args.user == "root"
        and args.root != "/"
        and os.access(args.root, os.R_OK | os.X_OK)
        and os.access(os.path.join(args.root, "root"), os.R_OK | os.W_OK | os.X_OK)
    )
):
    kernelargs.append("virtme_root_user=1")
```

rbmarliere: I agree there should be a check, but maybe somewhere else? What do you think of the following (still untested)?

```
index b3846d3914d6..1d7e102da3f2 100644
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -657,8 +657,6 @@ git reset --hard __virtme__
 
 def create_root(destdir, arch, release):
     """Initialize a rootfs directory, populating files/directory if it doesn't exist."""
-    if os.path.exists(destdir):
-        return
     # Use Ubuntu's cloud images to create a rootfs, these images are fairly
     # small and they provide a nice environment to test kernels.
     if release is None:
@@ -950,7 +948,13 @@ class KernelSource:
 
     def _get_virtme_root(self, args):
         if args.root is not None:
-            create_root(args.root, args.arch or get_host_arch(), args.root_release)
+            if os.path.exists(args.root):
+                if (not os.access(args.root, os.R_OK | os.X_OK)
+                    or not os.access(os.path.join(args.root, "root"), os.R_OK | os.W_OK | os.X_OK)):
+                    sys.stderr.write(f"Must have rx permissions on {args.root} and rwx on {args.root}/root")
+                    sys.exit(1)
+            else:
+                create_root(args.root, args.arch or get_host_arch(), args.root_release)
             self.virtme_param["root"] = f"--root {args.root}"
         else:
             self.virtme_param["root"] = ""
```

arighi (approved):
Looks good, thanks!

PR #373, opened 2025-10-28
External `--root` filesystem fixes for `--ssh`

rbmarliere:
Hello! Here are a few fixes to improve the usage of `--root` when used along with `--ssh`. Basically, instead of having /etc/ssh files in each root filesystem, let's mount `$HOME/.cache/virtme-ng/.ssh` as virtme.cache in the guest and use it as a central place for the files. This way, the user can leave his `id_*.pub` files for authentication there as well.

### DIFF
--- a/virtme/guest/virtme-sshd-script
+++ b/virtme/guest/virtme-sshd-script
@@ -70,6 +70,7 @@ AuthorizedKeysFile ${SSH_AUTH_KEYS}
 PubkeyAuthentication yes
 UsePAM yes
 PrintMotd no
+StrictModes no
 ${sftp_server}
 EOF
 


arighi:
I still need to investigate in details, but this seems to break `vng --ssh-client` for me.

Reproducer:
```
$ vng --sshd
...
$ vng --ssh-client
Connection closed by UNKNOWN port 0
```

With vng from the main branch everything works. Is this working for you?

arighi:
Thanks! I'll repeat my tests in a bit.

virtme/guest/virtme-sshd-script
```diff
@@ -17,14 +17,22 @@ rm -f /var/run/nologin
...
 # use an ssh location INSIDE of the users home directory to communicate pre-
 # created hostkeys and authorized keys
 SSH_HOME="$(getent passwd "${virtme_ssh_user}" | cut -d: -f6)"
-if [ ! -e "${SSH_HOME}" ]; then
+if [ -e "${SSH_HOME}" ]; then
+    SSH_CACHE="${SSH_HOME}"/.cache/virtme-ng/.ssh
+else
     # Setup an arbitrary ssh location, just to be able to start sshd.
-    SSH_HOME="${SSH_DIR}"
+    SSH_HOME="${SSH_CACHE}"
```
matttbe: could it be possible here not to have `SSH_CACHE` defined?

Would it not be safer to have:

```bash
if [ -e "${SSH_HOME}" ]; then
    SSH_CACHE="${SSH_HOME}"/.cache/virtme-ng/.ssh
elif [ -n "${SSH_CACHE}" ]; then
    # Setup an arbitrary ssh location, just to be able to start sshd.
    SSH_HOME="${SSH_CACHE}"
else
   SSH_HOME="${SSH_DIR}"
   SSH_CACHE="${SSH_HOME}"/.cache/virtme-ng/.ssh
fi
```

(or in a different order to avoid setting `SSH_CACHE` twice the same way)

virtme/guest/virtme-sshd-script
```diff
@@ -70,6 +70,7 @@ AuthorizedKeysFile ${SSH_AUTH_KEYS}
 PubkeyAuthentication yes
 UsePAM yes
 PrintMotd no
+StrictModes no
```
matttbe: because virtme is used for the development, and should be only available locally, I guess that's fine

matttbe (changes requested):
Thank you for adding `--ssh` support with `--root`!

I didn't test yet (but I saw Andrea was going to do that), but I have 2 small questions/suggestions if you don't mind.

virtme/guest/virtme-sshd-script
rbmarliere: Actually, I'm thinking of skipping this for `args.root != '/'`:
```
SSH_HOME="$(getent passwd "${virtme_ssh_user}" | cut -d: -f6)"
```
Because at this point if the same user from the host exist within the `--root` filesystem, it would use the `/.../root/home/$virtme_ssh_user/.cache/virtme-ng/.ssh` as cache... I will see how this part can be improved/simplified.

PR #378, opened 2025-11-24
Bash vs rust implementation

mhartmay:
Let’s work on making the implementations more consistent with each other.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -457,7 +457,6 @@ run_user_script() {
         echo ${ret} > /dev/virtio-ports/virtme.ret
     fi
 
-    # Hmm.  We should expose the return value somehow.
     poweroff_and_exit 0
 }
 


arighi (approved):
Left a comment, but overall LGTM.

mhartmay:
@arighi 
Sorry, I've reworked the whole series...

mhartmay:
My idea is that, since the structures and function names are now very similar, it becomes easier to keep the Bash variant and the Rust implementation in sync.

As you can see there are still some TODOs in the bash file where I had open questions or did not find the Rust part for it.

mhartmay:
Thanks for the feedback!

Hmm, I don’t have a strong opinion on that, and I’m fine with moving those changes into a separate PR.

matttbe:
> Hmm, I don’t have a strong opinion on that, and I’m fine with moving those changes into a separate PR.

I'm not sure whether that's needed, let see what @arighi thinks about that.

arighi (changes requested):
I agree with @matttbe about the commit "virtme-init: Performance improvements", no reason to be too aggressive with the bash script, it's actually better to prioritize stability and predictability in virtme-init (sometimes it can used as a "backup" init to verify potential bugs in virtme-ng-init, so it needs to be reliable).

Everything else looks good to me and can be merged. Thanks!

virtme/guest/virtme-init
```diff
@@ -4,8 +4,11 @@
 # Licensed under the GPLv2, which is available in the virtme distribution
 # as a file called LICENSE with SHA-256 hash:
 # 8177f97513213526df2cf6184d8ff986c675afb514d4e68a404010521b880643
+set -u
```
arighi: I know it makes debugging easier for us, but it can also make the script more "faulty", I'd drop this.

mhartmay: Ok.

mhartmay:
@arighi I've pushed a patch https://github.com/arighi/virtme-ng/pull/378/changes/157b9a6c1cd0dd96e86e21c636bf30a693485162 that shows the original ordering (more or less).

I find the structure of the Rust implementation more logical, which is why I decided to stick with it.

arighi:
> @arighi I've pushed a patch [157b9a6](https://github.com/arighi/virtme-ng/commit/157b9a6c1cd0dd96e86e21c636bf30a693485162) that shows the original ordering (more or less).
> 
> I find the structure of the Rust implementation more logical, which is why I decided to stick with it.

Yeah, I agree. Let's be consistent with the Rust ordering then. Once you remove the demo patch I think we can merge this PR. If there are issues we can fix them in tree, I think it's a nice cleanup overall. Thanks.

arighi:
@mhartmay looks like CI isn't happy https://github.com/arighi/virtme-ng/actions/runs/20272954733/job/58213586603, but I can't reproduce the failure locally. I'm going to re-run the failed test.

arighi:
Oh well, restarted and it's happy now. Let's see if it happens again.

PR #379, opened 2025-11-24
virtme-ng: Add MCP server for AI agent integration

arighi:
Provide a Model Context Protocol (MCP) server to integrate virtme-ng with AI assistants (i.e., Claude, Cursor, etc.). The server exposes basic kernel development and testing operations via as MCP tools, allowing AI agents to automate virtme-ng workflows.

With this in place, AI assistants can automatically execute kernel development workflows using virtme-ng, allowing natural-language requests to trigger operations like configure the kernel, apply a patch from lore.kernel.org, run a command inside the kernel, etc.

The MCP server is implemented as a standalone script (vng-mcp) to avoid adding extra dependencies to the standard vng tool. The vng command simply invokes vng-mcp when vng --mcp is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -602,6 +602,72 @@ Now you can simply run `vng --build` to build your kernel from the current
 working directory using the external build host, prepending the exec prefix
 command when running make.
 
+AI Agent Integration (MCP Server)
+==================================
+
+virtme-ng includes built-in support for the Model Context Protocol (MCP),
+which allows AI assistants (like Claude, Cursor, or other MCP-compatible
+tools) to interact with virtme-ng programmatically.
+
+What is MCP?
+------------
+
+The Model Context Protocol is an open protocol that enables AI assistants
+to securely access tools and data sources. The virtme-ng MCP server exposes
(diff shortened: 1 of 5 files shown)


marcosps:
> After a few days of use, this has become a part of my regular workflow and it seems to work pretty well. It's also pretty self‑contained, so it shouldn't disrupt anything else and, being a separate script, it also doesn't introduce any mandatory dependencies, so the risk of regression is low.
> 
> Therefore, I think I'm going to merge this. If there's any concerns or issues, we can probably address them in-tree later.

Sorry to now reviewing it, but I don't even know how MCP works... maybe I need to refresh my skills :P

arighi:
@marcosps basically if you use Claude, Cursor, or similar you can ask the AI to test a kernel and, with this thing, the AI is able to run virtme-ng commands. For example you can ask something like "recompile this kernel and test if everything is fine by checking for errors/warnings in dmesg", or "automatically bisect between A and B using this particular test" and the AI does the bisect for you. It's pretty cool. :)

PR #381, opened 2025-11-26
   vng: Add `--empty-passwords` support

mhartmay:
Add a `--empty-passwords` flag that sets all user passwords to empty.
    This is useful for simplifying debugging and enabling easier SSH access,
    e.g. if no SSH public key is available. In case of TCP-based SSH reject
    that with the following error message out of security reasons:

      $ vng -r --empty-passwords --ssh --ssh-tcp
      --empty-passwords can only be used in combination with SSH over vsock

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -319,9 +319,13 @@ def make_parser() -> argparse.ArgumentParser:
         "--disable-monitor", action="store_true", help="Disable QEMU STDIO monitor"
     )
 
-    g = parser.add_argument_group(
-        title="Guest userspace configuration"
-    ).add_mutually_exclusive_group()
+    g = parser.add_argument_group(title="Guest userspace configuration")
+    g.add_argument(
+        "--empty-passwords",
+        action="store_true",
+        help="Set all user passwords to empty",
+    )
+    g = g.add_mutually_exclusive_group()
     g.add_argument(
         "--pwd",
         action="store_true",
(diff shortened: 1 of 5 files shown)


arighi (commented):
Hi @mhartmay should we include just f93799474a3ba8d6075c0e8eda27b601e65e07c2 in this PR? I think the other changes can be merged in a separate PR, right?

mhartmay:
The code itself should be in good shape; the main question is how useful this feature is and what is the best name for the option.

virtme/commands/run.py
```diff
@@ -319,9 +319,13 @@ def make_parser() -> argparse.ArgumentParser:
...
     )
 
-    g = parser.add_argument_group(
-        title="Guest userspace configuration"
-    ).add_mutually_exclusive_group()
+    g = parser.add_argument_group(title="Guest userspace configuration")
+    g.add_argument(
+        "--empty-passwords",
+        action="store_true",
+        help="Set all user passwords to the empty",
```
arighi: nit: s/the empty/empty/

arighi (approved):
Small nit, but overall LGTM, it could be useful in some cases.

mhartmay:
@arighi 
Alternative option names:
`--all-passwords-empty`
`--force-empty-passwords`
`--force-all-passwords-empty`
`--wipe-passwords`

vs.

`--empty-passwords`

What's you preferred naming?

virtme/commands/run.py
mhartmay: Changed.

PR #382, opened 2025-11-28
Improve `vng-mcp` integration

mhartmay:
With this change it's possible to install `virtme-ng` with the extra requirements for MCP. See [1] for details.

 $ cd virtme-ng
 $ pipx install '.[mcp]'

Before:
  $ vng --mcp
  Error: mcp package not found. Install it with: pip install mcp

After:
  $ vng --mcp

[1] https://setuptools.pypa.io/en/latest/userguide/dependency_management.html


mhartmay:
Question: Why do we have both setup.py and pyproject.toml?

PR #383, opened 2025-11-30
virtme-ng: MCP: Document kernel build using virtme-ng

arighi:
Instruct AI agent to prefer using `vng --build` over `make`. The "build kernel" functionality is not implemented as an MCP command, because it may take a long time to complete and this may trigger timeouts with certain AI agents.

Documenting its usage enables AI agents to leverage all vng features including remote host building, automatic kernel configuration, etc., without hitting a timeout, since `vng --build` is treated like any other standard shell command.


arighi:
Merging right away, since it's just documentation / hints for AI agents.

PR #388, opened 2025-12-06
Add support for `--fb` option (like `--graphics` but without X)

intelfx:
Add support for a `--fb` option that starts QEMU in a graphical mode but without actually launching any graphical environment. This is intended for working with the framebuffer console (or anything else that might be run on top of such).

Additionally, this includes one change that synergize with the one above:
- the serial console is connected to stdio also when (any) graphics are used

This decouples the ability to inspect the kernel boot log or other console messages from the fact that the video output is (not) used. Previously, if a video output was used, the serial console on stdio was fully inhibited.

### DIFF
--- a/virtme/commands/configkernel.py
+++ b/virtme/commands/configkernel.py
@@ -234,6 +234,8 @@ _GENERIC_CONFIG_OPTIONAL = [
     "CONFIG_DRM_VIRTIO_GPU_KMS=y",
     "CONFIG_DRM_BOCHS=y",
     "CONFIG_VIRTIO_IOMMU=y",
+    "###: Graphics -> fbdev support (kernel console, vng --fb)",
+    "CONFIG_DRM_FBDEV_EMULATION=y",
     "##: Sound support",
     "CONFIG_SOUND=y",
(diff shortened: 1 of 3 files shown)


arighi (commented):
This looks interesting, do you have any special use case / example on how to use it? Also CI isn't happy, but should be an easy fix. Thanks!

matttbe (changes requested):
(Explicitly marking this PR as "Request Changes", because there are suggested changes.)

matttbe (changes requested):
Thank you for the update!

One small comment for me. The rest looks OK, but I will let @arighi looking at it as he reviewed the first version.

matttbe (approved):
Thank you for the update!

As previously mentioned, I will let Andrea looking at it as he reviewed the first version.

matttbe:
@arighi: when you have the opportunity, do you mind checking this one, please? The modifications look good to me, but I didn't merge it because you commented a previous version.

arighi (approved):
Sorry for the late review. Everything looks good!

Side note: it also works together with `--systemd` which is pretty cool. Good job!

PR #389, opened 2025-12-06
Add support for `vng --blk-disk` and appending disk I/O / topology options to either `--*disk`

intelfx:
Expand `vng`'s capabilities for configuring emulated disk drives. This has two meaningful parts.

First, this PR adds support to the `vng` wrapper for connecting both types of emulated disks to the VM. Originally, `vng` only accepted a `--disk` argument (which was internally translated to `vng --blk-disk`, despite the documentation), with no option to attach a SCSI disk using the `vng` wrapper (while at the same time, `virtme-ng --disk` had an opposite meaning).

This PR adds support for both `vng --scsi-disk` and `vng --blk-disk`, which now map 1:1 to identical `virtme-ng` options, and converts `vng --disk` / `virtme-ng --disk` into a legacy alias (with a warning on usage) due to its inconsistent behavior.

Second, this PR adds support to both `vng` and `virtme-run` for passing extended options in both `--scsi-disk` and `--blk-disk` arguments. Each value is now treated as a comma-separated list with a number of I/O driver options and disk topology options available.

There are several I/O driver options:

- `format=(raw|qcow2)` (with `raw` now explicitly passed to QEMU by default, since QEMU considers implicit format detection a deprecated feature and prohibits write operations to sector 0 if it is used)
- `cache=(...)` corresponds to QEMU's `-blockdev cache.*`, or `-drive cache=...`
- `aio=(...)` corresponds to QEMU's `-blockdev aio=...`, or `-drive aio=...`
- `discard=(on|off)` corresponds to QEMU's `-blockdev discard=(ignore|unmap)` or `-drive discard=...`
- `detect-zeroes=(on|off)` corresponds to QEMU's `-blockdev detect-zeroes=(on|off|unmap)` or `-drive detect-zeroes=...`
- `queues=(int)` corresponds to QEMU's `-device num_queues=...` on `virtio-blk-device` or `virtio-scsi-device` nodes

Then, there are topology options which change the characteristics and limits of the disk reported to the guest OS:

- `log-sec=(bytes)` sets the logical sector (LBA) size, e.g. to emulate 4Kn disks
- `phy-sec=(bytes)` sets the physical (underlying) sector size
- `min-io=(bytes)` sets the advisory minimum I/O size
- `opt-io=(bytes)` sets the advisory optimal I/O size
- `disc-gran=(bytes)` sets the advisory TRIM/UNMAP request granularity
- `rota=(on|off)` indicates whether the emulated drive is reported as rotational medium

Finally, there are two metaoptions:
- `iothread=(on|off)` adds a dedicated I/O thread for the disk and/or SCSI controller, and
- `topology=(on|off)` extracts host device topology data from sysfs and translates that into the values of the options above (naturally, this only works when the host path points to an actual block device).

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -20,6 +20,7 @@ import sys
 import tempfile
 import termios
 from base64 import b64encode
+from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
 from shutil import which
@@ -38,6 +39,7 @@ from virtme_ng.utils import (
     VIRTME_SSH_HOSTNAME_CID_SEPARATORS,
     get_conf,
     scsi_device_id,
+    strtobool,
 )
 
 from .. import architectures, mkinitramfs, modfinder, qemu_helpers, resources, virtmods
@@ -206,15 +208,21 @@ def make_parser() -> argparse.ArgumentParser:
         "--scsi-disk",
         action="append",
         default=[],
-        metavar="NAME=PATH",
-        help="Add a read/write virtio-scsi disk.  The device node will be /dev/disk/by-id/scsi-0virtme_disk_NAME.",
+        metavar="NAME=PATH[,OPT...]",
+        help="Add a read/write virtio-scsi disk, available at "
+        + "/dev/disk/by-id/scsi-0virtme_disk_NAME. "
+        + "OPTs are comma-separated KEY[=VALUE] pairs; "
+        + "pass 'help' to list them.",
     )
     g.add_argument(
         "--blk-disk",
         action="append",
         default=[],
-        metavar="NAME=PATH",
-        help="Add a read/write virtio-blk disk.  The device nodes will be /dev/disk/by-id/virtio-virtme_disk_blk_NAME.",
+        metavar="NAME=PATH[,OPT...]",
+        help="Add a read/write virtio-blk disk, available at "
+        + "/dev/disk/by-id/virtio-virtme_disk_blk_NAME. "
+        + "OPTs are comma-separated KEY[=VALUE] pairs; "
+        + "pass 'help' to list them.",
     )
     g.add_argument(
         "--memory",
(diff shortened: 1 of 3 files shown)


marcosps (approved):
Seems good to me. (I saw that this commit is also present on your patchset to add blk-disk, but this one is easier merged here :) )

arighi (commented):
Looks good, thanks for working on this!!

Just one comment, it'd be nice to not change the behavior of --disk (BTW, with the new syntax how do we specify virtio-blk disks?).

Maybe adding a separate --scsi-disk would be nicer, because we don't break the current behavior, someone may currently use --disk expecting to create virtio-blk devices.

arighi:
> > BTW, with the new syntax how do we specify virtio-blk disks?
> 
> Like I wrote, I made `vng` options map 1:1 to `virtme-run` options, while the semantics of the latter is not changed:
> 
> * `{vng,virtme-run} --disk /path/to/image,foo=bar` → virtio-scsi
> * `{vng,virtme-run} --blk-disk /path/to/image,foo=bar` → virtio-blk

Ah alright, it makes sense as it is then, thanks for clarifying it. Also I think not many people are using --disk, so let's stick with this syntax, which seems more sane.

Edit: I should read the whole thread before replying. This https://github.com/arighi/virtme-ng/pull/389#issuecomment-3656577197 would be the best, thanks!

matttbe (changes requested):
(Explicitly marking this PR as "Request Changes", because it is currently marked as "Approve" while the CI is not happy and there are suggested changes.)

matttbe (changes requested):
thank you for the new version. A few suggestions, but my main issue is with the last commit: can it be simpler?

virtme/commands/run.py
```diff
@@ -203,15 +205,21 @@ def make_parser() -> argparse.ArgumentParser:
         "--scsi-disk",
         action="append",
         default=[],
-        metavar="NAME=PATH",
-        help="Add a read/write virtio-scsi disk.  The device node will be /dev/disk/by-id/scsi-0virtme_disk_NAME.",
+        metavar="NAME=PATH[,OPT...]",
+        help="Add a read/write virtio-scsi disk, available at "
+        + "/dev/disk/by-id/scsi-0virtme_disk_NAME. "
+        + "OPTs are comma-separated KEY[=VALUE] pairs; "
+        + "pass 'help' to list them.",
```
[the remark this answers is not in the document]
intelfx: Regarding "just" referring people to QEMU's help and passing these options 1:1, it took me a ≈full day of digging through QEMU help, mailing lists, and QEMU and Linux kernel source code to come up with a good mental model of how these parameters work and how they have to be passed. The point here was to save everyone else the effort.

(not to mention, again, that these parameters have to be arranged across multiple devices in the QEMU emulated device model, so it is technically not possible to blindly pass them through 1:1)

matttbe:
Hello,

I just checked again this PR, and, even if it is quite a lot of code and mapping, I think it can indeed help to better support disk images. Because I don't know well this area, I still have a few opened questions:

- The "simpler" options are good to quickly find the right command, but is it supporting "enough" cases? What I mean is: because I don't know well this area, I don't know if what is covered here is enough to help most people, or will they quickly need to directly set the QEMU options? I guess it is covering enough, but just in case it is not, maybe documenting the cases supported by this PR in the README would be easier, and people can pass more "raw" options (without the mapping) to the `--*-disk` parameters?
- In the README, there is an example where the `qcow2` format is used, which seems to be the default format then. Here, we are now explicitly using the `raw` format by default → are we not changing the default behaviour? When I try without the PR with a `raw` format (hopefully I generated it correctly), QEMU doesn't want to load it. After a `qemu-img convert` to qcow2, that's fine.
- It might be good to add a few examples in the README, but I'm not sure which ones. Any ideas?

It would be good if someone else can have a look at this PR :)

@arighi @marcosps : you commented this PR earlier, WDYT about it?

virtme/commands/run.py
```diff
@@ -1779,14 +1916,64 @@ def do_it() -> int:
...
+            # log_sec = disk.get_opt("log-sec", parser=int, default=512)
+
+            drive_opts.extend(
+                [
+                    disk.pop_opt_qemu("format", "raw"),
+                    disk.pop_opt_qemu("cache", None),
+                    disk.pop_opt_qemu("aio", None),
+                    f"discard={'unmap' if discard else 'ignore'}"
+                    if discard is not None
+                    else None,
```
marcosps, line 1983: This seems too complicated, maybe they could be simplified? Same applies top below. This is not a nack for the patch, but too many nested `if else` clauses can get confusing quite easy.

marcosps:
It has been a couple years already since I used the `--disk` to run fstests with scratch devices, but the patches look good. Things get messy in some places of the code, like where I commented due to the nested if conditions, but this is not a blocker. I'll try to test it today to make sure that it works as expected, but feel free to merge if you like it.

marcosps:
in my tests, if you add two `scsi-disk` without a proper name, `/dev/disk/by-id` lists two different entries:

```
$ vng -r --user root --scsi-disk ./disk.img,format=raw --scsi-disk ./disk2.img -- ls -R /dev/disk/by-id
/dev/disk/by-id:
scsi-0virtme_disk_disk.img
scsi-0virtme_disk_disk2.img
```

But for `blk-disk`, it shows only one entry, which is a dir that contains both disks:

```
$ vng -r --user root --blk-disk ./disk.img,format=raw --blk-disk ./disk2.img
virtme-ng:/home/mpdesouza # ls /dev/disk/by-id/
virtio-.
virtme-ng:/home/mpdesouza # ls /dev/disk/by-id/virtio-./
disk.img  disk2.img
```

But if I add a name to one of the disks, it then has one dir, for the unnamed disk, and one symlink, as expected:

```
$ vng -r --user root --blk-disk bliu=./disk.img,format=raw --blk-disk ./disk2.img
virtme-ng:/home/mpdesouza # ls /dev/disk/by-id/
virtio-.  virtio-bliu
virtme-ng:/home/mpdesouza # ls /dev/disk/by-id/virtio-./
disk2.img
virtme-ng:/home/mpdesouza # ls /dev/disk/by-id/virtio-bliu 
/dev/disk/by-id/virtio-bliu
```

Is it expected?

virtme/commands/run.py
```diff
@@ -1745,36 +1905,201 @@ def do_it() -> int:
...
+                    # disk.pop_qemu("wzeroes", dest="max-write-zeroes-sectors", parser=lambda arg: int(arg) / log_sec),
+                    disk.pop_opt_qemu("min-io", dest="min_io_size"),
+                    disk.pop_opt_qemu("opt-io", dest="opt_io_size"),
+                    disk.pop_opt_qemu("queues", dest="num-queues"),
+                ]
+            )
+            # unused
+            disk.opts.pop("rota", None)
+
+            if disk.pop_opt("iothread", bool, False):
```
arighi, line 2008: Should use `strtobool` here instead of `bool`?

virtme/commands/run.py
```diff
@@ -1745,36 +1905,201 @@ def do_it() -> int:
...
+                        "rota",
+                        dest="rotation_rate",
+                        parser=lambda arg: "0" if strtobool(arg) else "1",
                     ),
                 ]
             )
 
-    if args.disk:
-        qemuargs.extend(["-device", "{},id=scsi".format(arch.virtio_dev_type("scsi"))])
+            if disk.pop_opt("iothread", bool, False):
```
arighi, line 2117: Ditto, `s/bool/strtobool/`.

virtme/commands/run.py
```diff
@@ -1745,36 +1905,201 @@ def do_it() -> int:
...
-                        arch.virtio_dev_type("blk"), driveid, name
+                    ",".join(o for o in device_opts if o is not None),
+                ]
+            )
+
+            # any options that were not consumed are errors
+            if disk.opts:
+                raise ValueError(
+                    f"invalid --disk parameter: {d!r}\n(keys were not consumed: {disk.opts.keys()})"
+                )
```
arighi, line 2032: We should probably use `arg_fail()` here instead of raise. Something like this?
```
arg_fail(f"unrecognized --blk-disk options for {d!r}: {', '.join(sorted(disk.opts))}")
```

virtme/commands/run.py
```diff
@@ -1745,36 +1905,201 @@ def do_it() -> int:
...
-                    f"scsi-hd,drive={driveid},vendor=virtme,product=disk,serial={name}",
+                    ",".join(o for o in device_opts if o is not None),
                 ]
             )
 
+            # any options that were not consumed are errors
+            if disk.opts:
+                raise ValueError(
+                    f"invalid --disk parameter: {d!r}\n(keys were not consumed: {disk.opts.keys()})"
+                )
```
arighi, line 2143: Ditto, arg_fail() instead of raise, like:
```
arg_fail(f"unrecognized --scsi-disk options for {d!r}: {', '.join(sorted(disk.opts))}")
```

virtme/commands/run.py
```diff
@@ -958,18 +976,153 @@ def create_guest_init_script(root: str, initsh: str) -> str | None:
...
+        # return DiskArg.qemu_opt(name=qemu if qemu is not None else name, value=opt)
+        if opt is None:
+            return None
+        if isinstance(opt, bool):
+            opt = "on" if opt else "off"
+        return f"{dest if dest is not None else name}={opt}"
+
+    def topology(self) -> dict[str, str]:
+        # Get the real device name (handles symlinks like /dev/mapper -> /dev/dm-X)
+        real_path = os.path.realpath(self.path, strict=True)
```
arighi, line 1088: Maybe add something like this to check if the path exists:
```
        try:
            real_path = os.path.realpath(self.path, strict=True)
        except FileNotFoundError:
            arg_fail(
                f"disk path does not exist (needed for topology=): {self.path!r}"
            )
```

arighi (changes requested):
I left a few comments (mostly stylish stuff). Overall looks good to me. I rarely use the --disk options, so I don't have a strong opinion on the redesign / behavior shifts.

But I share the same concern mentioned by @matttbe, we won't be able to maintain the scsi-disk opts mappings. However, I also don't have a strong opinion on this, if you guys think it's helpful and it can save time to hunt the right qemu options, I'm ok with that.

So, on my side, once those small nits that I reported are fixed, I'm ok to merge this.

matttbe:
Hi @intelfx

Thank you for the update. Same here, I missed the update.

In this PR, it is harder to see what has changed during the last update, because a rebase has been done at the same time as a modification of the code. Please next time do that in 2 steps (like you did in the other PR): rebase, push -f, fixup, push -f.

What did you do during the last modification? If you addressed @arighi's comments, do you mind either closing them, or leaving a comment, please.

If you addressed the last comments, and you uniform the style (`ruff format`) as requested by the CI (+ rebase), I guess we are good to merge!

virtme/commands/run.py
intelfx: Yes, indeed. Fixed, thanks.

virtme/commands/run.py
intelfx: I wish Python had Rust's optionals and match expressions, but in absence of those, this is the best I could come up with 😅 I'm open to suggestions.

virtme/commands/run.py
intelfx: Yes, definitely makes sense, thanks!

intelfx:
Hi @matttbe. I have addressed Andrea's suggestions and closed the relevant threads. Those threads where I had to defend my decisions (rather than just apply suggestions) I left open since it's not up to me to decide whether the concerns were, in fact, addressed.

Code reformatted commit-by-commit. I believe I did the rebase in two steps here (well, three; I forgot to fix one of the issues). I'd appreciate a review. Again, apologies for the protracted silence, I'm badly overcommitted.

Issue #353, opened 2025-10-04
virtiofsd: `-o cache=always` is confusing


matttbe:
Hello,

I played a bit with the new vng version, and I wonder if this modification from #340 with virtiofsd's `-o cache=always` is not too confusing. Here is an example:

- I start virtme-ng (from a container where kernel source code is mounted in the container, but I also tried without, I have the same issue)
- From the VM, I start a bash script from the kernel source (selftest)
- From the host, I modify this script
- On the VM side, the modifications are not seen, except if I execute: `echo 3 > /proc/sys/vm/drop_caches`

Because I remember this PR #340, I quickly identified the issue, but that might be quite confusing for someone else.

It is easy to reproduce: start vng, read a file from the guest, edit it from the host, read it again from the guest and the modifications are not there.

Because I never had this `ESTALE` error, I wonder if it is a good idea to use `cache=always` by default. Should it eventually be executed as an option, or only when needed?

matttbe:
> We can certainly add an option to change the virtiofsd caching, but, so far, I'm firmly convinced that the default should remain `cache=always`.

Thank you for your reply!

> The `ESTALE` errors were constantly disrupting my workflow, it's especially frustrating during long interactive debugging sessions when the shell history file on the host gets modified and the you completely lose the shell history in the guest. Errors like these are also perceived as bugs in vng, which isn't ideal.

Ah yes, I see how it can be annoying. On my side, most of the time, I mount a dedicated home for the work in the VMs. As I said, I never had these `ESTALE` errors.

With `-o cache=always` and your workflow, do you have any similar issues to the one I mentioned?
(_Start vng, read a file from the guest, edit it from the host, read it again from the guest and the modifications are not there._)

> Unfortunately, keeping the host and guest perfectly in sync isn't trivial because of caching. Conceptually, starting a vng guest should be seen like "forking" from the host, so (by default) there shouldn't be any inherent guarantee that the filesystems stay in sync. And we have ways to keep some directories in sync: `--rwdir` and `--rodir`.

Ah yes, the "forking" idea makes sense, especially if the modifications you do in the guest and not seen in the host.

Because I launch vng from a container, with a dedicated FS plus a couple of mount points (kernel source, home dir, localtime, some dev tools, etc.), I use `--rw`. I don't use `--rwdir` for my mount points not to use 9p (slower).

[...]

PR #391, opened 2025-12-15
Introduce share-path argument to allow a VM to mount a host directory

marcosps:
I'm adding this new argument in to make it easier to share test script to run inside the guests. I had to change the numa assignment inside the methods that export the virtiofsd, so please take a look if this is correct!

Currently it works for me, but maybe there are corner cases in the code that I didn't test.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -180,6 +180,16 @@ def make_parser() -> argparse.ArgumentParser:
         help="Add a read/write virtio-blk disk.  The device nodes will be /dev/disk/by-id/virtio-virtme_disk_blk_NAME.",
     )
     g.add_argument(
+        "--shared-path",
+        action="append",
+        default=[],
+        metavar="TAG=PATH_ON_HOST:PATH_ON_GUEST",
+        help="Share a directory with guest as virtiofs with tag (can be used multiple times). "
+        + "The expected format is: --shared-path tag_name=/path/on/host:/path/on/guest. "
+        + "This options is handy when used with --root, since it exposes one "
+        + "directory from the host in a different rootfs oon the guest.",
+    )
+    g.add_argument(
         "--memory",
         action="store",
         default=None,
@@ -905,6 +915,23 @@ def sanitize_disk_args(func: str, arg: str) -> tuple[str, str]:
     return name, fn
 
 
+# Validate tag=path arguments from --shared-path
+# Format expected:
+#   --shared-path tag_name=/path/on/host:/mount/path/on/guest
+def sanitize_shared_path_args(arg: str) -> tuple[str, str, str]:
+    tag, paths = arg.split("=")
+    path_host, path_guest = paths.split(":")
+
+    # Check if the directory exists in the
+    if not Path(path_host).exists():
+        arg_fail(f"shared_path {path_host} doesn't exists")
+
+    if not Path(path_host).is_dir():
+        arg_fail(f"shared_path {path_host} is not a directory")
+
+    return tag, path_host, path_guest
+
+
 def can_access_file(path):
     if not os.path.exists(path):
         return False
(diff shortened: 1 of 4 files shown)


virtme/commands/run.py
```diff
@@ -179,6 +179,13 @@ def make_parser() -> argparse.ArgumentParser:
         metavar="NAME=PATH",
         help="Add a read/write virtio-blk disk.  The device nodes will be /dev/disk/by-id/virtio-virtme_disk_blk_NAME.",
     )
+    g.add_argument(
+        "--shared-path",
+        action="append",
+        default=[],
+        metavar="TAG=PATH",
+        help="Share a direct using virtiofs with the guest.",
```
matttbe: I think a word is missing here. → _Share a direct XXX using (...)_

matttbe (changes requested):
That would be great to have multiple mount points with virtiofs! Just some suggestions: I initially didn't understand what you wanted to do when looking at the new option and description :)

(also, please squash the commit added by the CI :) )

virtme/commands/run.py
```diff
@@ -1323,6 +1353,26 @@ def do_it() -> int:
...
+        tag, path = sanitize_shared_path_args(arg)
+        virtiofs_config = VirtioFSConfig(
+            path=path,
+            mount_tag=tag,
+            guest_tools_path=guest_tools_path,
+            # virtiofsd requires a NUMA not, if --numa is specified simply use
+            # the user-defined NUMA node, otherwise create a NUMA node with all
+            # the memory.
+            memory=args.memory,
+            rw=args.rw,
```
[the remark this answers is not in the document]
matttbe: Good point! If it is in RW, we might want to have an option to control `virtiofsd -o cache=...`, see #353

marcosps:
> I guess the benefits of this is having a faster shared dir compared to --rwdir / --rodir, right?
> 
> I haven't tested this yet, but do we need to manually mount the shared dirs in the guest (I was expecting to see some changes in virtme-init/virtme-ng-init as well)?

I forgot to tell that this works when `--root` is passed, so this feature is similar to using the `--volume` on containers, mapping a dir from the host inside the VM.

virtme/commands/run.py
marcosps: Well, sorry for not making it clear. I'm using this new option with `--root`, so the directory is not available to the VM. I'm trying to test different versions of SLE, with different userspace rootfs, but my test scripts and livepatches are on the host, so I'm using virtiofsd to make them available to the guest.

Regarding this usage, I'm not sure if this is similar to what you proposed. I never used `overlay-rwdir` so I;m not how this can be used ...

matttbe:
> @matttbe @arighi since I'm using this new option together with `--root`, maybe we can restrict it to the same situation? I mean, it currently doesn't make much sense to use it when we are sharing the same rootfs with the host, right?

@marcosps I think it can be useful to share some directories in RW, while the rest is in an "overlayfs" mode. (In this case, we might need something similar to #355)

So having the same FS as the host, except some directories that are "shared" to pass data, but having a better performance than using 9P. No?

virtme/commands/run.py
matttbe: OK but then `--rodir` might be enough for your case then? Or do you need something faster? (I think it still makes sense to have it, but maybe you don't really need it for your case?)

marcosps:
> > @matttbe @arighi since I'm using this new option together with `--root`, maybe we can restrict it to the same situation? I mean, it currently doesn't make much sense to use it when we are sharing the same rootfs with the host, right?
> 
> @marcosps I think it can be useful to share some directories in RW, while the rest is in an "overlayfs" mode. (In this case, we might need something similar to #355)
> 
> So having the same FS as the host, except some directories that are "shared" to pass data, but having a better performance than using 9P. No?

Does `--overlay-rw` woprk with rootfs? I've being trying to read the code and checking using the flag (it doesn't really describe how it should be used, but per the code it seems that it needs a /guest/path=/home/path argument, like --sahred-path).

But I can't see the the guest directory being populated by vng. Am I using it wrong?

marcosps:
(Sorry to flood here...)
@matttbe ah, ok, now I get what you and @arighi were talking about: `--rodir` can also map directories to the guest, but for now I can't do that when using `--root` with /mnt or /tmp for example:

```
error: cannot initialize /tmp/ inside the guest (path must be defined inside a valid overlay)
```

I'll check what it means...

PR #394, opened 2025-12-16
pre-commit: disable autofix and switch auto-updates to monthly

matttbe:
This is linked to discussions on #391.

To reduce noise, pre-commit.ci will now have less auto-updates and auto-fixes. See the explanations in the different commits for more details about that.

Link to the doc: https://pre-commit.ci/#configuration


arighi (approved):
Awesome! Thanks for looking at this @matttbe!

PR #395, opened 2025-12-17
virtme-ng: MCP: Improve kselftest support

arighi:
As mentioned in https://github.com/arighi/virtme-ng/pull/393#discussion_r2626421014, make sure that all the required kernel configs are enabled when running a kselftest, if not automatically  enable them and recompile the kernel.

Also, run `make headers_install` when building a kselftest, since many  kselftests depend on the installed kernel headers and get rid of build_kselftest command, as the build is integrated in run_kselftest now.


matttbe (approved):
Thx! Just one question. Feel free to merge with or without a new modification.

PR #396, opened 2025-12-17
virtme-ng: MCP: Make verify_kernel asynchronous

arighi:
Each verify_kernel step can take a long time, due to the kernel rebuild and the command itself used to validate the kernel.

Therefore, make this step asynchronous as well, similarly to run_kselftest and run_kernel_async.

### DIFF
--- a/virtme_ng/mcp.py
+++ b/virtme_ng/mcp.py
@@ -390,27 +390,36 @@ IMPORTANT NOTES FOR AI AGENTS:
    ───────────────────────────────────────────
 
    Basic verification:
-   verify_kernel({
+   result = verify_kernel({
        "commit": "abc1234",
    })
+   # If "job_id" in result: poll with get_job_status({"job_id": result["job_id"]})
 
    With remote build host (faster):
-   verify_kernel({
+   result = verify_kernel({
        "commit": "abc1234",
        "build_host": "builder",
    })
 
    With custom config:
-   verify_kernel({
+   result = verify_kernel({
        "commit": "abc1234",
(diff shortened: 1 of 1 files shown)


matttbe:
> > Just to be sure, when you say that this task is running asynchronously, it means that if there are no results after 60 seconds, the validation will continue in the background, and the user will be able to prompt more tasks.
> 
> Correct. And when it runs in the background the AI agent is able to check the status periodically via `list_jobs()` and `get_job_status()`. So the user won't be able to prompt more tasks, it's just a way to bypass the timeout that most AI agents have when running MCP commands.

OK, thank you, that's clearer!

> > But it doesn't mean that multiple validations are done in parallel (parallel builds from the same workspace), right?
> > If that's the first one, that looks good to me.
> 
> Oh yes, definitely no multiple parallel validations.
> 
> > One last thing, I see that it checks that there is no running job for the specific commit, but should it not be a check per workspace instead? Not to build kernel from commit A, and run selftests scripts from commit B.
> 
> The idea is to use the commit as a unique identifier to identify a background job. In theory the unique id is not even needed, because you shouldn't run multiple verify_kernel in parallel. Also each AI agent session runs its own `vng --mcp` instance (so you don't have multiple clients connecting to the same server).
> 
> But a smarter model could work like a pipeline if you ask to verify multiple commits, like: run verify_kernel(A), wait for A to complete, then start verify_kernel(B), while B is running check the result of A (which is completed at this point). That's how the unique identifier can help.

OK, as long as there are no forbidden parallel tasks.

PR #399, opened 2025-12-25
virtme-ng-init: Mount /dev before logging

arighi:
Ensure that mount_kernel_filesystems() is called before configure_hostname() to guarantee that /dev (specifically /dev/kmsg) is available before any code attempts to use the log!() macro.

Fix by swapping the order: run mount_kernel_filesystems() before configure_hostname().

Fixes: 71de7b5 ("virtme_ng_init: Add multiple missing log statements")


matttbe (approved):
Good catch, thanks!

And Merry Christmas 🎄

PR #403, opened 2026-01-09
virtme-init: Mount /dev before logging

mhartmay:
Ensure that `mount_kernel_filesystems` is called before `configure_hostname` to guarantee that `/dev` (specifically `/dev/kmsg`) is available before any code attempts to use the `log` function.

See commit 954d73093116 ("virtme-ng-init: Mount /dev before logging").

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -658,8 +658,8 @@ setup_user_session() {
 
 # Basic system initialization (order is important here).
 configure_environment
-configure_hostname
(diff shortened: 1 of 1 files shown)


mhartmay:
One question that I've is, how does this work with the Systemd integration (`--systemd`)? Shouldn't we then write to stdout/stderr instead so the output of the "virtme-init" is managed by systemd?

marcosps:
Maybe @rbmarliere can give an opinion here

PR #404, opened 2026-01-13
systemd: Use a drop-in for serial-getty instead of a fragment

rbmarliere:
Old systemd versions do not support the use of partial unit fragments,
therefore let's override the system service with a drop-in replacement
instead. With this, we make sure virtme-init runs and therefore we don't
hit a login prompt, while also re-enabling `--exec` scripts. Tested in
SLE12-SP5 (systemd v228).


marcosps:
@matttbe @arighi @mhartmay could you please take a look? It would be great if this was merged, as we are using vng to test livepatches at SUSE!

PR #406, opened 2026-01-20
virtiofsd: Enable support for posix ACLs + small fixes

rbmarliere:
Setting file capabilities are not supported when using external roots
(`--root`) without uid 0, e.g.:

```
$make $rootfs
sudo chown $(whoami):$(whoami) $rootfs
cd $rootfs
unshare -r --map-auto -- vng --user root --root . --run boot/vmlinuz
```

It fails with ENOTSUP:

virtme-ng:/ # setcap 'cap_net_admin+ep' cat
Failed to set capabilities on file 'setcap': Operation not supported

Fix by passing `--posix-acl` to virtiofsd.

Link: https://gitlab.com/virtio-fs/virtiofsd/-/blob/c0d8495c79cb4d2337f6424e81c86a6e2ed7b23c/README.md?plain=1#L147

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -759,7 +759,7 @@ class VirtioFS:
         else:
             stderr = ""
         os.system(
-            f"{virtiofsd_path} --syslog --no-announce-submounts "
+            f"{virtiofsd_path} --syslog --no-announce-submounts --posix-acl "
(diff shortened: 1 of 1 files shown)


arighi:
I think we merged this one as part of #410, right @rbmarliere @marcosps ? Can we close this?

PR #407, opened 2026-01-20
setuptools is not needed at runtime

a-detiste:
Hi,

I'm disussing this patch in Debian bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1125855


a-detiste:
Thank you !

PR #411, opened 2026-01-30
ppc64le: Enable virtiofs support

marcosps:
While testing vng and running zypper to install new kernels, it was showing messages about rpm unknown db, like this:

virtme-ng:/ # zypper ref
warning: Rebuilding outdated index databases
rpmidxOpenXdb: Invalid argument
error: cannot open Name index using unknown db - Operation not permitted (1)

An strace run presented that the rpm binary was trying to execute flock system call, and it was failing. Ony then I found out that the VM was not using virtiofsd. After some research I discovered that 9pfs on qemu doesn't implement flock properly.

Either way, I found that if the virtiofs support is enabled the problem is gone, and it works properly.

### DIFF
--- a/virtme/architectures.py
+++ b/virtme/architectures.py
@@ -324,6 +324,10 @@ class Arch_ppc(Arch):
         self.gccname = "powerpc64le"
 
     @staticmethod
+    def virtiofs_support() -> bool:
+        return True
+
+    @staticmethod
     def qemuargs(is_native, use_kvm, use_gpu):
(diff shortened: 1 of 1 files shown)


marcosps:
Maybe @hramrach can say if this can break on ppc (not being ppc64le, the one I tested). Also, I'm not sure if @rbmarliere is testing kernels on ppc64le as well, so maybe more people can be interested into this fix :)

PR #413, opened 2026-02-03
virtme-init: Find correct busybox binary

rbmarliere:
When using external roots with foreign architectures, virtme-ng depends on
busybox-static. However, it always use the `busybox` binary. Some distros
might install a different binary for the static variant, e.g.
`busybox-static`. Check if there is `busybox-static` in the $PATH to take
precedence over `busybox`.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -330,6 +330,16 @@ override_system_files() {
     generate_lvm
 }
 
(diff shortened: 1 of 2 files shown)


marcosps:
Tested-by: Marcos Paulo de Souza <mpdesouza@suse.com>

Works as expected, thanks!

PR #414, opened 2026-02-04
Feature/qmp gdb

clintonkavai:
Motivation
Our team needed to run multiple virtme-ng VMs on the same host, but QMP was hardcoded to a TCP
    port, causing collisions. We switched QMP to a per‑VM Unix socket to remove the port
    conflict, and made the GDB port configurable to avoid 1234 clashes when running multiple
    debug sessions  

Summary

  - Switch QMP to a Unix socket per VM name to avoid TCP port collisions.
  - Add --gdb-port to make the GDB stub port configurable (default 1234).
  - Improve error messaging when GDB port 1234 is already in use.
  - Add safety check for existing QMP sockets and auto-clean them on exit.

  Tests

  - ./vng -r v6.6 -v -- uname -r
  - make
  - ./vng -r v6.6 -v -- cat /sys/devices/system/clocksource/clocksource0/current_clocksource

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -214,7 +214,8 @@ virtme-ng is based on virtme, written by Andy Lutomirski <luto@kernel.org>.
     parser.add_argument(
         "--gdb-port",
         action="store",
-        default="1234",
+        type=int,
+        default=1234,
         help="GDB TCP port to use when --debug is enabled (default: 1234)",
     )
 
@@ -725,25 +726,11 @@ def get_host_arch():
     }
     return arch_map.get(arch, None)
 
+
 def _get_qmp_path(name):
     base = name or socket.gethostname()
     safe = re.sub(r"[^A-Za-z0-9_.-]+", "_", base)
     return f"/tmp/virtme-ng-{safe}.qmp"
-
-
-def _is_tcp_port_free(port):
-    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    try:
-        port = int(port)
-    except ValueError:
-        return False
-    try:
-        sock.bind(("127.0.0.1", port))
-        return True
-    except OSError:
-        return False
-    finally:
-        sock.close()
 
 
 def _qmp_socket_available(path):
@@ -1385,13 +1372,10 @@ class KernelSource:
                 )
                 sys.exit(1)
             self.qmp_path = qmp_path
-            gdb_port = args.gdb_port or 1234
-            if not _is_tcp_port_free(gdb_port):
-                sys.stderr.write(
-                        f"GDB port {gdb_port} is in use. Rerun with --gdb-port <PORT>.\n"
-                    )
-                sys.exit(1)
-            qemu_args += f"-gdb tcp::{gdb_port} -qmp unix:{qmp_path},server,nowait "
+
+            gdb_port = args.gdb_port
+            qemu_args += f"-gdb tcp:127.0.0.1:{gdb_port} "
+            qemu_args += f"-qmp unix:{qmp_path},server,nowait "
         if args.qemu_opts is not None:
             qemu_args += " ".join(args.qemu_opts)
         if qemu_args != "":
@@ -1506,7 +1490,7 @@ class KernelSource:
         try:
             check_call(cmd, shell=True)
         finally:
-            #Cleanup the QMP socket
+            # Cleanup the QMP socket
             if self.qmp_path and os.path.exists(self.qmp_path):
                 try:
                     os.remove(self.qmp_path)
@@ -1752,12 +1736,14 @@ def dump(kern_source, args):
     kern_source.dump(args)
     return True
 
+
 def is_kernel_source_tree():
     return (
         os.path.exists("Kconfig")
         and os.path.exists("Makefile")
         and os.path.isdir("arch")
     )
+
 
 def do_it() -> int:
     """Main body."""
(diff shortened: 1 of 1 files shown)


virtme_ng/run.py
```diff
@@ -1732,6 +1738,8 @@ def do_it() -> int:
         elif args.dump is not None:
             dump(kern_source, args)
         elif args.build or args.kconfig:
+            if not is_kernel_source_tree(): 
+                arg_fail("Error: must run from a kernel source tree", show_usage=False)
```
arighi: I think this should be moved to a separate PR, it's not really related to QMP or gdb.

Also, I've always been a bit reluctant to rely on file-matching logic to detect whether we're in a kernel source tree or not, it feels a bit fragile...

arighi (commented):
Except for the is_kernel_source_tree() part, everything else looks good to me. If we can move/discuss that in a separate PR it'd be great, then we can merge the rest to me.

virtme_ng/run.py
```diff
@@ -715,6 +725,36 @@ def get_host_arch():
...
+    safe = re.sub(r"[^A-Za-z0-9_.-]+", "_", base)
+    return f"/tmp/virtme-ng-{safe}.qmp"
+
+
+def _is_tcp_port_free(port):
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    try:
+        port = int(port)
+    except ValueError:
+        return False
```
matttbe: Can you check the port before creating the socket please? (Or close it in case of error or use `with` to narrow the scope.)

Or maybe there is no need to do the conversion? (I didn't check but it feels it is not needed)

virtme_ng/run.py
```diff
@@ -210,6 +211,12 @@ def make_parser():
         action="store_true",
         help="Attach a debugging session to a running instance started with --debug",
     )
+    parser.add_argument(
+        "--gdb-port",
+        action="store",
```
matttbe: can you add `type=int` to force the conversion and complain here if it is not an integer please?

virtme_ng/run.py
```diff
@@ -1337,7 +1378,20 @@ def _get_virtme_qemu_opts(self, args):
         qemu_args = ""
         if args.debug or args.pin:
             # Enable debug mode and QMP (to trigger memory dump via `vng --dump`)
-            qemu_args += "-s -qmp tcp:localhost:3636,server,nowait "
+            qmp_path = _get_qmp_path(args.name)
```
matttbe: you could also use `str(args.gdb_port)` instead of `args.name` → so only the port needs to be given and be unique, no?

(if yes, please adapt the error message below, and probably better to check GDB first ; and maybe if the socket already exists, it could be silently overridden? → Not sure, to be checked)

virtme_ng/run.py
```diff
@@ -1337,7 +1378,20 @@ def _get_virtme_qemu_opts(self, args):
...
+                )
+                sys.exit(1)
+            self.qmp_path = qmp_path
+            gdb_port = args.gdb_port or 1234
+            if not _is_tcp_port_free(gdb_port):
+                sys.stderr.write(
+                        f"GDB port {gdb_port} is in use. Rerun with --gdb-port <PORT>.\n"
+                    )
+                sys.exit(1)
+            qemu_args += f"-gdb tcp::{gdb_port} -qmp unix:{qmp_path},server,nowait "
```
matttbe: please keep `localhost`, otherwise external people could also connect.

virtme_ng/run.py
```diff
@@ -1337,7 +1378,20 @@ def _get_virtme_qemu_opts(self, args):
...
-            qemu_args += "-s -qmp tcp:localhost:3636,server,nowait "
+            qmp_path = _get_qmp_path(args.name)
+            if not _qmp_socket_available(qmp_path):
+                sys.stderr.write(
+                    "QMP socket already exists. Rerun with a unique --name or stop the existing VM.\n"
+                )
+                sys.exit(1)
+            self.qmp_path = qmp_path
+            gdb_port = args.gdb_port or 1234
+            if not _is_tcp_port_free(gdb_port):
```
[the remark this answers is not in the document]
clintonkavai: > I agree unix sockets are nice, but I’d keep GDB on TCP (localhost‑only) for now to avoid breaking common workflows. We can consider adding an opt‑in --gdb-socket later if we want parity with QMP

virtme_ng/run.py
clintonkavai: > Good catch, and it’s already addressed. I'll remove the int() conversion and make --gdb-port type=int, so _is_tcp_port_free no longer creates a socket before
  validation. That eliminates the leak risk.

arighi:
Yeah, I agree with @matttbe, it's a bit difficult for us to follow the changes if we open a new PR...
@clintonkavai can you just force push your new branch into this one and re-open this PR? Thanks!

clintonkavai:
> > I'll close this and reopen a fresh PR
> 
> Please keep the same one for us to be able to track changes. Except if you changed the scope, split, etc. But I guess no
> 
> I will try to continue my review tomorrow

> I deleted the repository so I'm unable to re open. I Would appreciate if you we can continue the discussion on the new PR, sorry for this

virtme_ng/run.py
```diff
@@ -1449,13 +1503,28 @@ def run(self, args):
         )
         if args.pin:
             self.set_affinity(args)
-        check_call(cmd, shell=True)
+        try:
+            check_call(cmd, shell=True)
+        finally:
+            #Cleanup the QMP socket
```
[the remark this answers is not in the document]
matttbe: > My thought process was since the socket is created by QEMUm it should only be unlinked after QEMU exits. Using atexit risks removing the path while the VM is still running (e.g., if our process is interrupted), which would break QMP for a live guest. The console/script‑sh case is different because those files are created by us upfront.

I'm not sure to understand: `check_call(cmd, shell=True)` will launch the VM and stop when the VM is stopped. `atexit` will be executed at the end, so also after `check_call()`, no? Using `atexit` next to where the file has been created seems clearer and easier to maintain.

PR #416, opened 2026-02-06
GDB: unify code in 'virtme' + localhost only + option to change port

matttbe:
Hello,

This is linked to [this comment](https://github.com/arighi/virtme-ng/pull/414#discussion_r2768348061), it seems better to explain that with code.

The idea is to unify the code around GDB in the `virtme` part. The code to run the server was previously done in `vng`, while the one to connect to an instance was done in `virtme`.

While at it, restrict the server side to localhost only, and add the ability to change the TCP port number (just because it was simple).

cc: @clintonkavai

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -114,13 +114,21 @@ def make_parser() -> argparse.ArgumentParser:
     )
     g.add_argument(
         "--gdb",
-        action="store_true",
+        action="store",
+        nargs="?",
+        type=int,
+        const=1234,
+        metavar="PORT",
         help="Attach a GDB session to a running instance started with --gdb-server",
     )
     g.add_argument(
         "--gdb-server",
-        action="store_true",
-        help="Open a GDB server on TCP port 1234 to connect to the guest",
+        action="store",
+        nargs="?",
+        type=int,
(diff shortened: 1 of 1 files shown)


arighi:
@clintonkavai @matttbe I lost track of the PRs for this feature. Since this was suggested by @clintonkavai I'd like to maintain the authorship/credit as well. Does this include all the changes suggested by @clintonkavai? Is there anything missing?

matttbe:
@arighi thank you for your message, I see it is indeed not clear. This PR doesn't replace #414 but it is linked to a [comment](https://github.com/arighi/virtme-ng/pull/414#discussion_r2768348061) I sent which was not understood at that time, and I thought it would be quicker to modify a few lines of code than explaining it: I think it is better to gather the code around GDB in one place, in `virtme-run`, not some there and in `vng` not to deal with hardcoded in multiple places.

While at it, I added a restriction to localhost that I mentioned in another review comment, and prepared `virtme-run` to be able to change the GDB port number. @clintonkavai PR is still needed for those who want to use multiple `vng --debug` in parallel. After this PR here, only the `vng` side needs to be modified I think.

PR #417, opened 2026-02-06
virtme: Always set virtme_console kernel parameter

rbmarliere:
`virtme_console` is only set in the interactive path. In the script path (`--script-sh`, used by `vng --exec`), this can cause virtme-init to pick an invalid console on some external roots.

Set `virtme_console` unconditionally so virtme-init selects the intended serial console.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1500,6 +1500,8 @@ def do_it() -> int:
     qemuargs.extend(["-parallel", "none"])
     qemuargs.extend(["-net", "none"])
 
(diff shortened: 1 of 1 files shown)


virtme/commands/run.py
```diff
@@ -1542,10 +1544,6 @@ def do_it() -> int:
         if not args.disable_monitor:
             qemuargs.extend(["-mon", "chardev=console"])
```
matttbe: I forgot how all of this work, but do you not need some of the configuration here above to have for `virtme_console` to work?

I guess no (I should check how `virtme_console` is used, the code here and commit message doesn't help me to remember :) )

PR #423, opened 2026-02-09
shell: move option handling to virtme + support in 'console' mode

matttbe:
`--shell` in `vng` was the only option setting a kernel option directly from there, while all other options are handled from `virtme-run`. So the handle of this option is now done there as well.

Thanks to that, it is possible to have `--shell` being used when `--console-client` is passed. While at it, now `--console-client` use the shell previously specified when launching the VM with `--console --shell=X`.

Note: I don't know how to properly have something similar with SSH... If someone has an idea, feel free to suggest! My initial thought was to use `chsh`, but `/etc/passwd` is locked by default. Also, I don't think it is a good idea to force the remote command. Maybe using `ssh -t <shell>` if no remote command is passed? If someone is interested in that, it might be easier if this person can open another PR, no?

Note: This PR depends on #422 (to avoid a simple conflict).

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1030,7 +1030,10 @@ def console_client(args):
 
     user = args.user if args.user else "${virtme_user:-root}"
 
-    shell = '${virtme_shell:+-s "${virtme_shell}"}'
+    if args.shell is not None:
+        shell = f'-s "{args.shell}"'
+    else:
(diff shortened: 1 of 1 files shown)


arighi (approved):
LGTM. I think we can ignore the `--ssh` case for now, it was ignored before as well, so we're not changing or breaking old behavior.

Issue #405, opened 2026-01-15
Invalid json in .virtme-ng.conf example


kdave:
In section "Default options" the json is not entirely valid, the "," on line 5 must not be there:

```json
{
    "default_opts": {
        "build_host": "kathleen",
        "build_host_exec_prefix": "schroot -c chroot:lunar-amd64 --"
    },
}
```

`jq` says:
```
jq: parse error: Expected another key-value pair at line 6, column 1
```

PR #424, opened 2026-02-18
README.md: Remove wrong comma in default options json

marcosps:
Closes: #405


matttbe (approved):
Thank you for the fix and the report!

PR #425, opened 2026-02-20
virtme: Treat root as a valid overlay for --rwdir

### LINKED ISSUE #409: `--rwdir` with absolute path no longer works
I would like to mount one specific directory as read-write for the guest. Previously (1.18), simply specifying `--rwdir /some/absolute/path/to/directory` worked. With 1.40, that does not work:

```
error: cannot initialize some/absolute/path/to/directory inside the guest (path must be defined inside a valid overlay)
```

However, doing `--rwdir /some/absolute/path/to/directory=/some/absolute/path/to/directory` works. But the two should be equivalent, no? Am I misunderstanding how single arg `--rwdir` is supposed to work?

arighi:
The guest-path check for --rwdir only considered overlay_rwdir when the path is inside a valid overlay. However, it never treats the root filesystem as valid. So single-arg --rwdir could fail with:

  error: cannot initialize ... inside the guest (path must be defined inside a valid overlay)

Fix by considering the root a valid overlay, so --rwdir with a single argument works for any path under root without relying on default overlay_rwdir or cwd.

This fixes #409.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1490,8 +1490,9 @@ def do_it() -> int:
         # Check if paths are accessible both on the host and the guest.
         if not os.path.exists(hostpath):
             arg_fail(f"error: cannot access {hostpath} on the host")
-        # Guest path must be defined inside one of the overlays
-        guest_path_ok = False
+        # Guest path must be defined inside a valid overlay (root or overlay_rwdir).
+        # The root is always mounted in the guest, so any path under root is valid.
+        guest_path_ok = not os.path.normpath(guestpath).startswith("..")
         for d in args.overlay_rwdir:
             if os.path.exists(guestpath) or is_subpath(guestpath, d):
                 guest_path_ok = True


virtme/commands/run.py
```diff
@@ -1490,8 +1490,9 @@ def do_it() -> int:
         # Check if paths are accessible both on the host and the guest.
         if not os.path.exists(hostpath):
             arg_fail(f"error: cannot access {hostpath} on the host")
-        # Guest path must be defined inside one of the overlays
-        guest_path_ok = False
+        # Guest path must be defined inside a valid overlay (root or overlay_rwdir).
+        # The root is always mounted in the guest, so any path under root is valid.
+        guest_path_ok = not os.path.normpath(guestpath).startswith("..")
```
matttbe: out of curiosity, why do you need `normpath()`? Will it impact the beginning of the path?

Maybe did you want to use `os.path.relpath()` like it is done at the line 1486?

matttbe: I see, thank you, good point.

So when only one path is given (what is done in #409), this check will always be true, and that's normal because any valid path under root is OK.

Just to be sure, the goal is to catch issues like that one where `../BAR` is not under root?

    --root ROOT --rwdir=/GUEST=../HOST_PARENT

If yes, is it not clearer to check this instead?

    os.path.relpath(guestpath, args.root).startswith("..")

If yes, then maybe the `arg_fail()` at line 1488 can be removed? And we could even have `hostpath = guestpath = m.group(1)`, less confusing?

arighi: Yes. The goal is to catch when the path is not under root. And we only care whether the guest path escapes the guest root.

normpath() collapses things like var/../../etc o ../etc, so checking normpath(guestpath).startswith("..")  detects "escapes guest root" using only the guest path string, without mixing in host paths.

We can probably do this with relpath as well, but it looks more complicated to me, unless I'm missing something...?

PR #432, opened 2026-03-04
commands: run: Add systemd.hostname when --systemd is used

marcosps:
When running vng with a script, sometimes I can observe systemd setting the hostname to localhost, later on virtme_ng_init set's the hostname, but /proc/sys/kernel/hostname still shows "localhost".

Ensure that both the vng-init and systemd have the same value when --systemd is required.


arighi (approved):
Looks good, thanks!

PR #433, opened 2026-03-05
virtme: Improve shell handling and support minimal rootfs

arighi:
Always prefer the user's shell in the vng sessions: use the host $SHELL or --shell when starting the guest session and fall back to /bin/sh only if that shell is not present in the guest.

Use /bin/sh in virtme-ng-init instead of hard-coding bash, so minimal rootfs (without bash) can be supported. Keep using bash for virtme-init since strict /bin/sh would just exit without TTY control.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1917,6 +1917,8 @@ def do_it() -> int:
     if args.user and args.user != "root":
         kernelargs.append(f"virtme_user={args.user}")
 
+    # Pass shell for guest only when --shell is explicitly set, otherwise
+    # init uses /bin/sh.
     if args.shell is not None:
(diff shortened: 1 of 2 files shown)


arighi:
Re-opening, passed all my testing now.

Unfortunately it seems that we can't completely rewrite virtme-init to work 100% using /bin/sh, since without TTY control the shell is just exiting, so interactive mode in a rootfs that only has /bin/sh works only using virtme-ng-init.

matttbe (approved):
Good idea not to hardcode bash! I thought using `$SHELL` might break things, but it looks OK.

Out of curiosity, why not using `/bin/sh` instead of bash, except when `--shell` is explicitly passed?

PR #434, opened 2026-03-06
Simplify ssh authentication

rbmarliere:
The goal here is to standardize ssh authentication around a dedicated `id_virtme` identity created at runtime if needed and always use the cache directory for this. This makes the authentication more robust for setups that need to use `--root`.

EDIT: Also, make sure that `--user` option is carried over as the ssh user.


arighi (approved):
LGTM, thanks!

PR #439, opened 2026-03-08
vng-init: switch to sh by default

matttbe:
It looks like bash is currently used to execute simple commands, or as fallback. No need to add this hard dependence to bash for such simple tasks. Instead, use 'sh' to support minimal rootfs.

Tested with:

```
  $ docker run -it -v $PWD:$PWD -w $PWD -v --privileged --rm \
      alpine:latest sh
  # apk add qemu-system-x86_64 python3 busybox-static py3-argcomplete \
      py3-requests linux-virt file
  # ./vng -r /boot/vmlinuz-virt -v
  ># ls -l /proc/$$/exe
   lrwxrwxrwx 1 root root 0 Mar 8 12:16 /proc/1069/exe -> /bin/busybox
```

Note that this doesn't change the login prompt: 'su' will continue to be executed and load the registered prompt, except if --shell is used.

Also, USER_SCRIPT is still launched with 'bash' when a GUI is used. I was not able to test on my headless server, by the look of it, it should be fine to use 'sh' as well, but such environment will certainly have bash support, so no need to change. When a user script is launched without GUI, 'sh' was already used.

The first commit reverts #433.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -653,18 +653,13 @@ run_user_gui() {
 
 run_user_shell() {
     local consdev=$1 uid=$2
+    local args=()
 
     print_logo
 
-    if [[ -n ${virtme_user:-} ]]; then
-        shell=""
-        if [[ -n ${virtme_shell:-} ]]; then
-            shell="-s ${virtme_shell}"
-        fi
-        setsid bash -c "su ${shell} -- ${virtme_user}" 0<> "/dev/$consdev" 1>&0 2>&0
-    else
-        setsid bash 0<> "/dev/$consdev" 1>&0 2>&0
-    fi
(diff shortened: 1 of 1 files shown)


arighi:
A couple of issues:
1) `vng -r v6.6 --root ~/rootfs/ubuntu/ --shell /bin/tmux` doesn't work, it drops me to a `/bin/sh` session instead of tmux
2) using --root with an alpine rootfs is still broken for me (virtme-ng-init exits), which is the original issue that I was trying to fix

I started to fix these issues, but I'm essentially re-adding the same logic form #437... I know it's not nice to have more complexity, but everything was working with #437. At this point I'd prefer to just revert #433.

@matttbe do you have any other concerns with #437 apart from the extra complexity? wdyt?

PR #449, opened 2026-03-14
Improve compatibility with older rootfs

rbmarliere:
Hi! Here are a few small fixes to increase usage with older root filesystems.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -359,6 +359,7 @@ setup_network() {
         # udev is liable to rename the interface out from under us.
         for d in /sys/bus/virtio/drivers/virtio_net/virtio*/net/*; do
             virtme_net=$(basename -- "${d}")
+            ip link set dev "$virtme_net" up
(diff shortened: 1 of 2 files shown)


arighi:
I share the same concern as @matttbe on #451, I'm not sure if we want to support such old systems. Everything looks good now, but we will likely end up breaking the compatibility again as we change or add new things, unless we put an old rootfs in the CI, which I wouldn't do honestly.

This commit look good in general:
- virtme-init: bring virtio-net links up before DHCP: this can fix a real issue (DHCP on DOWN interface)

Commits with potential issues:
- virtme-init: don’t block on udevd --daemon: we tried starting udevd asynchronously in the past and it was causing issues, so nack on this one
- virtme-init: avoid Bash 4-only features: it works now, we'll break compatibility in the future for sure, so nack on this

About CONFIG_EXPERIMENTAL and CONFIG_SMP, I'm not opposed to add them, but they're old options, I think we should move forward.

matttbe:
> About CONFIG_EXPERIMENTAL and CONFIG_SMP, I'm not opposed to add them, but they're old options, I think we should move forward.

A note can maybe be added in the README instead? It is easy to add kernel options.

PR #458, opened 2026-03-31
Add support for repeated `--verbose` and demote `debug` to that

intelfx:
(This contains commits that were split off from #388)

THis makes two more changes that synergize with the `--fb` option:
- a single `--verbose` only enables `loglevel=7` and `--show-boot-console`
- a repeated `--verbose` is now required to also enable `debug`

This makes it easier to view the kernel boot log on the stdio console (as well as the early console log, which is also tied to `--verbose` elsewhere) without also suffering the effects of the `debug` cmdline option (which creates much more output and also potentially affects the userspace).

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -150,7 +150,10 @@ def make_parser() -> argparse.ArgumentParser:
         + "An argument can be optionally specified to start a graphical application.",
     )
     g.add_argument(
-        "--verbose", action="store_true", help="Increase console output verbosity."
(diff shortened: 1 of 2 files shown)


matttbe (approved):
Thank you for the update. It looks good to me, but I prefer to wait a bit to get more feedback from others that might have heard of issues this could cause.

arighi (approved):
Looks good to me, I don't see any issues or potential side effects. Thanks!

PR #461, opened 2026-04-01
run: set `IdentitiesOnly yes` in the ssh client config

intelfx:
This way, ssh will strictly use the specified IdentityFile, which is
the intended behavior (as a dedicated key is generated and injected
into both ssh server in the guest and ssh client on the host).

Otherwise, the agent takes precedence, potentially triggering hardware
tokens or other annoyances, if the user's main private key in the agent
is also added to their own `authorized_keys`.


matttbe (approved):
Good idea: here an agent is not needed, and the identity file will not have any password.

PR #462, opened 2026-04-01
virtme_ng_init: better `sudoers` integration

intelfx:
This PR does two things in order to improve ergonomics around sudo usage within the virtme instance.

- **virtme_ng_init: install override into /etc/sudoers.d if supported**
  If `/etc/sudoers.d` exists and is included, write a drop-in instead of overriding the entire config. This way, we preserve the distro-specific or user-specific customizations that might have been done (e.g., the user might have added custom `Defaults env_keep` statements important to their workflow).

- **virtme_ng_init: do not override $PATH if adding a sudoers.d dropin**
  In the same vein, if we are writing a drop-in, do not provide a `Defaults secure_path` statement. Injecting a reasonable typical secure_path makes sense if we are writing a config from scratch, but if we are adding to a pre-existing config, it is fair to assume that the secure_path would already be set (or unset!) according to the user's expectations (e.g., the user might have elected to _not_ use a secure_path to access binaries importatnt to their workflow, and we should not override that decision).

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -313,7 +313,7 @@ generate_sudoers() {
         printf -- '%s ALL = (ALL) NOPASSWD: ALL\n' "${virtme_user}" >> "$tmpfile"
     fi
     chmod 440 "$tmpfile"
-    if [ ! -f "$real_sudoers" ]; then
+    if [[ ! -f $real_sudoers ]]; then
         touch "$real_sudoers"
     fi
     mount --bind "$tmpfile" "$real_sudoers"
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -316,7 +316,9 @@ generate_sudoers() {
         real_sudoers=/etc/sudoers.d/99-virtme
     fi
     tmpfile="$(mktemp --tmpdir=/run/tmp)"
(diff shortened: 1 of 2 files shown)


matttbe (changes requested):
(I forgot to change the status + the CI spot some style issues if you don't mind fixing them as well please)

matttbe:
Hi @intelfx,

I just noticed the modifications you did, I thought it was only a rebase. The modifications in `virtme-init` look good to me, but not in `virtme-init-ng`: apparently when `/etc/sudoers` is not readable (`0440`), the file is no longer replaced via a new `mount --bind` point.

matttbe:
Hi @intelfx 

Thank you for the update. It looks good to me, but the CI is asking if you can keep the code uniform by using `cargo fmt`. Do you mind doing that, please?

(Sorry for the delay. Please next time also add a comment explaining what you did, to make it clearly visible. Again this time, I missed it: I guess I saw the new push, but the CI was failing, and I probably thought this was still WIP.)

intelfx:
I believe I have fixed both cases (sudoers or sudoers.d do not exist, are not included, or not readable by the user ⇒ fallback to overmounting /etc/sudoers) in both inits. Code reformatted commit-by-commit. Apologies for the delay; I'm sorely overcommitted.

@matttbe please re-review.

PR #465, opened 2026-04-22
vng: Add --jobs option to control CPU nums used for building kernel

mmz-zmm:
For people with limited building resources(just one PC), default build uses all the cpu core and makes other daily work unaviable to perform. Even with a remote machine with rich cores, it is not always nice to occupy all the core. So, add two config options to control CPU nums used for building kernel:
1. --jobs, -j N: used to control local building.
2. --build-host-jobs N: used to control remote host building.

I met this two situations myself before, and now I find someone else also met this, so we better provides a way to limite the building jobs, just like what make -j N did.

Closes #460

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -472,6 +472,14 @@ virtme-ng is based on virtme, written by Andy Lutomirski <luto@kernel.org>.
     )
 
     parser.add_argument(
+        "--jobs",
+        "-j",
+        metavar="N",
+        type=int,
+        help="Build kernel with N jobs (default: number of cores).",
+    )
+
+    parser.add_argument(
         "--build-host",
         action="store",
         help="Perform kernel build on a remote server (ssh access required)",
@@ -661,7 +669,7 @@ REMOTE_BUILD_SCRIPT = """#!/bin/bash
 cd ~/.virtme
 git reset --hard __virtme__
 [ -f debian/rules ] && fakeroot debian/rules clean
-{} {}
+{} {} -j{}
 """
 
 
@@ -812,7 +820,8 @@ class KernelSource:
             tmp.write(
                 REMOTE_BUILD_SCRIPT.format(
                     args.build_host_exec_prefix or "",
-                    shlex.join(make_command) + " -j$(nproc --all)",
+                    shlex.join(make_command),
+                    args.jobs if args.jobs else "$(nproc --all)",
                 )
             )
             tmp.flush()
@@ -901,7 +910,7 @@ class KernelSource:
         make_command += args.envs
         if args.build_host is None:
             # Build the kernel locally
-            make_command += ["-j", self.cpus]
+            make_command += ["-j", str(args.jobs) if args.jobs else self.cpus]
             if args.verbose:
                 print(f"cmd: {shlex.join(make_command)}")
             check_call_cmd(make_command, quiet=not args.verbose, dry_run=args.dry_run)


arighi (commented):
Why do we need two separate options? Can we just use `--jobs N` both for local and remote builds? Thanks.

mmz-zmm:
> Why do we need two separate options? Can we just use `--jobs N` both for local and remote builds? Thanks.

Hi arighi, I have read the code in `_make_remote()` in run.py, it seams that there are 3 steps:
1. run `make -j$(nproc --all)` in remote host to get vmlinux, bzImage and modules
2. use rsync to copy them back to local host
3.  run `make -j self.cpus modules_prepare` in local host

Usually, the remote host has more core(in step 1) than local host(in step 3), so single --jobs may be appropriate for remote host, but too larger than the local host cpu nums, in which case , our local host would be too busy to do other daily work.

matttbe:
> 3\. run `make -j self.cpus modules_prepare` in local host

I don't think `modules_prepare` is doing a lot of things, and it might even take a single job. So no need to modify that one I think. (and probably `-j` could be removed, but to be tested)

mmz-zmm:
> > 3. run `make -j self.cpus modules_prepare` in local host
> 
> I don't think `modules_prepare` is doing a lot of things, and it might even take a single job. So no need to modify that one I think. (and probably `-j` could be removed, but to be tested)

Hi matttbe, sounds really reasonable, I will make a new patch and test it by removing -j in step 3.

mmz-zmm:
Hi @matttbe @arighi, matttbe is right, make modules_prepare just generates headers and prepare dirs, the build part is very limited, so just one --jobs options is enough. I have pushed a new version, glad to hear your feedbacks.

virtme_ng/run.py
```diff
@@ -859,7 +876,7 @@ def _make_remote(self, args, make_command):
                     ["fakeroot", "debian/rules", "clean"], quiet=not args.verbose
                 )
             check_call_cmd(
-                make_command + ["-j", self.cpus, "modules_prepare"],
+                make_command + ["-j", self._get_jobs(args), "modules_prepare"],
```
matttbe: I wonder if this one needs to be modified: here `args.jobs` will be specified for the remote server.

Maybe here `-j` is not needed? I currently don't have the setup to test, but after the `rsync`, what's left to do?
If there are still work that can be done in multiple jobs, probably best to keep `self.cpus` here.

mmz-zmm: After the rsync, vng run `make modules_prepare` on the local machine, I have test it with my PC(8 core) with a remote build-host, the log is:

```text
vng -bv --build-host root@xx.xx.xx.xx -j 30 --build-host-vmlinux
...
virt/lib/modules.order
0 100%    0.00kB/s    0:00:00 (xfr#670, to-chk=0/7770)
Authorized users only. All activities may be monitored and reported.
UPD     include/config/kernel.release
UPD     include/generated/compile.h
DESCEND objtool
DESCEND bpf/resolve_btfids
CC      arch/x86/kernel/asm-offsets.s
INSTALL libsubcmd_headers
CC      kernel/sched/rq-offsets.s
CALL    scripts/checksyscalls.sh
INSTALL libsubcmd_headers
```
So the modules_prepare just upate files, do small part of compilation and install headers.  Under --jobs 30, even at local machine, make -j 30 will not exceed the max 8 core. For these part of work, single jobs or multiple jobs doesn't make too much difference(2 secs vs 6 secs).

To avoid the confusion, I would follow your advice and use self.cpus here. Thanks.

virtme_ng/run.py
```diff
@@ -477,6 +471,15 @@ def make_parser():
         " or bundled together: --qemu-opts='...'",
     )
 
+    parser.add_argument(
+        "--jobs",
+        "-j",
+        metavar="N",
+        type=int,
+        help="Build kernel with N jobs (or remotely if used with --build-host)"
+        ", used by make command.",
```
matttbe: I think here, we can reduce the text and have:

    Build kernel with N jobs (default: number of cores).

WDYT?

virtme_ng/run.py
```diff
@@ -814,11 +817,16 @@ def _make_remote(self, args, make_command):
...
+            remote_j = f" -j{args.jobs}"
+        else:
+            remote_j = " -j$(nproc --all)"
+
         with tempfile.NamedTemporaryFile(mode="w+t") as tmp:
             tmp.write(
                 REMOTE_BUILD_SCRIPT.format(
                     args.build_host_exec_prefix or "",
-                    shlex.join(make_command) + " -j$(nproc --all)",
+                    shlex.join(make_command) + remote_j,
```
matttbe: If you modify [`REMOTE_BUILD_SCRIPT`](https://github.com/mmz-zmm/virtme-ng/blob/59f9a8d156cdd8c5e6565e75606067c7101e4b1e/virtme_ng/run.py#L669-L674) to have this at the last line: `{} {} -j{}`, then you could simplify this by writing:

```suggestion
        with tempfile.NamedTemporaryFile(mode="w+t") as tmp:
            tmp.write(
                REMOTE_BUILD_SCRIPT.format(
                    args.build_host_exec_prefix or "",
                    shlex.join(make_command),
                    args.jobs if args.jobs else "$(nproc --all)",
```

WDYT?

matttbe (changes requested):
Thank you for the new version. Just two small suggestions, then I think we are good.

virtme_ng/run.py
mmz-zmm: Do you mean we shall support `vng -j` form command? or just a comment about the sutiation when without -j option?  currently, if used with `-j`, user has to specify a N value, or it vill trigger error.

virtme_ng/run.py
mmz-zmm: It's better, I will update it.

virtme_ng/run.py
matttbe: I meant having this instead of the two lines you have:

```py
help="Build kernel with N jobs (default: number of cores).",
```

matttbe:
```suggestion
        help="Build kernel with N jobs (default: number of cores).",
```

mmz-zmm: Got it, will update it.

virtme_ng/run.py
```diff
@@ -818,7 +820,8 @@ def _make_remote(self, args, make_command):
             tmp.write(
                 REMOTE_BUILD_SCRIPT.format(
                     args.build_host_exec_prefix or "",
-                    shlex.join(make_command) + " -j$(nproc --all)",
+                    shlex.join(make_command),
+                    str(args.jobs) if args.jobs else "$(nproc --all)",
```
matttbe:
```suggestion
                    args.jobs if args.jobs else "$(nproc --all)",
```

Small detail: do you need the conversion to `str()` with `'{}'.format()`?

mmz-zmm: My bad, sorry about my poor python usage.;)

matttbe (approved):
Thank you for the last version, it looks good to me.

PR #467, opened 2026-04-30
README: disk: `virtio-blk` is used by `vng --disk`

matttbe:
Which might be a bit confusing, because `virtio-scsi` is used by `virtme-run --disk`.


arighi (approved):
Makes sense, thanks!

PR #469, opened 2026-05-11
virtme-ng: Fix vCPU pinning QMP connect timing out

arighi:
When `vng --pin` is used together with `--mods=auto` (default), a separate thread is spawned that immediately begins retrying the QMP connection to the QEMU/KVM instance, with a budget of five retries performed once per second.

The underlying virtme-run script, however, runs virtme-prep-kdir-mods (install kmods in .virtme_mods + depmod) before launching the QEMU/KVM instance. On large module trees this preparation can easily exceed the 5-second window: the QMP socket isn't open yet and pinning fails with `WARNING: Failed to pin vCPUs`.

Rather than extending the QMP retry budget (which would also delay real failures), gate the QMP connection attempt on a deterministic filesystem signal: virtme-prep-kdir-mods deletes modules.dep at the start and depmod recreates it at the end; this is the same staleness signal that virtme/commands/run.py already uses to decide whether to refresh the .virtme_mods directory.

Fix this introducing _kmods_prep_will_run() that mirrors the conditions used by virtme-run to decide whether prep will actually run, and a waiter _wait_for_kmods_ready() that polls modules.dep until it exists and is at least as fresh as modules.order. The polling runs in the main thread so it snapshots the staleness state before virtme-run has a chance to delete modules.dep; the waiter runs in the affinity worker so the foreground subprocess isn't blocked.

When prep isn't going to run (--skip-modules, --kimg / args.run set, missing modules.order, etc.) the worker falls through to the existing QMP retry with no added latency.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -10,15 +10,18 @@ import atexit
 import errno
 import fcntl
 import itertools
+import json
 import os
 import platform
 import re
 import shlex
 import signal
+import socket
 import subprocess
 import sys
 import tempfile
 import termios
+import threading
 from base64 import b64encode
 from pathlib import Path
 from shutil import which
@@ -139,6 +142,21 @@ def make_parser() -> "VirtmeArgumentParser":
         const=3636,
         metavar="PORT",
         help="Enable QMP (QEMU Machine Protocol) support on TCP",
+    )
+    g.add_argument(
+        "--pin",
+        "-P",
+        nargs="?",
+        const="all",
+        metavar="CPULIST",
+        help=(
+            "Pin each QEMU vCPU thread to a distinct host CPU. "
+            "If given without arguments, all host CPUs are used. "
+            "If given with an argument, it must be a list/range of CPUs "
+            "(e.g., 0,1,3-5,9). "
+            "NOTE: only a single guest can be pinned at a time, "
+            "pinning multiple guests is not supported yet."
+        ),
     )
     g.add_argument(
         "--graphics",
@@ -1805,6 +1823,11 @@ def do_it() -> int:
         qemuargs.extend(["-gdb", f"tcp:localhost:{args.gdb_server}"])
         kernelargs.extend(["-a", "nokaslr"])
 
+    # --pin requires QMP; auto-enable on the default port if the user didn't
+    # request one explicitly.
+    if args.pin and args.qmp is None:
+        args.qmp = 3636
+
     if args.qmp is not None:
         qemuargs.extend(["-qmp", f"tcp:localhost:{args.qmp},server,nowait"])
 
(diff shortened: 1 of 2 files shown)


virtme_ng/run.py
```diff
@@ -1659,11 +1659,69 @@ def pin_vcpus(self, args):
...
+        """Block until virtme-prep-kdir-mods has finished refreshing .virtme_mods.
+
+        Uses the same staleness signal as virtme/commands/run.py: depmod (re)creates
+        modules.dep at the end of the prep step, after virtme-prep-kdir-mods has
+        deleted it at the start.
+        """
+        modules_order = os.path.join(kdir, "modules.order")
+        marker = os.path.join(kdir, ".virtme_mods/lib/modules/0.0.0/modules.dep")
+        deadline = time.monotonic() + timeout
+        while time.monotonic() < deadline:
```
matttbe: out of curiosity, with a timeout of 5 minutes by default, if there is an issue to build this file and the `virtme` process ends, will this thread be cancelled at some points?

matttbe (commented):
Good catch, thank you for fixing this!

Instead of mirroring the conditions, why not moving the code to `virtme/commands/run.py`? I guess it makes sense to keep `virtme-ng` as a wrapper to `virtme` to start a VM, and this fix still looks like a workaround. (I wonder if `--pin` is not even the last option linked to the start of the VM that is still on `virtme-ng` side.)

Otherwise, the fix looks good to me, just a bit long :)

(Also, `ruff` complained: `pre-commit.ci autofix` can help)

arighi:
> Good catch, thank you for fixing this!
> 
> Instead of mirroring the conditions, why not moving the code to `virtme/commands/run.py`? I guess it makes sense to keep `virtme-ng` as a wrapper to `virtme` to start a VM, and this fix still looks like a workaround. (I wonder if `--pin` is not even the last option linked to the start of the VM that is still on `virtme-ng` side.)

Yeah, that's a good idea, I'll move the logic into virtme/commands/run.py. Moving the logic there can also fix the thread being cancelled issue.

> 
> Otherwise, the fix looks good to me, just a bit long :)
> 
> (Also, `ruff` complained: `pre-commit.ci autofix` can help)

Will fix that.

Thanks!

virtme/commands/run.py
```diff
@@ -1330,6 +1348,11 @@ def ssh_server(args, arch, qemuargs, kernelargs):
 def do_it() -> int:
     args = _ARGPARSER.parse_args()
 
+    # --pin requires QMP; auto-enable on the default port if the user didn't
+    # request one explicitly.
+    if args.pin and args.qmp is None:
```
matttbe: (I would have moved this check down, where `args.qmp` is first used, line 1830, but it's fine here if you prefer)

matttbe (approved):
Thank you for the new version, it looks less hackish (to me)!

I have one comment, but it is not blocking, feel free to merge with or without addressing this comment :)

virtme/commands/run.py
arighi: Makes sense, I'll move it down below. Thanks!

PR #470, opened 2026-05-11
README: use the `console` language to present examples

matttbe:
I recently discovered [this mode](https://github.com/github-linguist/linguist/blob/e535c9adf5306132e9df0b75ffe1ce2679873fe8/lib/linguist/languages.yml#L7269-L7280) that is specially there to better present shell sessions, where commands are prefixed with $, #, etc.

The previous 'shell' mode is to present shell scripts, where these prefixes are not used, and thus the colouring can appear a bit random. Here with 'console', lines not prefixed with such characters are seen as "output" and they appear in a different colour, making it easier to spot commands vs output.

In the Quick start section, there was a small indentation that was not needed, and made all the lines appears as "output".

See the diff: [original](https://github.com/arighi/virtme-ng#what-is-virtme-ng) - [new](https://github.com/matttbe/virtme-ng/tree/readme-console#what-is-virtme-ng)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,29 +29,29 @@ virtme-ng is based on virtme, written by Andy Lutomirski <luto@kernel.org>
 Quick start
 ===========
 
-```shell
- $ uname -r
- 5.19.0-23-generic
- $ git clone git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git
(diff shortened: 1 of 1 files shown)


arighi (approved):
Yeah, that looks nicer. Thanks.

PR #471, opened 2026-05-20
virtiofsd: use `--cache` instead of the deprecated version

matttbe:
I noticed this issue when looking at system logs:

    Use of deprecated option format '-o': Please specify options without it (e.g., '--cache auto' instead of '-o cache=auto')

The `--cache` option is available since [v1.0.0](https://gitlab.com/virtio-fs/virtiofsd/-/commit/0e73286762eb) (Dec 2021), and the warning has been added in [v1.2.0](https://gitlab.com/virtio-fs/virtiofsd/-/commit/8074419d0c71) (May 2022).


arighi (approved):
Looks good to me, thanks!

PR #472, opened 2026-05-20
vng, run: use virtiofs for internal helper exports

ralflici:
While testing kernel modules in distro root filesystems with virtme-ng, I hit a boot failure with AlmaLinux/RHEL-style kernels: the rootfs itself could be exposed without 9p, but virtme-ng still tried to mount its guest tools through 9p before starting `virtme-init`.

Those kernels do not provide 9p support, so the guest panicked before userspace could run, even though the root filesystem transport was otherwise usable.

This PR adds an explicit `--root-guesttools[=PATH]` option for external-root users. It copies virtme’s guest tools into the rootfs and runs them from there, avoiding the extra 9p guesttools mount while leaving the current default behavior unchanged.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -879,22 +879,43 @@ class VirtioFSConfig:
         self,
         path: str,
         mount_tag: str,
-        guest_tools_path=None,
-        memory=None,
         rw=False,
         posix_acl=False,
     ):
         self.path = path
         self.mount_tag = mount_tag
-        self.guest_tools_path = guest_tools_path
-        self.memory = memory
         self.rw = rw
         self.posix_acl = posix_acl
+
+
+class VirtioFSState:
+    def __init__(self, guest_tools_path, memory=None, memory_configured=False):
+        self.guest_tools_path = guest_tools_path
+        self.memory = memory
+        self.memory_configured = memory_configured
+
+
+def ensure_virtiofs_memory(
+    arch: architectures.Arch,
+    qemuargs: list[str],
+    state: VirtioFSState,
+) -> None:
+    if state.memory_configured:
+        return
+
+    memory = state.memory if state.memory is not None else "128M"
+    qemuargs.extend(["-object", f"memory-backend-memfd,id=mem,size={memory},share=on"])
+    if arch.numa_support():
+        qemuargs.extend(["-numa", "node,memdev=mem"])
+    else:
+        qemuargs.extend(["-machine", "memory-backend=mem"])
+    state.memory_configured = True
 
 
 def export_virtiofs(
     arch: architectures.Arch,
     qemuargs: list[str],
+    state: VirtioFSState,
     config: VirtioFSConfig,
     verbose=False,
 ) -> bool:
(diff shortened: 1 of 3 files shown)


matttbe:
Hello,

Thank you for this patch. The modification looks good to me, but I wonder if we cannot do better.

Small detail: usually, we try not to modify the external rootfs. Here, it should be fine because the option is explicit. But maybe that can be a source of issues: "polluting the rootfs"?

After this modification, 9p can still be used in other places, for some options, e.g. for the cache, `--console`, busybox, extra mount points, etc. Would it not be OK to stop having a mix of 9p and virtiofs, and only use virtiofs instead, except if `--force-9p` is used? I think multiple virtiofs can be used with different tags. Would it not work for you?

ralflici:
Thanks, that makes sense. The current patch was intentionally narrow, it simply fixes the RHEL/no-9p case we hit by avoiding the guesttools 9p mount for external rootfs boots.

Initially I considered a broader and more systematic approach that would avoid 9p completely unless forced, but I was a bit intimidated by the amount of hard-coded 9p mounts and the fact that I didn't fully understand all of them. So I need to look at this a bit more carefully, because cache, guesttools, busybox, and initmounts currently all go through the 9p path, and the existing virtiofs helper also wires shared memory/NUMA state for the rootfs export.

I'll give it a try!

ralflici:
Hello,

I dropped the `--root-guesttools` idea and reworked the branch to address your suggestion about avoiding a 9p/virtiofs mix.

The updated series keeps the rootfs virtiofs path unchanged, factors the shared memory setup so multiple virtiofs exports can reuse it, and switches the internal helper exports to virtiofs when 9p is not forced. This now covers guesttools, cache, busybox, initmounts, and vsockmount. `--force-9p` still preserves the old 9p behavior.

ralflici:
Pylint was unhappy with the number of positional arguments in two helper call sites, so I updated the branch to silence that locally with targeted `R0917` suppressions.

arighi (approved):
We should update the PR description at the top, since we are not introducing a --root-guesttools option anymore, but we are creating multiple virtiofs exports now (which is a much better approach).

Apart from the description, this looks good to me. Glad to see that we're moving towards virtiofs and we can potentially get rid of 9p in the future.

One thing I noticed: it looks like virtiofsd has a `--readonly` option now (old versions didn't have it). I think we should honor that otherwise a read-write remount inside the guest could get write access to the virtiofs exported dirs. We can address this in a separate PR.

ralflici:
Thanks for merging!

> We should update the PR description at the top, since we are not introducing a --root-guesttools option anymore, but we are creating multiple virtiofs exports now (which is a much better approach).

Sure, will do.

> One thing I noticed: it looks like virtiofsd has a `--readonly` option now (old versions didn't have it). I think we should honor that otherwise a read-write remount inside the guest could get write access to the virtiofs exported dirs. We can address this in a separate PR.

Yes, I noticed that too, but I didn't include it here for two reasons: it felt a bit out of scope for this patch, and using `--readonly` unconditionally would break compatibility with older virtiofsd versions. I think the approach in #479, where support is probed before passing the option, is the right way to handle it.

arighi:
> > One thing I noticed: it looks like virtiofsd has a `--readonly` option now (old versions didn't have it). I think we should honor that otherwise a read-write remount inside the guest could get write access to the virtiofs exported dirs. We can address this in a separate PR.
> 
> Yes, I noticed that too, but I didn't include it here for two reasons: it felt a bit out of scope for this patch, and using `--readonly` unconditionally would break compatibility with older virtiofsd versions. I think the approach in #479, where support is probed before passing the option, is the right way to handle it.

Thanks for checking!

PR #474, opened 2026-06-02
virtme-ng: Add Debian cloud image sandbox for kernel testing

yskzalloc:
When testing kernel changes that require a full userspace (working `modprobe`, systemd, networking, package manager), the standard virtme-ng approach of overlaying the host filesystem falls short. Filesystem modules (ksmbd, cifs, NFS), network subsystem changes, and tests that need `apt install` (kselftest deps, xfstests, smbtorture) all need a real root filesystem with a properly installed kernel.

This patch series adds `--debian`, which downloads a Debian Sid cloud image, builds the kernel as a `.deb` package, installs it into the VM via GRUB, and boots it — giving you a complete isolated system where `modprobe`, `apt`, and all kernel facilities work out of the box.

The workflow is two commands:

```
$ vng --build --debian --resize 20G   # one-time: download image, build, install
$ vng --debian                         # boot with custom kernel
```

The VM mounts the kernel source tree at `/home/debian/vng/` via 9P, so test scripts in the tree are immediately available inside the VM without copying.

**Design decisions:**

- GRUB-based boot (not `-kernel`/`-initrd`) so modules installed by `dpkg` work via standard `modprobe` paths
- All artifacts stored in `<kernel_dir>/.virtme_debian/` for per-tree isolation (multiple kernel trees can each have their own VM)
- Host architecture auto-detected (`platform.machine()`) to select the correct cloud image (amd64, arm64, ppc64el, riscv64, s390x)
- Reuses existing `--graphics`, `--sound`, `--ssh` flags rather than introducing new ones
- Port 2222 checked before VM launch to give a clear error if a stale VM is still running
- `--verbose` shows the VM serial console during build/install steps for debugging cloud-init or dpkg failures

**New flags:**

| Flag | Description |
|------|-------------|
| `--debian` | Use Debian cloud image for build/run |
| `--resize SIZE` | Set VM disk size (e.g. `20G`) |
| `--target PATH` | Use a custom qcow2 image |

**Run modes with `--debian`:**

| Command | Behavior |
|---------|----------|
| `vng --debian` | Serial console (Ctrl-A X to quit) |
| `vng --debian --ssh` | Boot background, open SSH session |
| `vng --debian --graphics` | QEMU GTK window |
| `vng --debian --exec ./test.sh` | Run script, auto-shutdown |

**Testing:**

- Built and installed kernel 7.1.0-rc6+ into Debian Sid VM
- Verified `modprobe ksmbd` and `modprobe cifs` work with GRUB-installed kernel
- Verified `--verbose` shows full boot log during install phase
- Verified port-in-use check triggers when a VM is already running

### DIFF
--- a/virtme_ng/debian.py
+++ b/virtme_ng/debian.py
@@ -1,0 +1,595 @@
+# -*- mode: python -*-
+# Copyright 2024 virtme-ng contributors
+"""virtme-ng: Debian cloud image integration."""
+
+import glob
+import os
+import shlex
+import shutil
+import sys
+import time
+from pathlib import Path
+from subprocess import (
+    DEVNULL,
+    PIPE,
+    CalledProcessError,
+    Popen,
+    check_call,
+    check_output,
+)
+
+
+def _get_run_helpers():
+    """Lazy import to avoid circular dependency."""
+    from virtme_ng.run import arg_fail, check_call_cmd
+
+    return arg_fail, check_call_cmd
+
+
+def arg_fail(message, show_usage=True):
+    """Proxy to run.arg_fail (avoids circular import at module load)."""
+    _af, _ = _get_run_helpers()
+    _af(message, show_usage=show_usage)
+
+
+def check_call_cmd(command, quiet=False, dry_run=False, pass_stdin=False):
+    """Proxy to run.check_call_cmd."""
+    _, _ccc = _get_run_helpers()
+    _ccc(command, quiet=quiet, dry_run=dry_run, pass_stdin=pass_stdin)
+
+
+DEBIAN_IMAGE_NAME = "debian-sid-generic-amd64-daily"
+DEBIAN_IMAGE_URL = (
+    f"https://cloud.debian.org/images/cloud/sid/daily/latest/{DEBIAN_IMAGE_NAME}.qcow2"
+)
+SSH_PORT = 2222
+
+
+def get_ssh_key_path():
+    """Return SSH key path inside .virtme_debian/."""
+    return os.path.join(get_debian_dir(), "id_ed25519")
+
+
+def check_port_available(port):
+    """Check if TCP port is available. Kill stale QEMU using same image if needed."""
+    import socket as sock
+
+    s = sock.socket(sock.AF_INET, sock.SOCK_STREAM)
+    try:
+        s.bind(("127.0.0.1", port))
+        s.close()
+        return True
+    except OSError:
+        s.close()
+        return False
+
+
+def get_debian_dir():
+    """Return <kernel_dir>/.virtme_debian as the artifact directory."""
+    return os.path.join(os.getcwd(), ".virtme_debian")
+
+
+class DebianMixin:
+    """Mixin providing --debian functionality for KernelSource."""
+
+    def _debian_build(self, args):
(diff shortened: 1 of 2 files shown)


yskzalloc:
Currently, this feature prompts for a password when logging in via the TTY console. I will address this in a separate commit to remove the password requirement.

rbmarliere (changes requested):
There's a default download of ubuntu when specifying `--root <DIR>` and `$DIR` does not exist. Perhaps this mechanism should be expanded, instead? IMHO adding a new `--debian` is not advisable.

arighi:
I agree with @rbmarliere. If we believe that switching from Ubuntu images to Debian images provides a clear benefit, I'm happy to consider doing that. But I wouldn't support multiple distros. Even supporting Ubuntu cloud images is somewhat debatable, since users can easily download and extract these images by themselves. The auto-create rootfs feature is mainly provided as a convenience rather than a necessity. Let's keep in mind that virtme-ng is a kernel development tool, not a virtualization orchestrator (threre are better tools for this).

yskzalloc:
Yes, I agree with the @rbmarliere @arighi opinion. I think it should be supported the other tool I’ve been working on [debsb](https://pypi.org/project/debsb/). Thank you for the review!

rbmarliere:
By the way you can use [debootstrap](https://wiki.debian.org/Debootstrap) to create any Debian rootfs, including foreign arches, and use the output with vng `--root`. I use that quite often and it works amazingly well.

yskzalloc:
Thank you @rbmarliere for the great tips! The idea for this came about because I wanted to take a custom kernel built via `vng --build` with `binddeb-pkg` and install it inside the guest with `.deb package`, instead of using ` -kernel` and `-initrd` approach.

rbmarliere:
In that case wouldn't it be simpler to just:

```bash
cd $kernel_src
vng --kconfig
# scripts/kconfig/merge_config.sh .config tools/testing/selftests/whatever_config # optional, just illustrative
vng --build
vng -r . --root $DEBOOSTRAP_OUTPUT # --rw
```

But if you're actively testing bindeb-pkg targets then that can get tricky indeed.

matttbe:
> By the way you can use [debootstrap](https://wiki.debian.org/Debootstrap) to create any Debian rootfs, including foreign arches, and use the output with vng `--root`. I use that quite often and it works amazingly well.

@rbmarliere Is it how SUSE is being validated? :-D

rbmarliere:
> @rbmarliere Is it how SUSE is being validated? :-D

Our choice for the UltimateQEMUWraper(tm) for testing/validation is still openQA, both for SUSE and [openSUSE](https://openqa.opensuse.org). But I know at least one instance of `vng` usage by @marcosps ! Personally when I have more intricate test scenarios to run locally I've been making use of [k-lab](https://github.com/rbmarliere/k-lab) (because everyone seems to have their own harness these LLM days, why not 😅)?

marcosps:
@matttbe as @rbmarliere mentioned, we are trying to switch over to virtme-ng to test livepatching at SUSE, and we prepare a rootfs per SLE version (SLE12-SP5, SLE15-SP5, and so on), so we are using it like @rbmarliere described it.

For now we still rely in the old method, our own qemu scripts, but the idea is move to virtme-ng now that it support `--systemd`, which would be a good improvement in our testing!

PR #479, opened 2026-06-27
virtme: Enforce read-only helper exports with virtiofsd

arighi:
Mounting a virtio-fs helper export with -o ro only applies a guest-side restriction. A privileged guest can remount it read-write and gain write access to the host export.

Use virtiofsd --readonly for read-only helper exports when available. Older virtiofsd releases lack this option, so probe once per vng invocation to check when to use this flag.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -760,6 +760,9 @@ def find_kernel_and_mods(arch, args) -> Kernel:
     return kernel
 
 
+_VIRTIOFSD_READONLY_SUPPORT: dict[str, bool | None] = {"value": None}
+
+
 class VirtioFS:
     def __init__(self, guest_tools_path):
         self.sock = None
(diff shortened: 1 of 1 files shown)


arighi:
cc: @ralflici

ralflici:
Hi,

The compatibility strategy looks good to me; it should correctly gate `--readonly` usage to virtiofsd binaries that support it.

One thing I noticed, though, is that the ROOTFS virtiofs export does not get host-side read-only enforcement. The commit message talks about helper exports, so this may be outside the intended scope of this patch, but I think the same security argument may also apply to the rootfs export.

Taking a step back, maybe the confusion comes from the fact that `VirtioFSConfig` now carries both `rw` and `readonly` as separate booleans with opposite meanings. In `export_hostfs()` they are always set as exact inverses, but the ROOTFS path sets only `rw=args.rw`, which makes it easy for the two concepts to diverge.

arighi:
> One thing I noticed, though, is that the ROOTFS virtiofs export does not get host-side read-only enforcement. The commit message talks about helper exports, so this may be outside the intended scope of this patch, but I think the same security argument may also apply to the rootfs export.
> 
> Taking a step back, maybe the confusion comes from the fact that `VirtioFSConfig` now carries both `rw` and `readonly` as separate booleans with opposite meanings. In `export_hostfs()` they are always set as exact inverses, but the ROOTFS path sets only `rw=args.rw`, which makes it easy for the two concepts to diverge.

Good catch! We should definitely enforce readonly both for ROOTFS and the helper exports. I've updated the PR, now it should be consistent for all the exports.

arighi:
I've been testing this on multiple platforms and I haven't seen any regression. It looks like a reasonable fix, so I'm merging this, if there are issues we can address/fix them in tree.

PR #480, opened 2026-06-28
Support running on a NixOS host

nikableh:
## What

This makes `vng` (default snapshot mode) boot to a working shell when the host — and therefore the guest snapshot — is NixOS. Today it kernel-panics during init and never reaches a shell.

## Why it fails on NixOS

NixOS isn't FHS: programs don't live in `/bin` or `/usr/bin`, they're in `/nix/store` and exposed through the system profile at `/run/current-system/sw/bin`. `virtme-ng-init` assumes an FHS guest, so on a NixOS snapshot it can't find `stty`, `ip`, `su`, or the login shell, and a missing tool aborts PID 1 (`Kernel panic - Attempted to kill init!`).

## Changes

Four small, independent commits:

1. **use portable `#!/usr/bin/env bash` shebang** — guest helper scripts (and the scripts generated by `run.py`/`mcp.py`) hardcoded `#!/bin/bash`, which doesn't exist on NixOS. `env bash` resolves via `PATH` and still works on FHS guests.

2. **vng-init: don't panic if stty is unavailable** — `configure_terminal()` unwrapped the result of running `stty`, so a missing `stty` killed init. Terminal sizing is cosmetic; log a warning and continue, matching how `run_cmd()` already treats other optional tools.

3. **vng-init: preserve NixOS /run symlinks across the tmpfs mount** — init mounts a fresh tmpfs on `/run`, which hides `/run/current-system` (a symlink into the Nix store that the whole userland depends on). Save the targets of `/run/current-system` and `/run/booted-system` before the mount and recreate them after. No-op on guests that don't have them.

4. **vng-init: add the NixOS system profile to the default PATH** — add `/run/current-system/sw/bin` to the init's default `PATH` (it doesn't exist on other distros, so it's harmless there) and append the defaults to any inherited `PATH` instead of overwriting it. This is where `stty`, `ip`, `su`, and the login shell live on NixOS.

## Result

On a NixOS host, `vng` now boots straight to a normal shell — no `--shell` or other workarounds — while non-NixOS guests are unaffected (the added path simply doesn't exist, the symlinks aren't there to save, and the shebang/stty changes are pure improvements).

Tested on a NixOS host booting a v7.0 kernel: clean boot to an interactive shell with `stty`/`ip`/`su` all resolving. `cargo fmt`/`clippy` clean.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -6,7 +6,8 @@
 # 8177f97513213526df2cf6184d8ff986c675afb514d4e68a404010521b880643
 
 configure_environment() {
-    export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
+    local defaults=/run/current-system/sw/bin:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
+    export PATH="${PATH:+$PATH:}$defaults"
 }
 
 log() {
@@ -25,8 +26,33 @@ mount_kernel_filesystems() {
     if [[ $$ -eq 1 ]]; then
         # only mount /run if systemd is not the init
         if ! grep -q "run /run" /proc/mounts; then
+            # The fresh tmpfs hides whatever the rootfs kept under /run. On NixOS
+            # the userland is reached through /run/current-system (a symlink into
+            # the Nix store), so save and restore it across the mount; without it
+            # PATH, the login shell, etc. break.
+            local link target ln_cmd=""
+            local -a preserve=()
+            for link in /run/current-system /run/booted-system; do
+                if target=$(readlink "$link" 2> /dev/null); then
+                    preserve+=("$link" "$target")
+                fi
+            done
+            if ((${#preserve[@]})); then
+                # `ln` lives in /run/current-system/sw/bin on NixOS, which the
+                # mount hides. Resolve its directory out of /run but keep the
+                # `ln` name, so a coreutils multi-call binary still acts as ln.
+                ln_cmd=$(command -v ln)
+                ln_cmd="$(readlink -f "${ln_cmd%/*}")/${ln_cmd##*/}"
+            fi
+
             # Mount tmpfs dirs
             mount -t tmpfs -o mode=0755 run /run/
+
+            local i
+            for ((i = 0; i < ${#preserve[@]}; i += 2)); do
+                "$ln_cmd" -s -- "${preserve[i + 1]}" "${preserve[i]}" ||
+                    log "WARNING: failed to restore symlink ${preserve[i]}"
+            done
         fi
     fi
 
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # virtme-init: virtme's basic init (PID 1) process
 # Copyright © 2014 Andy Lutomirski
 # Licensed under the GPLv2, which is available in the virtme distribution
--- a/virtme_ng_init/src/main.rs
+++ b/virtme_ng_init/src/main.rs
@@ -464,6 +464,25 @@ fn symlink_fds() {
     }
 }
 
+// Symlinks under /run that must survive the tmpfs mount on /run. On NixOS the whole userland is
+// reached through /run/current-system, so losing it breaks PATH, the login shell, etc.
+const PRESERVED_RUN_SYMLINKS: &[&str] = &["/run/current-system", "/run/booted-system"];
+
+fn save_run_symlinks() -> Vec<(&'static str, PathBuf)> {
+    PRESERVED_RUN_SYMLINKS
+        .iter()
+        .filter_map(|link| std::fs::read_link(link).ok().map(|target| (*link, target)))
+        .collect()
+}
+
+fn restore_run_symlinks(links: &[(&'static str, PathBuf)]) {
+    for (link, target) in links {
+        if let Err(e) = std::os::unix::fs::symlink(target, link) {
+            log!("WARNING: failed to restore symlink {link}: {e}");
+        }
+    }
+}
+
 fn mount_kernel_filesystems() {
     for mount_info in KERNEL_MOUNTS {
         // In the case where a rootfs is specified when launching virtme-ng, it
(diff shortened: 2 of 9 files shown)


javierhonduco:
Thank you for working on making virtme-ng work on NixOS, very interested :)

virtme/guest/virtme-init
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
arighi, line 1: With this change we may regress some minimal environments that don't have `/usr/bin/env`. However, even alpine has that, so I guess the regression potential is limited. Something to think about. Maybe we can mention this in troubleshooting section in the main README.md?

virtme_ng_init/src/main.rs
```diff
@@ -229,7 +229,11 @@ fn wait_for_child(child: i32) -> Option<i32> {
 }
 
 fn configure_environment() {
-    env::set_var("PATH", "/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin");
+    let defaults = "/run/current-system/sw/bin:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin";
+    match env::var("PATH") {
+        Ok(path) if !path.is_empty() => env::set_var("PATH", format!("{path}:{defaults}")),
+        _ => env::set_var("PATH", defaults),
+    }
```
arighi, line 236: I think we need to mirror this logic also in virtme-ng-init (the bash init), otherwise `--no-virtme-ng-init` is still broken on NixOS.

Maybe add something like this on top (untested):
```
diff --git a/virtme/guest/virtme-init b/virtme/guest/virtme-init
index 9c21af1..7994ea1 100755
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -6,7 +6,13 @@
 # 8177f97513213526df2cf6184d8ff986c675afb514d4e68a404010521b880643
 
 configure_environment() {
-    export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
+    local defaults=/run/current-system/sw/bin:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
+
+    if [[ -n "${PATH:-}" ]]; then
+        export PATH="${PATH}:${defaults}"
+    else
+        export PATH="$defaults"
+    fi
 }
 
 log() {
@@ -18,6 +24,10 @@ log() {
 }
 
 mount_kernel_filesystems() {
+    local -a run_links=()
+    local -a run_targets=()
+    local link target readlink_cmd ln_cmd
+
     # Mount procfs and sysfs (needed for stat, sadly)
     mount -t proc -o nosuid,noexec,nodev proc /proc/
     mount -t sysfs -o nosuid,noexec,nodev sys /sys/
@@ -25,8 +35,30 @@ mount_kernel_filesystems() {
     if [[ $$ -eq 1 ]]; then
         # only mount /run if systemd is not the init
         if ! grep -q "run /run" /proc/mounts; then
+            # Preserve the NixOS system profiles that would otherwise be
+            # hidden by the new tmpfs. Resolve the helper binaries to their
+            # /nix/store paths while /run/current-system is still available.
+            for link in /run/current-system /run/booted-system; do
+                if [[ -L "$link" ]]; then
+                    if [[ -z "${readlink_cmd:-}" ]]; then
+                        readlink_cmd="$(command -v readlink)"
+                        readlink_cmd="$($readlink_cmd -f -- "$readlink_cmd")"
+                        ln_cmd="$($readlink_cmd -f -- "$(command -v ln)")"
+                    fi
+                    target="$($readlink_cmd -- "$link")"
+                    run_links+=("$link")
+                    run_targets+=("$target")
+                fi
+            done
+
             # Mount tmpfs dirs
-            mount -t tmpfs -o mode=0755 run /run/
+            if mount -t tmpfs -o mode=0755 run /run/; then
+                for i in "${!run_links[@]}"; do
+                    if ! "$ln_cmd" -s -- "${run_targets[$i]}" "${run_links[$i]}"; then
+                        log "WARNING: failed to restore ${run_links[$i]}"
+                    fi
+                done
+            fi
         fi
     fi
 
```

virtme_ng_init/src/main.rs
```diff
@@ -464,6 +468,25 @@ fn symlink_fds() {
...
+        .collect()
+}
+
+fn restore_run_symlinks(links: &[(&'static str, PathBuf)]) {
+    for (link, target) in links {
+        if let Err(e) = std::os::unix::fs::symlink(target, link) {
+            log!("WARNING: failed to restore symlink {link}: {e}");
+        }
+    }
+}
```
arighi, line 488: Ditto, see patch above.

arighi (approved):
I left a couple of comments, but overall I like this. A proper NixOS support has been requested for quite a while, thanks for looking at this @nikableh 

I'm ok to merge this and address the issue I pointed at in a later PR (the `virtme-ng-init` one), or address it in this PR. However you prefer.

virtme/guest/virtme-init
nikableh: > Maybe we can mention this in troubleshooting section in the main README.md?

Done

virtme_ng_init/src/main.rs
nikableh: Done and tested — `--no-virtme-ng-init` works on NixOS

arighi (approved):
There are probably few other things to adjust, but I think it's good overall. Let's get this merged and improve things in tree if needed. Thanks!

virtme/guest/virtme-init
matttbe: It looks like this modification breaks this init script on Ubuntu >=26.04

```console
$ ./vng -r --no-virtme-ng-init -v
  (...)
  Security violation: Requested utility `env` does not match executable name:
    /usr/libexec/virtiofsd
  [ 1.344090] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000100
  [ 1.344221] CPU: 0 UID: 0 PID: 1 Comm: virtme-init Not tainted 7.0.0-22-generic #22-Ubuntu PREEMPT(lazy)
```

(Note that `virtiofsd` is just an "artifact": if I don't use it with `--force-9p`, it will complain about `/usr/bin/qemu-system-x86_64` instead)

Switching back to `#!/bin/bash`, or using `gnuenv`, or coreutils >=0.8.0 binaries from upstream fixes the issue. It looks like this is due to an Ubuntu-specific patch, see https://github.com/uutils/coreutils/issues/10135 or https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2137745

It is not clear to me whether this patch is still required, so I asked on LP what's their plan.

For the moment, the current workaround is:

```console
$ sudo ln -sf gnuenv /usr/bin/env
## revert with: sudo ln -sf ../lib/cargo/bin/coreutils/env /usr/bin/env
```

matttbe: Edit: it looks like this currently only affects Ubuntu Devel (Ubuntu >26.04) and the patch causing that should certainly be removed. (By chance not applied on Ubuntu 26.04, probably by accident, see the comments on LP)

PR #482, opened 2026-07-10
vng, run: add root POSIX ACL control

ralflici:
This came out of testing distro rootfs boots in a GitHub Actions matrix with virtme-ng.

Some arm64 guests, notably Debian 11 and Ubuntu 20.04 in our test matrix, fail to boot when the root filesystem is exported through virtiofs with POSIX ACLs enabled. The failure happens during the rootfs mount path, after the kernel has already booted, and shows up as a broken virtiofs root followed by switch_root failing.

During local reproduction this mapped to the virtiofs client not being able to negotiate POSIX ACL support during FUSE_INIT, so this cannot be reliably probed from the host before starting the guest.

This patch keeps the current default unchanged: external virtiofs roots still use POSIX ACLs by default. It only adds an explicit opt-out (`--no-root-posix-acl`) for setups where the guest/kernel combination cannot handle that negotiationn. The option is ignored for 9p and for host-root sharing, where POSIX ACLs are already not enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -714,6 +714,12 @@ Troubleshooting
    distribution, including minimal ones such as Alpine, but may be missing from
    a hand-built `--root` chroot.
 
+ - Some guest kernels may fail to mount an external virtiofs root filesystem
+   when POSIX ACL support is enabled. This can show up as the root filesystem
+   mount apparently succeeding, followed by `switch_root` failing with
+   `Connection refused` and the guest panicking. In this case, try disabling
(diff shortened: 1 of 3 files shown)


arighi (approved):
Looks good to me, thanks.

arighi:
@ralflici can we add a comment about `--no-root-posix-acl` in the "Troubleshooting" section in the main README.md?

ralflici:
Sure! I’ll add a Troubleshooting bullet.

PR #483, opened 2026-07-10
vng, run: normalize arm64 EFI zboot images

ralflici:
Several current arm64 distro kernels are no longer plain Image files from QEMU's point of view. Fedora 44 and recent Ubuntu development releases ship EFI zboot images, and some Ubuntu kernels add another signed PE wrapper around the bootable payload. Passing those files directly through QEMU's `-kernel` either fails or hangs before the guest reaches userspace.

The practical symptom was confusing: virtme-ng appeared to be waiting for virtiofsd/rootfs setup, but the real issue was earlier: QEMU had not consumed a bootable arm64 kernel image.

This patch normalizes only arm64 kernel images before passing them to QEMU. If the selected image is a PE wrapper with a `.linux` payload, it extracts that payload. If the result is an EFI zboot image, it unwraps and decompresses it to a plain arm64 Image. The normalized Image is cached by input content hash, so the conversion is deterministic and does not repeat unnecessarily. Already-plain images and non-arm64 kernels are left untouched. Module/rootfs discovery still uses the original distro kernel path, while QEMU receives the normalized Image.

This follows the same general path modern QEMU already uses for arm64 direct kernel boot: [`load_aarch64_image()`](https://gitlab.com/qemu-project/qemu/-/blob/ab2056a0b7c944b16b224a8feabd99023772ae91/hw/arm/boot.c#L816) tries to normalize EFI zboot images before checking for the plain arm64 Image magic, and [`unpack_efi_zboot_image()`](https://gitlab.com/qemu-project/qemu/-/blob/ab2056a0b7c944b16b224a8feabd99023772ae91/hw/core/loader.c#L902) parses the Linux EFI zboot header metadata.

So QEMU already supports part of this path for arm64 direct kernel boot. However, it still leaves two practical gaps for virtme-ng users: QEMU does not unwrap an outer PE/UKI-style wrapper to extract the `.linux` payload before applying its zboot handling, and older versions do not support the zboot compression used by current distros. Doing this in virtme-ng makes the behavior deterministic.

This was tested against real distro rootfs images in CI. In particular, Fedora 44, Ubuntu 25.10, and Ubuntu 26.04 arm64 moved from failing before boot to booting far enough to compile the module successfully.

### DIFF
--- a/virtme/kernel_image.py
+++ b/virtme/kernel_image.py
@@ -57,11 +57,43 @@ def normalize_kernel_image(arch, path: str, cache_dir: Path, verbose=False) -> s
 
     image = Path(path)
     try:
+        with image.open("rb") as stream:
+            header = stream.read(ARM64_IMAGE_MAGIC_OFFSET + len(ARM64_IMAGE_MAGIC))
+    except OSError as exc:
+        raise KernelImageError(f"failed to read kernel image {image}: {exc}") from exc
+
+    # Avoid reading and hashing the complete file in the common case where QEMU
+    # can consume an already-plain arm64 Image directly.
+    if _is_arm64_image(header):
+        return path
+
+    output = _cached_image_path(cache_dir, image)
+    if output.is_file():
+        if verbose:
+            print(
+                f"virtme: using normalized arm64 kernel image {output}",
+                file=sys.stderr,
+            )
+        return str(output)
+
+    try:
         data = image.read_bytes()
     except OSError as exc:
         raise KernelImageError(f"failed to read kernel image {image}: {exc}") from exc
 
-    output = _cached_image_path(cache_dir, data)
+    # The image may have been replaced after the streaming hash above. Derive
+    # the publication path from the exact bytes that will be normalized so the
+    # cache remains content-addressed. This also notices a cache entry published
+    # by a concurrent invocation after the first lookup.
+    output = _cached_image_path_from_digest(cache_dir, hashlib.sha256(data).hexdigest())
+    if output.is_file():
+        if verbose:
+            print(
+                f"virtme: using normalized arm64 kernel image {output}",
+                file=sys.stderr,
+            )
+        return str(output)
+
     if not _prepare_arm64_image(data, image, output):
         return path
 
@@ -295,8 +327,18 @@ def _decompressed_image_too_large(image: Path) -> KernelImageError:
     )
 
 
-def _cached_image_path(cache_dir: Path, data: bytes) -> Path:
-    digest = hashlib.sha256(data).hexdigest()
+def _cached_image_path(cache_dir: Path, image: Path) -> Path:
+    digest = hashlib.sha256()
+    try:
+        with image.open("rb") as stream:
+            for chunk in iter(lambda: stream.read(1024 * 1024), b""):
+                digest.update(chunk)
+    except OSError as exc:
+        raise KernelImageError(f"failed to read kernel image {image}: {exc}") from exc
+    return _cached_image_path_from_digest(cache_dir, digest.hexdigest())
+
+
+def _cached_image_path_from_digest(cache_dir: Path, digest: str) -> Path:
     return Path(cache_dir, "kernel-images", "arm64", digest, "Image")
 
 
--- a/virtme/kernel_image.py
+++ b/virtme/kernel_image.py
@@ -1,0 +1,332 @@
+# -*- mode: python -*-
+# kernel_image: Normalize kernel images for QEMU direct boot
+# Licensed under the GPLv2, which is available in the virtme distribution
+# as a file called LICENSE with SHA-256 hash:
+# 8177f97513213526df2cf6184d8ff986c675afb514d4e68a404010521b880643
+
+"""Helpers to normalize distro kernel images before passing them to QEMU."""
+
+import gzip
+import hashlib
+import io
+import os
+import resource
+import signal
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+GZIP_MAGIC = b"\x1f\x8b"
+PE_MSDOS_MAGIC = b"MZ"
+PE_SIGNATURE = b"PE\0\0"
+ARM64_IMAGE_MAGIC = b"ARM\x64"
+ZBOOT_MAGIC = b"zimg"
+EFI_PE_LINUX_MAGIC = b"\xcd\x23\x82\x81"
+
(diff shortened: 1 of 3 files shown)


virtme/kernel_image.py
```diff
@@ -0,0 +1,266 @@
...
+    try:
+        data = image.read_bytes()
+    except OSError as exc:
+        raise KernelImageError(f"failed to read kernel image {image}: {exc}") from exc
+
+    prepared = _prepare_arm64_image(data, image)
+    if prepared is None:
+        return path
+
+    output = _cached_image_path(cache_dir, data)
```
arighi: IIUC cache is mounted read-write, so a guest could potentially overwrite the cached kernel image and next invocation will boot this new file. I'm wondering if we should split the cache into a host-only/read-only cache + a guest-writable cache for SSH and runtime files.

jimc:
On Fri, Jul 10, 2026 at 12:35 PM Andrea Righi ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In virtme/kernel_image.py
> <https://github.com/arighi/virtme-ng/pull/483#discussion_r3561178101>:
>
> > +    # QEMU's arm64 -kernel path expects the bootloader to hand it a plain Image,
> +    # but distro kernels may be wrapped as EFI zboot images.
> +    if getattr(arch, "linuxname", None) != "arm64":
> +        return path
> +
> +    image = Path(path)
> +    try:
> +        data = image.read_bytes()
> +    except OSError as exc:
> +        raise KernelImageError(f"failed to read kernel image {image}: {exc}") from exc
> +
> +    prepared = _prepare_arm64_image(data, image)
> +    if prepared is None:
> +        return path
> +
> +    output = _cached_image_path(cache_dir, data)
>
> IIUC cache is mounted read-write, so a guest could potentially overwrite
> the cached kernel image and next invocation will boot this new file. I'm
> wondering if we should split the cache into a host-on
>

Ive found myself scripting a kind of TELEMETRY output dir, to deliver
test-results to the host/harness.  Is that something to standarize or
formalise?

virtme/kernel_image.py
ralflici: Right, that’s a fair concern. Just to confirm, do you mean splitting and exporting only the guest-writeable cache at /run/virtme/cache? Or should the host cache remain visible to the guest read-only?

Unless there is a reason for the guest to access the host cache, I would prefer the first option, but I may be overlooking a use case for the read-only export.

arighi: Yes, first option. For external root filesystems, ~/.cache/virtme-ng should remain host-only and only a dedicated guest-facing subtree should be exported at /run/virtme/cache. I don't see a need to expose the host cache read-only. Then the normalized and downloaded kernels can remain host only.

arighi:
> Ive found myself scripting a kind of TELEMETRY output dir, to deliver test-results to the host/harness. Is that something to standarize or formalise?

for exporting test results, `--rwdir=guestpath=hostpath` already provides a dedicated guest-writable host directory. For example you can do:
```  
  mkdir -p "$PWD/results"
  vng --rwdir="/run/virtme/results=$PWD/results" ...
```
I think it makes sense keeping result exchange separate from the internal cache. If you think a standard result layout would be useful, maybe we can discuss this in a separate issue.

virtme/kernel_image.py
arighi: @ralflici I pushed a follow-up commit implementing the first option. sketched using AI. The host-trusted cache is no longer exported to external-root guests and each VM gets a fresh guest-writable runtime dir. Persistent SSH host keys remain host-only, while only the server keys and client public key are copied into the per-run directory. This should also prevent guest-created files from affecting later invocations. Can you check if it makes sense to you? Thanks!

ralflici: Thanks! Yes, I checked it and it looks good to me.

arighi:
I'm still conflicted about this one, I still see a couple of issues:
1) we always need a write cache, we should probably make it conditional and have something like:
```
  needs_guest_cache = args.systemd or (
      args.root != "/" and args.server == "ssh"
  )
```
2) decompression has no output-size limit, so a corrupted or malicious kernel file can potentially exhaust the entire host memory, we should run some checks / add some limitation instead of uncompressing it unconditionally.

@ralflici if you have some time can you check if there's a way to address both these issues? Otherwise I'll take a look. Thanks.

ralflici:
Yes, I took a look and I agree with both points.

I’ve folded the first suggestion into the cache commit. The second one is a little less obvious because it requires choosing a defensible output-size limit. Fortunately, QEMU already defines a [256 MiB decompression limit](https://github.com/qemu/qemu/blob/eca2c16212ef9dcb0871de39bb9d1c2efebe76be/include/hw/core/loader.h#L84) and applies it when decompressing kernel images, including EFI zboot payloads. I used the same limit since the normalized image is ultimately passed to QEMU.

I’m pushing amended versions of both commits.

arighi:
Thanks @ralflici, sorry I'm being picky, I've pushed another commit on top of yours, I think it can help avoiding some unnecessary overhead, we want to be really fast at booting, that's vng main goal. :)

Let me know if this change makes sense to you. Everything else LGTM.

ralflici:
Nice optimization indeed. I did a quick local check with a Fedora 44 ARM64 zstd kernel image: the plain Image path went from about 104 ms to effectively zero, while a cached zstd wrapper went from about 32 ms to 17 ms. The uncached normalization path remained roughly unchanged at about 220 ms. So the improvement is clearly effective.

One minor, non-blocking observation: on a cache miss, the input is hashed and then reopened with `read_bytes()`. If the file changes between those operations, the normalized content could be stored under the digest of the previous content. This seems unlikely for kernel images, so I don’t consider it a blocker, but we could rehash the in-memory data on a miss if we want to preserve the content-addressed invariant strictly.

Otherwise, the change looks good to me.

arighi:
> One minor, non-blocking observation: on a cache miss, the input is hashed and then reopened with `read_bytes()`. If the file changes between those operations, the normalized content could be stored under the digest of the previous content. This seems unlikely for kernel images, so I don’t consider it a blocker, but we could rehash the in-memory data on a miss if we want to preserve the content-addressed invariant strictly.

Ah that's actually a valid point, I've updated my commit. Can you check again? Thanks!

ralflici:
Yes, the race is gone and the change looks good to me, thanks!

arighi (approved):

PR #484, opened 2026-07-20
Introduce `--root-disk` and `--root-disk-partition`

rbmarliere:
Hello!

Just as a heads up, this PR has been written with the assistance of a LLM, but I cleaned it up and reviewed it and I think it has value.

The TLDR; is that while `--root` is quite useful, it's still limited for foreign architectures.

Therefore, instead of trying to hack my way through 9p and/or patch virtiofs[d], I chose a simpler alternative: why not simply use a block device directly? It's trivial to convert any root filesystem you have lying around into an image, anyway. It will also make other workflows easier, such as when downloading an image asset from somewhere (say, [openQA](https://openqa.opensuse.org/tests/6112711/asset/hdd/opensuse-Tumbleweed-x86_64-20260718-textmode@64bit.qcow2)) and just booting it directly with `vng` instead of going through the boring process of adding a new VM under `virsh`/`virt-manager`.

Although a bit verbose, the changes are quite straightforward. Thanks in advance for the time spent reviewing!


virtme/commands/configkernel.py
```diff
@@ -180,6 +180,16 @@ def arg_fail(message):
     "CONFIG_SCSI_VIRTIO=y",
     "CONFIG_BLK_DEV_SD=y",
     "",
+    # "##: virtio-blk support",
+    # "CONFIG_VIRTIO_BLK=y",
+    # "CONFIG_EXT4_FS=y",
+    # "CONFIG_XFS_F=y",
```
arighi: typo? CONFIG_XFS_FS

virtme/commands/configkernel.py
```diff
@@ -180,6 +180,16 @@ def arg_fail(message):
...
+    # "CONFIG_BTRFS_FS=y",
+    # "CONFIG_F2FS_FS=y",
+    # "CONFIG_PARTITION_ADVANCED=y",
+    # "CONFIG_MSDOS_PARTITION=y",
+    # "CONFIG_EFI_PARTITION=y",
+    # "",
```
arighi, line 192: These lines are all commented so a default kernel built with vng won't support --root-disk, right?

And I'm not sure that I like enabling all of these by default, it'd slow down the build time even for those that are not using --root-disk.

virtme/mkinitramfs.py
```diff
@@ -155,19 +215,40 @@ def install_modules(cw, modfiles):
...
     return out.getvalue().encode("utf-8")
 
 
 class Config:
-    __slots__ = ["modfiles", "virtme_data", "virtme_init_path", "busybox", "access"]
+    __slots__ = [
+        "modfiles",
+        "virtme_data",
+        "virtme_init_path",
+        "busybox",
```
arighi, line 238: Are we still adding the host's busybox into the initramfs? If that's the case I think we should require `--busybox` to support foreign-architecture disk roots.

arighi (commented):
I still have some concerns with this... especially adding more filesystems to the default vng build, which could potentially slow down builds for everyone, even those not using --root-disk/--root-disk-partition. I need to think more about this I guess... in the meantime I left a few comments. Thanks!

virtme/commands/configkernel.py
matttbe: I agree, it would be good not to increase this list.

It might be good to be able to split this already longish list. Maybe some arguments could be re-used, e.g. using ...

```
vng --build --graphics
```

... would enable DRM related configs. Same with `--sound` and `--snap`, plus either `--disk` or a new `--root-disk`. Not sure about `--network` as it is needed for 9P (and the loopback), nor `--debug` because the debug options might always be handy and not really heavy.

Note that on my side, I usually disable those after having called `virtme-configkernel`:

```
-d PCCARD -d MACINTOSH_DRIVERS -d SOUND -d USB_SUPPORT
-d NEW_LEDS -d SURFACE_PLATFORMS -d DRM -d FB -d ATA  
-d MISC_FILESYSTEMS -d SCSI -d AGP                    
```

(+ `-d RETPOLINE`)

(I also enable some `DEBUG_INFO*` configs + `DYNAMIC_DEBUG` and some `FTRACE` ones + some networking specific ones)

rbmarliere: Yes, I left them commented just to point out what the user might need. But I agree it shouldn't be added for a default vng build.

virtme/mkinitramfs.py
rbmarliere: I've been already using `--busybox` with `--root` at all times and making it a requirement seems OK to me.

virtme/commands/configkernel.py
arighi: I like @matttbe's suggestion to have some extra --build options to enable certain subsystems. But at the same time I'm afraid it's going to be a pain to maintain all these configs across newer and older kernels.

arighi: If we decide to add --root-disk / --root-disk-partition, maybe we should have a way to print a warning or something similar to show which config options are recommended, so we don't have to enable them or add them as comments?

rbmarliere: IMHO I think having a single minimal (perhaps even smaller than current one) is better. For example, I'm quite used of doing things such as:

```sh
vng --kconfig
scripts/kconfig/merge_config.sh .config tools/testing/selftests/bpf/{config,config.x86_64}
vng -b
```

This is no trouble at all, I think the user will know better about which Kconfig he's targeting, anyway.

rbmarliere: > If we decide to add --root-disk / --root-disk-partition, maybe we should have a way to print a warning or something similar to show which config options are recommended, so we don't have to enable them or add them as comments?

Oops, I didn't see this one when I replied. This sounds rather complex. I imagine something would check against the current Kconfig for all the vng-expected symbols, which would change depending on which vng arguments were passed? I'm not against it, but sounds like out of scope here tbh -- happy to follow up on that later though

matttbe: > I like @matttbe's suggestion to have some extra --build options to enable certain subsystems. But at the same time I'm afraid it's going to be a pain to maintain all these configs across newer and older kernels.

Just in case, there are already some sections in configkernel.py:

https://github.com/arighi/virtme-ng/blob/5d1d47c7cee0ff69f3bc0a09ab3b0011612ce28d/virtme/commands/configkernel.py#L208-L255

I guess it is just a question of splitting `_GENERIC_CONFIG_OPTIONAL` into multiple parts, and add the ones that are needed depending on the given options (+ a new options to get all of them to get back to the previous behaviour)

arighi: Yeah we can try splitting _GENERIC_CONFIG_OPTIONAL into multiple vars and include them selectively based on the "features" that we want to enable. I guess it's still going to be challenging to maintain all these sections, but we can try this way.

arighi:
After thinking more about this, I'm still concerned that this feature would add more maintenance cost than benefits.

It's true that being able to boot directly from a plain image or a root partition is certainly convenient, but supporting it properly means carrying additional initramfs logic and kernel configuration requirements for virtio-blk, partition handling, and filesystems. That makes the default configuration and the overall code path harder to maintain over time.

I think a simpler approach would be to document a recipe for mounting a disk image or its root partition on the host and then using the existing --root workflow to export that filesystem to the guest. This keeps the feature set focused and reuses a path that vng already supports well. Also let's keep in mind that vng is primarily a kernel development tool, rather than a general-purpose VM provisioning or image-booting solution.

For these reasons, I think we should just add documentation for the host-mount + --root workflow.

@rbmarliere @matttbe opinions?

matttbe:
> @matttbe opinions?

I agree with you. I didn't know what to say about this PR, because I don't know well this area (not very useful in my case for the kernel dev), and the modifications are quite big, with a custom init mount script, more config, etc. So at the end, I'm glad you sent this message :)

Instead, could we not have a separate script to do the required modifications before calling vng?

rbmarliere:
> After thinking more about this, I'm still concerned that this feature would add more maintenance cost than benefits.
> 
> It's true that being able to boot directly from a plain image or a root partition is certainly convenient, but supporting it properly means carrying additional initramfs logic and kernel configuration requirements for virtio-blk, partition handling, and filesystems. That makes the default configuration and the overall code path harder to maintain over time.

This is not done in this PR though, I agree that we should **not** add any Kconfig symbols to the existing list.

> I think a simpler approach would be to document a recipe for mounting a disk image or its root partition on the host and then using the existing --root workflow to export that filesystem to the guest. This keeps the feature set focused and reuses a path that vng already supports well.

I mentioned booting an image from openQA just as a consequence of what this PR enables, but my use case is actually the contrary: starting from a root filesystem that is laying around in my disk, I want to be able to assemble it into a disk image and boot it using `vng`. Because using `--root` with foreign architectures does not quite work consistently unless you're using 9p, which just sucks honestly (e.g. try this: `void *map = mmap(NULL, 4096, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);`). OTOH, `virtiofs` is FUSE which is better but also has limitations, and I simply could not make it work with foreign architectures.

In this workflow, the kernel is still built separately and used in --run. It's just that userspace, instead of --root, would come from a --root-disk.

> Also let's keep in mind that vng is primarily a kernel development tool, rather than a general-purpose VM provisioning or image-booting solution.

Sure, but downstream development sometimes require **testing** within an environment closer to reality, which is what virtio-blk allows IMHO.

> For these reasons, I think we should just add documentation for the host-mount + --root workflow.

Please reconsider one more time, but if this is your decision then please just close this PR :(
Thanks anyway!

rbmarliere:
I'll be sending the fixes here unrelated to the feature itself separately. Although not really mergeable in current form, my question is: would `--root-disk` be acceptable in some form? If so, what should I have in mind for next revision? @arighi @matttbe

virtme/commands/run.py
```diff
@@ -1352,12 +1352,16 @@ def ssh_client(args):
         ssh_destination = f"ssh://{VIRTME_SSH_DESTINATION_NAME}:{args.port}"
 
     force_tty = False
-    if args.cwd is not None:
+    if args.pwd:
+        cwd = get_guest_relative_path(os.getcwd(), args.root)
+        if cwd is None:
```
matttbe, line 1357: I think this check should be removed

virtme/commands/run.py
```diff
@@ -1352,12 +1352,16 @@ def ssh_client(args):
...
+            arg_fail("current working directory is not contained in the root")
+    elif args.cwd is not None:
         cwd = get_guest_relative_path(args.cwd, args.root)
         if cwd is None:
             arg_fail("specified working directory is not contained in the root")
     else:
-        cwd = get_guest_relative_path(os.getcwd(), args.root)
```
matttbe, line 1360: I wonder if this part was not like that no to have to pass `--pwd` (or `--cwd`) again in the client command.

If you change that, I think you should modify vng to pass `--pwd` when `--ssh-client` is used. But then, in your case with a rootfs, you will get an error with the new `arg_fail` you added, but it can be removed I think.

matttbe:
> I'll be sending the fixes here unrelated to the feature itself separately.

Thank you! I just reviewed the fixes part.

> Although not really mergeable in current form, my question is: would `--root-disk` be acceptable in some form?

I'm not sure to understand the issues (but I might not remember all the context here). Do you already have the code to share? Or do you mind explaining why `--root-disk` is still needed, please?

rbmarliere:
> I'm not sure to understand the issues (but I might not remember all the context here). Do you already have the code to share? Or do you mind explaining why `--root-disk` is still needed, please?

Simply put, I would like to use a proper filesystem such as `ext4` for `/` mount point instead of `virtiofs` or `9p`. As I tried to outline in https://github.com/arighi/virtme-ng/pull/484#issuecomment-5103113014, I need a virtual machine that better resemble a proper userspace to investigate failures from other CIs seen in such disks, while keeping the many nice amenities from `vng`. Unfortunately `virtiofs` only works reliably in `x86_64` and being stuck with `9p` in foreign architectures is unfortunate, because it is quite limited by itself and limits the kind of testing one can do with it.

IMHO this is something that could potentially bring in more users. If the idea itself is reasonable, I'm willing to take up the time to present it in a proper way and work it out with you guys the best way to achieve it (reduce duplication, re-organize things as preparation, etc).

matttbe:
> IMHO this is something that could potentially bring in more users.

Maybe, I don't know. Personally, I see virtme as a tool for the kernel dev, to quickly start a VM with the same FS as the host (or another one). If the goal is to use an image for the rootfs, it feels like virtme's "main feature" is not used, and qemu could be used directly. But I understand virtme also bring many other options.

If the integration is small, maybe? But it feels like it might not be if we have to deal with any image. Sorry, I don't have a good vision on what will need to be modified and maintain.

PR #487, opened 2026-08-07
virtme-init: Use setsid -c to steal controlling tty under systemd

rbmarliere:
Hi,

Please consider merging these two small fixes:

virtme-init: Prevent udevd from blocking initialization
virtme-init: Use setsid -c to steal controlling tty under systemd

They were found while testing SLES version 12.5

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -559,9 +559,9 @@ run_user_script_on_console() {
     log 'starting script (console fallback)'
     if [[ -n ${virtme_user:-} ]]; then
         chmod +x /run/tmp/.virtme-script
-        setsid su -c /run/tmp/.virtme-script -- "${virtme_user}" 0<> "/dev/$consdev" 1>&0 2>&0
+        setsid -c su -c /run/tmp/.virtme-script -- "${virtme_user}" 0<> "/dev/$consdev" 1>&0 2>&0
     else
-        setsid bash /run/tmp/.virtme-script 0<> "/dev/$consdev" 1>&0 2>&0
+        setsid -c bash /run/tmp/.virtme-script 0<> "/dev/$consdev" 1>&0 2>&0
     fi
     ret=$?
     log "script returned {$ret}"
@@ -712,9 +712,9 @@ run_user_gui() {
         # Try to fix permissions on the virtual consoles, we are starting X
         # directly here so we may need extra permissions on the tty devices.
         chown -- "${virtme_user}" /dev/char/*
-        setsid bash -c "su -c 'xinit ${xinit_rc}' -- ${virtme_user}" 0<> "/dev/$consdev" 1>&0 2>&0
+        setsid -c bash -c "su -c 'xinit ${xinit_rc}' -- ${virtme_user}" 0<> "/dev/$consdev" 1>&0 2>&0
     else
-        setsid bash -c "xinit ${xinit_rc}" 0<> "/dev/$consdev" 1>&0 2>&0
+        setsid -c bash -c "xinit ${xinit_rc}" 0<> "/dev/$consdev" 1>&0 2>&0
     fi
 }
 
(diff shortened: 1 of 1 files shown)


virtme/guest/virtme-init
```diff
@@ -320,9 +320,9 @@ run_udevd() {
             echo '' > /sys/kernel/uevent_helper
         fi
         log "starting udevd"
-        udev_out=$("$udevd" --daemon 2>&1)
+        "$udevd" --daemon > /run/tmp/udevd.log 2>&1
         if ! grep -q "quiet" /proc/cmdline; then
-            log "udev: $udev_out"
+            log "udev: $(cat /run/tmp/udevd.log)"
```
matttbe, line 325: Can this file be big on such ancient version? (I guess no, but just to be on the safe side: it could be redirected to /dev/null if not in verbose mode)

(Note that on the rust version, no redirection is done, maybe not needed here too?)

virtme/guest/virtme-init
```diff
@@ -559,9 +559,9 @@ run_user_script_on_console() {
     log 'starting script (console fallback)'
     if [[ -n ${virtme_user:-} ]]; then
         chmod +x /run/tmp/.virtme-script
-        setsid su -c /run/tmp/.virtme-script -- "${virtme_user}" 0<> "/dev/$consdev" 1>&0 2>&0
+        setsid -c su -c /run/tmp/.virtme-script -- "${virtme_user}" 0<> "/dev/$consdev" 1>&0 2>&0
```
matttbe, line 562: so passing `-c` is similar to what is done with the rust init program, no?

matttbe (commented):
Thanks! Just two small questions, probably not blocking.

When you edit the init script/program, it would be good to explain the differences between the two, e.g. why you only edit one, do you get closer or diverge from the other one, etc.

virtme/guest/virtme-init
rbmarliere: From my understanding, yes!

rbmarliere:
> When you edit the init script/program, it would be good to explain the differences between the two, e.g. why you only edit one, do you get closer or diverge from the other one, etc.

I tried to include a patch to the Rust init, but honestly gave up and instead was relying on some feedback. It seems to accomplish that outcome we'd need to give up using `utils::run_cmd(udevd_path, args);` so I was unsure. WDYT?

```diff
diff --git a/virtme_ng_init/src/main.rs b/virtme_ng_init/src/main.rs
index ca2faa0bd6ed..8022c3eed02c 100644
--- a/virtme_ng_init/src/main.rs
+++ b/virtme_ng_init/src/main.rs
@@ -712,6 +712,11 @@ fn run_udevd() -> Option<thread::JoinHandle<()>> {
             disable_uevent_helper();
             let args: &[&str] = &["--daemon"];
             utils::run_cmd(udevd_path, args);
+            let _ = Command::new(udevd_path)
+                .arg("--daemon")
+                .stdout(Stdio::null())
+                .stderr(Stdio::null())
+                .status();
             log!("triggering udev coldplug");
             utils::run_cmd("udevadm", &["trigger", "--type=subsystems", "--action=add"]);
             utils::run_cmd("udevadm", &["trigger", "--type=devices", "--action=add"]);
```

virtme/guest/virtme-init
rbmarliere: Noted, I'll do some more testing. Thx!

matttbe:
> > When you edit the init script/program, it would be good to explain the differences between the two, e.g. why you only edit one, do you get closer or diverge from the other one, etc.
> 
> I tried to include a patch to the Rust init, but honestly gave up and instead was relying on some feedback.

Did you notice the same issue with this init?
If you are not able to try, maybe we don't need to support this very old systemd version with this init?

> It seems to accomplish that outcome we'd need to give up using `utils::run_cmd(udevd_path, args);` so I was unsure. WDYT?

If the issue is with `stdout` and `stderr`, I guess yes.

rbmarliere:
> Did you notice the same issue with this init?

Yes it happens, I tried although I don't use it TBH

rbmarliere:
OK, I dropped commit `virtme-init: Prevent udevd from blocking initialization` -- this was actually a corner case due to me passing `--verbose` twice. So there's no need for it.

matttbe (approved):
Thank you for having dropped the problematic commit, and for having updated the commit message

PR #488, opened 2026-08-13
systemd: Don't re-mount /proc, /sys or cgroup2

rbmarliere:
When virtme-init runs as PID 1 it needs to set up procfs, sysfs and the cgroup2 hierarchy from scratch. But under --systemd, virtme-init is instead invoked later as a plain service (the serial-getty@ override), in the same non-isolated mount namespace as the already-running real init, which has already mounted /proc, /sys and /sys/fs/cgroup itself and is actively managing them.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -19,11 +19,11 @@ log() {
 }
 
 mount_kernel_filesystems() {
-    # Mount procfs and sysfs (needed for stat, sadly)
-    mount -t proc -o nosuid,noexec,nodev proc /proc/
-    mount -t sysfs -o nosuid,noexec,nodev sys /sys/
-
     if [[ $$ -eq 1 ]]; then
+        # Mount procfs and sysfs (needed for stat, sadly).
+        mount -t proc -o nosuid,noexec,nodev proc /proc/
+        mount -t sysfs -o nosuid,noexec,nodev sys /sys/
+
         # only mount /run if systemd is not the init
         if ! grep -q "run /run" /proc/mounts; then
             # The fresh tmpfs hides whatever the rootfs kept under /run. On NixOS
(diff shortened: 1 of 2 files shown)


virtme/guest/virtme-init
```diff
@@ -780,7 +780,9 @@ setup_user_session() {
 configure_environment
 mount_kernel_filesystems
 configure_hostname
-mount_cgroupfs
+if [[ $$ -eq 1 ]]; then
+    mount_cgroupfs
```
matttbe: Maybe better to do an early return in this function (like it is drive in some others I think), to keep this bit clean?

virtme_ng_init/src/main.rs
```diff
@@ -1323,7 +1327,9 @@ fn main() {
     configure_environment();
     mount_kernel_filesystems();
     configure_hostname();
-    mount_cgroupfs();
+    if id() == 1 {
+        mount_cgroupfs();
```
matttbe: Same here

matttbe (changes requested):
Thank you for having extracted this fix!

virtme_ng_init/src/main.rs
```diff
@@ -527,6 +531,10 @@ fn mount_kernel_filesystems() {
 }
 
 fn mount_cgroupfs() {
+    if id() != 1 {
+        // systemd is the current init, skip mounting /run
```
matttbe: Detail: cgroup

virtme/guest/virtme-init
```diff
@@ -277,6 +277,10 @@ run_systemd_tmpfiles() {
 mount_cgroupfs() {
     # Set up cgroup mount points (mount cgroupv2 hierarchy by default)
     #
+    if [[ $$ -ne 1 ]]; then
+        # systemd is the current init, skip mounting /run
```
matttbe: Detail: cgroup

matttbe (changes requested):

virtme/guest/virtme-init
rbmarliere: oops

PR #489, opened 2026-08-13
virtme-init: Tolerate a missing /etc/hosts when setting the guest hostname

rbmarliere:
generate_hosts() copies /etc/hosts to /run/tmp/hosts, appends the guest
hostname entries and bind-mounts the result back over /etc/hosts. However,
minimal external roots may ship no /etc/hosts at all, so the copy fails and
the bind-mount then has no source or target:

  cp: cannot stat '/etc/hosts': No such file or directory
  mount: /etc/hosts: mount point does not exist.

### DIFF
--- a/virtme/guest/virtme-init
+++ b/virtme/guest/virtme-init
@@ -192,7 +192,11 @@ generate_fstab() {
 
 generate_hosts() {
     if [[ -n ${virtme_hostname:-} ]]; then
-        cp /etc/hosts /run/tmp/hosts
+        if [[ -e /etc/hosts ]]; then
+            cp /etc/hosts /run/tmp/hosts
+        else
+            touch /etc/hosts
+        fi
         printf '\n127.0.0.1 %s\n::1 %s\n' "$virtme_hostname" "$virtme_hostname" >> /run/tmp/hosts
         mount --bind /run/tmp/hosts /etc/hosts
     fi
--- a/virtme_ng_init/src/main.rs
+++ b/virtme_ng_init/src/main.rs
@@ -396,8 +396,15 @@ fn generate_lvm() -> io::Result<()> {
 
 fn generate_hosts() -> io::Result<()> {
     if let Ok(hostname) = env::var("virtme_hostname") {
-        std::fs::copy("/etc/hosts", "/run/tmp/hosts")?;
-        let mut h = OpenOptions::new().append(true).open("/run/tmp/hosts")?;
+        if Path::new("/etc/hosts").exists() {
+            std::fs::copy("/etc/hosts", "/run/tmp/hosts")?;
+        } else {
+            utils::create_file("/etc/hosts", 0o0644, "")?;
+        }
+        let mut h = OpenOptions::new()
+            .append(true)
+            .create(true)
+            .open("/run/tmp/hosts")?;
         writeln!(h, "\n127.0.0.1 {hostname}\n::1 {hostname}")?;
         utils::do_mount(
             "/run/tmp/hosts",


virtme/guest/virtme-init
```diff
@@ -192,8 +192,10 @@ generate_fstab() {
 
 generate_hosts() {
     if [[ -n ${virtme_hostname:-} ]]; then
-        cp /etc/hosts /run/tmp/hosts
+        : > /run/tmp/hosts
+        [ -e /etc/hosts ] && cp /etc/hosts /run/tmp/hosts
         printf '\n127.0.0.1 %s\n::1 %s\n' "$virtme_hostname" "$virtme_hostname" >> /run/tmp/hosts
+        touch /etc/hosts
```
matttbe: Maybe best to avoid that when the file already exists, e.g. when `--rw` is used. 

What about: if the file exists, copy it. Otherwise, touch it.
(While at it, please use `[[ ... ]]` like above)

virtme_ng_init/src/main.rs
```diff
@@ -396,9 +396,13 @@ fn generate_lvm() -> io::Result<()> {
 
 fn generate_hosts() -> io::Result<()> {
     if let Ok(hostname) = env::var("virtme_hostname") {
-        std::fs::copy("/etc/hosts", "/run/tmp/hosts")?;
+        File::create("/run/tmp/hosts")?;
+        std::fs::copy("/etc/hosts", "/run/tmp/hosts").ok();
```
matttbe: Same here for the logic: if exists, copy or touch

matttbe (changes requested):
Thanks!

virtme_ng_init/src/main.rs
```diff
@@ -396,7 +396,12 @@ fn generate_lvm() -> io::Result<()> {
 
 fn generate_hosts() -> io::Result<()> {
     if let Ok(hostname) = env::var("virtme_hostname") {
-        std::fs::copy("/etc/hosts", "/run/tmp/hosts")?;
+        if Path::new("/etc/hosts").exists() {
+            std::fs::copy("/etc/hosts", "/run/tmp/hosts").ok();
+        } else {
+            utils::create_file("/etc/hosts", 0o0644, "").unwrap_or(());
+        }
+        File::create("/run/tmp/hosts")?;
```
arighi: File::create() truncates /run/tmp/hosts, including when the previous branch has just copied /etc/hosts into it, this loses all existing hosts entries. Could this creation be moved into the else branch so the copied file is preserved?

virtme_ng_init/src/main.rs
```diff
@@ -396,7 +396,12 @@ fn generate_lvm() -> io::Result<()> {
 
 fn generate_hosts() -> io::Result<()> {
     if let Ok(hostname) = env::var("virtme_hostname") {
-        std::fs::copy("/etc/hosts", "/run/tmp/hosts")?;
+        if Path::new("/etc/hosts").exists() {
+            std::fs::copy("/etc/hosts", "/run/tmp/hosts").ok();
```
arighi: Should we propagate potential errors with ? as it was before instead of using .ok()?

virtme_ng_init/src/main.rs
rbmarliere: good catch, thanks!

virtme_ng_init/src/main.rs
```diff
@@ -396,7 +396,12 @@ fn generate_lvm() -> io::Result<()> {
...
 fn generate_hosts() -> io::Result<()> {
     if let Ok(hostname) = env::var("virtme_hostname") {
-        std::fs::copy("/etc/hosts", "/run/tmp/hosts")?;
+        if Path::new("/etc/hosts").exists() {
+            std::fs::copy("/etc/hosts", "/run/tmp/hosts")?;
+        } else {
+            utils::create_file("/etc/hosts", 0o0644, "")?;
+            File::create("/run/tmp/hosts")?;
+        }
         let mut h = OpenOptions::new().append(true).open("/run/tmp/hosts")?;
```
matttbe: A detail, but if we want to align the two init programs, we could do that I suppose:

```suggestion
        }
        let mut h = OpenOptions::new().append(true).create(true).open("/run/tmp/hosts")?;
```

(Remove `File::create` and use `.create(true)`)

To create the file if it doesn't exist.

matttbe (commented):
A small detail, tell me if you prefer not to fix it, fine for me.

arighi (approved):
LGTM, thanks!

matttbe (approved):
Thanks for the update!

PR #490, opened 2026-08-13
run: Mount devtmpfs before the fstab bind-mount on external roots

rbmarliere:
The external-root init one-liner bind-mounts /dev/null over /etc/fstab to blank out entries that may reference devices absent from the guest. This assumes /dev/null already exists as a static node in the target rootfs, which minimal images built without a populated /dev do not provide. The bind-mount then fails right after switch_root, before virtme-init gets a chance to mount devtmpfs itself:

  mount: /etc/fstab: special device /dev/null does not exist.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1729,7 +1729,10 @@ def do_it() -> int:
         else:
             initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
     else:
-        initsh = ["mount -t tmpfs run /run"]
+        initsh = [
+            "mount -t tmpfs run /run",
+            "mount -t devtmpfs devtmpfs /dev",
+        ]
         if needs_guest_cache:
             assert guest_cache_dir is not None
             guest_cache_fstype = export_hostfs(


virtme/commands/run.py
```diff
@@ -1729,7 +1729,10 @@ def do_it() -> int:
         else:
             initcmds = [f"init={guest_tools_path}/{virtme_init_cmd}"]
     else:
-        initsh = ["mount -t tmpfs run /run"]
+        initsh = [
+            "mount -t tmpfs run /run",
+            "mount -t devtmpfs devtmpfs /dev",
```
matttbe, line 1734: Just to be sure: safe to execute when the special files are already there? Will there be an error? If yes, should it be muted and ignored?

matttbe (commented):
Thanks!

virtme/commands/run.py
rbmarliere: From my understanding it's safe. devtmpfs will be mounted on top of /dev, so nothing therein would matter.

PR #491, opened 2026-08-14
run: Only cd ssh guest to host cwd when --pwd is passed

rbmarliere:
ssh_client() cd'd the guest shell into the host's current directory, mapped through --root, even when neither --pwd nor --cwd was requested. That's only meaningful when --root is "/", the default. But with any other --root the resulting path belongs to the host tree, not the guest one, and usually doesn't exist there.

Compute that path only when --pwd is passed, like console_client() and main() already do, so by default the guest's starting directory is left alone.

### DIFF
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -1028,7 +1028,7 @@ class KernelSource:
     def _get_virtme_cwd(self, args):
         if args.cwd is not None:
             self.virtme_param["cwd"] = "--cwd " + args.cwd
-        elif args.root is None:
+        elif args.root is None or args.ssh_client is not None:
             self.virtme_param["cwd"] = "--pwd"
         else:
             self.virtme_param["cwd"] = ""
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1331,12 +1331,14 @@ def ssh_client(args):
         ssh_destination = f"ssh://{VIRTME_SSH_DESTINATION_NAME}:{args.port}"
 
     force_tty = False
-    if args.cwd is not None:
+    if args.pwd:
+        cwd = get_guest_relative_path(os.getcwd(), args.root)
+    elif args.cwd is not None:
         cwd = get_guest_relative_path(args.cwd, args.root)
         if cwd is None:
             arg_fail("specified working directory is not contained in the root")
     else:
-        cwd = get_guest_relative_path(os.getcwd(), args.root)
+        cwd = None
 
     if cwd is not None:
         guest_cwd = "/" if cwd == "." else f"/{cwd}"


matttbe (changes requested):
Hello, 
Thank you for this, but I think you missed my comments from https://github.com/arighi/virtme-ng/pull/484/changes/BASE..b12168af80ed6432c62ef9e507623e762bc9688b#r3759677789

rbmarliere:
> Hello, Thank you for this, but I think you missed my comments from https://github.com/arighi/virtme-ng/pull/484/changes/BASE..b12168af80ed6432c62ef9e507623e762bc9688b#r3759677789

ah yes sorry for overlooking, should be good now hopefully!

Tested with:

```
$ unshare --map-auto --map-root-user --mount ./vng --root /roots/Tumbleweed/x86_64 --run /roots/Tumbleweed/x86_64/boot/vmlinuz --no-virtme-ng-init --user root --ssh 22111

$ ./vng --root /roots/Tumbleweed/x86_64 --ssh-client 22111
Warning: Permanently added 'virtme-ng%22111' (ED25519) to the list of known hosts.
Have a lot of fun...
virtme-ng:~ # logout
Connection to virtme-ng%22111 closed.

$ ./vng --root /roots/Tumbleweed/x86_64 --ssh-client 22111 --cwd /foo/bar
specified working directory is not contained in the root

$ ./vng --root /roots/Tumbleweed/x86_64 --ssh-client 22111 --cwd /roots/Tumbleweed/x86_64
Warning: Permanently added 'virtme-ng%22111' (ED25519) to the list of known hosts.
virtme-ng:/ #



$ ./vng --run /roots/Tumbleweed/x86_64/boot/vmlinuz --ssh 22111

$ ./vng --ssh-client 22111

$ ./vng --ssh-client 22111 --cwd /tmp
( both respects cwd )
```

Commit message updated, thank you!

virtme_ng/run.py
```diff
@@ -1028,7 +1028,7 @@ def _get_virtme_no_root_posix_acl(self, args):
     def _get_virtme_cwd(self, args):
         if args.cwd is not None:
             self.virtme_param["cwd"] = "--cwd " + args.cwd
-        elif args.root is None:
+        elif args.root is None or args.ssh_client is not None:
```
matttbe, line 1031: (I forgot that `--pwd` was passed even with `--ssh-client`.)

Just to be sure: here you need this modification to support the case where the current directory is in the external root, `--root` is given, and you use `--ssh-client`?

When looking at your examples, it is not clear you used this case. But I think it is valid, it is just to be sure ;)

rbmarliere: > Just to be sure: here you need this modification to support the case where the current directory is in the external root, `--root` is given, and you use `--ssh-client`?
> 
> When looking at your examples, it is not clear you used this case. But I think it is valid, it is just to be sure ;)

Yes, that case works too!

matttbe (approved):
thanks for the update!

PR #493, opened 2026-08-17
run: Add a helper for external guest roots

rbmarliere:
Five call sites spell out `args.root != "/"` (or its negation) to ask the same question: is the guest root something other than the host root filesystem, and does it therefore need the guest tools, the cache and the fstab exported and mounted by the init one-liner?

They are easy to confuse with the several other uses of args.root, which resolve host paths against it and genuinely mean the directory rather than the question. Give the question a name.

No functional change.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -683,7 +683,7 @@ def find_kernel_and_mods(arch, args) -> Kernel:
             # If we are using the entire host filesystem or if we are using
             # a chroot (via --root) we don't have to do anything special action
             # the modules, just rely on /lib/modules in the target rootfs.
-            if root_dir == "/" or args.root != "/":
+            if root_dir == "/" or has_external_root(args):
                 kernel.use_root_mods = True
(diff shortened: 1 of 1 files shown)


matttbe (changes requested):
Good idea!

It looks like there are a few more:

```console
$ git grep -n 'args\.root .= "/"'
virtme/commands/run.py:686:            if root_dir == "/" or args.root != "/":
virtme/commands/run.py:736:            if args.root != "/":
virtme/commands/run.py:1225:    return args.root != "/"
virtme/commands/run.py:1607:        if args.root == "/":
virtme/commands/run.py:1670:            posix_acl=(args.root != "/" and not args.no_root_posix_acl),
virtme/commands/run.py:2311:        and args.root != "/"
virtme/commands/run.py:2407:    if args.root != "/":
```

I think it would be better to convert them all, than just a part of them. Do you mind converting the remaining ones then, please?

arighi (approved):
Looks good to me, thanks!

PR #495, opened 2026-08-24
run: fail clearly when `make` is missing for --kconfig/--build

### LINKED ISSUE #485: vng --kconfig return 1 without any hints if there is no make
With a freash f44 system and `virtme-ng 1.41`. Run `vng --kconfig` in linux repo, it returns 1 without any error hints if `make` is not installed.

liuhangbin:
When `make` is not installed (or not in PATH), `vng --kconfig` and `vng --build` can fail with exit code 1 and no user-facing error.

Add an explicit check before invoking configkernel/make flows and print a clear diagnostic so users can install the missing dependency.

Closes: #485

### DIFF
--- a/virtme/commands/configkernel.py
+++ b/virtme/commands/configkernel.py
@@ -405,6 +405,10 @@ def do_it():
             print(conf_item.rstrip("\n"))
         return 0
 
+    if shutil.which("make") is None:
+        sys.stderr.write("Error: 'make' is not installed or not found in PATH.\n")
+        return 1
+
     linuxname = shlex.quote(arch.linuxname)
     archargs = [f"ARCH={linuxname}"]
 
--- a/virtme_ng/run.py
+++ b/virtme_ng/run.py
@@ -921,6 +921,17 @@ class KernelSource:
             make_command += [f"CROSS_COMPILE={cross_compile}", f"ARCH={cross_arch}"]
         # Propagate additional Makefile variables
         make_command += args.envs
+
+        if (
+            not args.dry_run
+            and (args.build_host is None or not args.skip_modules)
+            and shutil.which("make") is None
+        ):
+            arg_fail(
+                "Error: 'make' is not installed or not found in PATH.",
+                show_usage=False,
+            )
+
         if args.build_host is None:
             # Build the kernel locally
             make_command += ["-j", str(args.jobs) if args.jobs else self.cpus]


arighi (changes requested):
Good catch. A small issue: an existing .config combined with `--build-host --skip-modules` doesn't require a local `make`, so we could potentially reject a valid remote-build workflow with this change.

How about moving the config-side check into virtme/commands/configkernel.py, after the `--dry-run` and existing-config/--no-update early returns, but before the first make invocation?

And for the build phase, we can check after config only, when building locally or when the remote build will subsequently run local modules_prepare. Essentially, when `make` is actually needed. Thanks!

matttbe:
@arighi I was writing the same thing :)

But I was also wondering: why is the config part always done on the current host and not the remote one? I was thinking: some kernel options depend on the build environment → but probably not the ones we set here. Except if someone asks for some, e.g. KCSAN, etc. where `make *defconfig` might remove them locally, while they could have been enabled remotely. But maybe we don't need to support that.

If we do, we could send `virtme-configkernel` and run it remotely. But I guess we can wait to see if someone needs that :) (we could add a task or a TODO in the code)

liuhangbin:
> And for the build phase, we can check after config only, when building locally or when the remote build will subsequently run local modules_prepare. Essentially, when `make` is actually needed. Thanks!

Do we need the `make` cmd checking for build phase? In `virtme_ng/run.py: do_it()`, the `make(kern_source, args)
` is after `config(kern_source, args)`

arighi:
> @arighi I was writing the same thing :)
> 
> But I was also wondering: why is the config part always done on the current host and not the remote one? I was thinking: some kernel options depend on the build environment → but probably not the ones we set here. Except if someone asks for some, e.g. KCSAN, etc. where `make *defconfig` might remove them locally, while they could have been enabled remotely. But maybe we don't need to support that.
> 
> If we do, we could send `virtme-configkernel` and run it remotely. But I guess we can wait to see if someone needs that :) (we could add a task or a TODO in the code)

Yes, that's a valid point, Kconfig performsn compiler/toolchain probes, so a local .config might be changed when building locally. Running virtme-configkernel on the remote build host would be more accurate, but then we'd need to handle remote virtme installs, etc. Maybe we can address this separately in another PR if we want to explore this, if someone really needs this.

arighi:
> > And for the build phase, we can check after config only, when building locally or when the remote build will subsequently run local modules_prepare. Essentially, when `make` is actually needed. Thanks!
> 
> Do we need the `make` cmd checking for build phase? In `virtme_ng/run.py: do_it()`, the `make(kern_source, args) ` is after `config(kern_source, args)`

Yes, because config() doesn't always execute, with --build and --no-update, if .config already exists, virtme-configkernel returns early (skipping the check for the local make). So, I think after config() and the args.kconfig early return, we still need something like:
```
if (not args.dry_run and shutil.which("make") is None and
    (args.build_host is None or not args.skip_modules):
    # report missing local make
```

matttbe:
> Running virtme-configkernel on the remote build host would be more accurate, but then we'd need to handle remote virtme installs, etc.

I was initially thinking about sending `virtme-configkernel`, but it needs other internal files. We could package it, but we would still need some Python dependences. Maybe we could execute it locally, except the `make` command that would be done remotely from vng, after the sync.

> Maybe we can address this separately in another PR if we want to explore this, if someone really needs this.

Of course, I can create a new issue.

arighi (approved):

PR #497, opened 2026-08-25
virtme: Find the correct mod_path to use kmod utils

marcosps:
When vng is ran targetting a kernel that was extracted from packages, sometimes the modules.dep file doesn't exists. In these cases the process of creating might fails due to the fixed moduledir setting when compiling kmod with usrmerge.

For example, when running

    vng -r kernels/kernel-default-6.4.0-35.1.x86_64/usr/lib/modules/6.4.0-35-default/vmlinuz

The code will eventually call get_rootfs_from_kernel_path, which might return kernels/..../usr due to other constratins and interoperability.

If modules.dep does not exists for the returned path + lib/modules/{kver}, it then calls depmod, passing the path ending with usr/ as --basedir.

In the case of a newer kmod, or if the default --moduledir is /usr/lib/modules, it will fail to create the modules.dep, and also fail to find the necessary modules to create the initrd to boot the VM.

To solve this issue, and keep everything working for other distributions, keep the root_dir in place, but find a working module path, and use it to create the modules.dep file, if necessary, and to find the other modules.

I found this situation when I extracted the kernel-default package on openSUSE, and just tried to use it with a random rootfs. The extracted directory didn't contain the modules.dep, making it get stuck.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -690,8 +690,8 @@ def find_kernel_and_mods(arch, args) -> Kernel:
                 kernel.modfiles = []
                 kernel.moddir = None
             else:
-                mod_file = os.path.join(kernel.moddir, "modules.dep")
-                if not os.path.exists(mod_file):
+                mod_path = modfinder.get_mod_path(root_dir, kver)
+                if not Path(kernel.moddir, "modules.dep").exists():
                     depmod = find_binary_or_raise(["depmod"])
 
                     if args.verbose:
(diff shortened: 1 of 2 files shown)


marcosps:
I checked the failure here in my machine, and it also fails to boot v6.6 from Ubuntu on openSUSE Tumbleweed, maybe because the version of kmod in machine is too new for it?

Also, for me, the depmod call isn't catching any errors, so maybe it's failing but the modprobe later on works on Ubuntu?

Can you please double check? Thanks!

marcosps:
@rbmarliere has a better approach for this issue. I'll close this PR and let him push his solution.

PR #499, opened 2026-08-25
run: Don't kill ssh-client session on a missing guest cwd

rbmarliere:
ssh_client() chained "cd DIR && exec $SHELL", so a directory that doesn't exist in the guest (e.g. host cwd not present under a --root-disk or a different root than assumed) aborted the whole shell instead of just failing the cd.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1347,10 +1347,10 @@ def ssh_client(args):
 
     if cwd is not None:
         guest_cwd = "/" if cwd == "." else f"/{cwd}"
-        remote_cmd_str = f"cd -- {shlex.quote(guest_cwd)}" + (
-            f" && {args.remote_cmd}"
+        remote_cmd_str = f"cd -- {shlex.quote(guest_cwd)} ; " + (
+            f"{args.remote_cmd}"
             if args.remote_cmd is not None
-            else ' && exec "${SHELL:-/bin/sh}" -i'
+            else 'exec "${SHELL:-/bin/sh}" -i'
         )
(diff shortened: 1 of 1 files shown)


virtme/commands/run.py
```diff
@@ -1348,9 +1348,9 @@ def ssh_client(args):
     if cwd is not None:
         guest_cwd = "/" if cwd == "." else f"/{cwd}"
         remote_cmd_str = f"cd -- {shlex.quote(guest_cwd)}" + (
-            f" && {args.remote_cmd}"
+            f"; {args.remote_cmd}"
             if args.remote_cmd is not None
-            else ' && exec "${SHELL:-/bin/sh}" -i'
+            else '; exec "${SHELL:-/bin/sh}" -i'
```
matttbe: detail: we could take this opportunity to move `;`:

```py
remote_cmd_str = f"cd -- {shlex.quote(guest_cwd)} ; " + (
    f"{args.remote_cmd}"
    if args.remote_cmd is not None
    else 'exec "${SHELL:-/bin/sh}" -i'
)
```

matttbe: Out of curiosity, if the directory doesn't exist, what do you see? Better to keep the error or hide it in your case?

matttbe (commented):
thanks! Just one question (and a detail)

virtme/commands/run.py
rbmarliere: I see the error, and I think it's best to keep it.

matttbe (approved):
thanks for the update!

PR #501, opened 2026-09-01
run: qemu: Use -object monitor-hmp instead of deprecated -mon

arighi:
QEMU 11.1 converted the monitor objects to QOM and deprecated the -mon option in favor of -object with monitor-hmp or monitor-qmp types.

Using -mon produces the following warning on recent QEMU:
```
 qemu-system-x86_64: -mon chardev=console: warning: '-mon' is deprecated,
 use '-object' with 'monitor-hmp' or 'monitor-qmp' types instead
```
Switch to the new -object monitor-hmp syntax when running QEMU >= 11.1 and keep the old -mon syntax as a fallback for older QEMU versions that don't support the monitor-hmp object type yet.

Note that -object requires an explicit id, unlike -mon, so the monitor object is created with id=monitor.

### DIFF
--- a/virtme/commands/run.py
+++ b/virtme/commands/run.py
@@ -1964,7 +1964,10 @@ def do_it() -> int:
         )
         qemuargs.extend(["-serial", "chardev:console"])
         if not args.disable_monitor:
-            qemuargs.extend(["-mon", "chardev=console"])
+            if qemu.has_monitor_object:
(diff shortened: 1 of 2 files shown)


matttbe (approved):
I don't have QEmu 11.1 to test, but I read QEmu's [changelog](https://wiki.qemu.org/ChangeLog/11.1) and it looks good to me, thanks!

PR #502, opened 2026-09-01
vng -g fixes

arighi:
Two independent bugs, both of which break `vng -g`.
- Duplicate console chardev, a bare `-g` (no command) never gets past QEMU
- Missing X11 lock file cleanup: Xorg in the guest fails whenever the host runs an X server on `:0` owned by another user (e.g., a display manager greeter)


matttbe (approved):
Thank you for these fixes!

PR #504, opened 2026-09-10
vng, run, mkinitramfs: Mount the root disk from the initramfs when needed

rbmarliere:
Currently, --root-disk requires the kernel to reach the root device on its own: CONFIG_VIRTIO_BLK, the image's filesystem and the virtio transport all has to be built in, since an initramfs would otherwise take over the root mount before the kernel got a chance.

Fix the initramfs to do this instead. Pull in virtio_blk and the common filesystem modules for --root-disk, and mount root= from the init script the same way a bare kernel would. This lets --root-disk work with module-based kernels, which is the common case for --build'd or distro kernels.

---

Hi @arighi I hope you don't mind this addition to `--root-disk` (~55 lines):

```
 README.md              | 24 +++++++++++++++-------
 virtme/commands/run.py | 78 ++++++++++++++++++++++++++++++++++++-----------------------------------
 virtme/mkinitramfs.py  | 56 ++++++++++++++++++++++++++++++++++++++++-----------
 virtme/virtmods.py     | 14 +++++++++++++
 virtme_ng/run.py       | 13 +++++-------
 5 files changed, 120 insertions(+), 65 deletions(-)
```

I made sure to keep it very direct and as simple as possible.

The goal here is being able to boot a distro kernel using a disk image as rootfs. The appeal for this is that, when testing things and drilling down weird bugs, it's often useful to try a reproducer across different userspaces and existing packaged kernels. Being limited to CONFIG_VIRTIO_BLK and the CONFIG_*_FS builtin and having to build everything over just for a quick confirmation  is far from ideal.

The only caveat is that for foreign architectures, we must force the use of a statically linked `busybox`. This is because the normal `--root` path tries to find a busybox binary within the guest but that's not really possible with `--root-disk` unless it's mounted somewhere through `qemu-nbd` or a loop device. But the user is warned nevertheless so I think this is a good compromise. It's not ideal but it works!

A new example has been added to README.md which hopefully illustrates the usefulness of this:
```
$ vng --arch riscv64 \
      --run ./debian/boot/vmlinux-6.12.107+deb13-riscv64 \
      --root-disk ./debian-sid-nocloud-riscv64-daily.qcow2 \
      --root-dev /dev/vda1 \
      --busybox /path/to/static/riscv64-busybox
```

Cc: @marcosps


virtme/virtmods.py
```diff
@@ -26,3 +26,17 @@
...
+    "fs-ext4",
+    "fs-xfs",
+    "fs-btrfs",
+    "fs-f2fs",
+    "fs-erofs",
+    "fs-squashfs",
+    "fs-vfat",
+    "fs-exfat",
+    "fs-isofs",
+]
```
arighi: No, this is bad, loading every filesystem module can OOM small guests. Only the root filesystem's module should be loaded, ideally selected through rootfstype=, or a dedicated option, or detected where possible. Loading every possible filesystem is wasteful and unsafe with the current default memory.

virtme/commands/run.py
```diff
@@ -2456,6 +2456,18 @@ def get_net_mac(index):
...
 
+    if args.root_disk is not None:
+        # Let the kernel find, probe and mount the root device by itself,
+        # exactly like it would on real hardware. Without rootfstype= it
+        # tries every filesystem it has, and it lists the partitions it
+        # can see if none of them works out.
+        kernelargs.extend(
+            [
+                f"root={qemu.quote_optarg(args.root_dev or '/dev/vda')}",
+                "rootwait",
```
arighi: It's ok to advertise rootwait, but I think the init doesn't implement it, so it's ignored for now (not a blocker, but maybe we should add a comment as a TODO).

arighi (changes requested):

virtme/virtmods.py
rbmarliere: I agree, I was just unsure about adding yet another option. But I think this is sensible, thus added `--root-fstype` in https://github.com/arighi/virtme-ng/commit/878389e4880f310b5011a72d0a84269af07940c1.

Thanks!

arighi (approved):
Looks good now, thanks!

PR #505, opened 2026-09-14
systemd fixes for `--root-disk`

rbmarliere:
Two small fixes (edit: three!) for when using `--systemd` with `--root-disk`. These are hopefully the last remaining bits for the `--root-disk` stuff!

The global disable of SELinux is a bit broader in scope but it's because I've had problems with it in the past apart from `--root-disk`. Seems to be sane to simply disable it.


marcosps:
Worked as expected, thanks for fixing these last issues @rbmarliere !

Just to show here how we are planning to use vng to test our livepatches:

```
vng \
--systemd \
--network user \
--memory 4G \
--user root \
--name bsc1267388 \
--root-disk klp-test-img.qcow2 \
--root-dev /dev/vda3 \
--root-fstype xfs \
--append "panic=10 oops=panic panic_on_warn=1 softlockup_panic=1 hung_task_panic=1" \
--qemu-opts "-no-reboot" \
--verbose \
--disable-microvm \
--cpus 2 \
--run "kernels/kernel-default-6.12.0-160000.26.1.x86_64/usr/lib/modules/6.12.0-160000.26-default/vmlinuz" \
--rodir "/mnt/kgr-test-support/=support-sw" \
--rodir "/mnt/bsc1267388=/home/mpdesouza/klp/livepatches/bsc1267388/tests/x86_64/bsc1267388/" \
--overlay-rw "/mnt" \
--exec "/mnt/bsc1267388/repro/bsc1267388_test_script.sh"
```

As you can see, we can have multiple kernels to start multiple VMs, using the same qcow2 image in RDONLY. For us, the most interesting part is to be able to start the VM and run the script, loading the kernel livepatch and test it properly. This is possible due to the mapping done of the directory inside the VM.

vng makes all this process much easier for us. The usage above is still a prototype, as we are still using some other scripts to trigger qemu manually, requiring doing a lot of things manually that vng gives us for free.

marcosps (approved):
LGTM!

virtme/commands/run.py
```diff
@@ -1801,6 +1802,8 @@ def do_it() -> int:
         kernelargs.append("audit=off")
         # disable zram-generator: may hang at boot if CONFIG_ZRAM is not enabled
         kernelargs.append("systemd.zram=0")
+        # disable SELinux: virtiofs/9p mounts lack xattr support, so guests get denied
+        kernelargs.append("selinux=0")
```
arighi: Hm... this changes existing non --root-disk behavior and prevents using virtme-ng to test SELinux.Could this be scoped more narrowly? Something like:
```
     if args.root_disk is not None:
          kernelargs.append("selinux=0")
```

virtme/commands/run.py
```diff
@@ -1952,11 +1955,14 @@ def do_it() -> int:
                 "/run/virtme/busybox", os.path.basename(host_busybox)
             )
         if args.systemd:
-            initsh.extend(
-                [
-                    "SYSTEMD_UNIT_PATH=/run/virtme/cache: exec /sbin/init",
-                ]
-            )
+            if module_link_path is not None:
```
arighi, line 1962: I think this fixes module setup before systemd only for --root-disk. With a normal --systemd --kdir boot, virtme_link_mods is still handled later by the getty's virtme-init, after boot-time module loading has occurred. Could the early symlink setup apply to every systemd boot that uses virtme_link_mods?

virtme/commands/run.py
```diff
@@ -1952,11 +1955,14 @@ def do_it() -> int:
...
+                initsh.extend(
+                    [
+                        "mount -n -t tmpfs none /lib/modules",
+                        f"ln -s {module_link_path} /lib/modules/$(uname -r)",
```
arighi: Two concerns here:
1. /lib/modules may not exist, virtme-init explicitly creates it first,
2. virtme_link_mods is forwarded to the systemd getty environemnt, so the later virtme-init mounts another tmpfs over /lib/modules on every invocation.

Maybe we can create /lib/modules and set up the tmpfs/symlink before starting systemd. Then pass virtme_root_mods=1 instead of virtme_link_mods, so the later getty virtme-init knows modules are already configured and does not mount another tmpfs.

arighi (commented):
Thanks @rbmarliere for the extra fixes! I added some comments, let me know if they make sense.

virtme/commands/run.py
rbmarliere: Yes, this is what I mentioned in the PR description. I've had problems with SELinux outside the usage of `--root-disk` before and unfortunately they're not quite obvious when it happens. I don't have any examples for this, though, so I'm OK with narrowing the scope for now.

virtme/commands/run.py
rbmarliere: hmm Good point, I'm gonna rework this one. Thanks for the suggestion!

virtme/commands/run.py
rbmarliere: great catch! thank you

arighi (approved):
Looks good to me, thanks!

